PR #231, opened 2014-12-22
Fix calculation of sendfile throughput on OSX

candlerb:
Note that sendfile can return -1 for EINTR whilst having already sent
partial results; these were not being counted previously.


bmah888:
Thanks for the patch!  Possible merge candidate for the 3.0.x stable branch?

PR #242, opened 2015-01-30
Add -a flag to specify authentication.

codyhanson:
If a server is started with an authentication token, only clients who specify the same authentication token are able to run tests.

This is a pretty basic change, and provides _some_ small level of protection from unauthorized iperf server use, but is vulnerable to an attacker snooping the authentication token on the wire between client and server. I started to investigate the use of a hashing function to protect the credential, but wanted to see what other iperf3 users and developers thought.

Future Work:
- Better error reporting when the wrong password is used. Right now it
  just cuts off the socket abruptly, same as other errors.
- Hash the passwords before sending across the wire, and before
  comparing. This way, the secret is never exposed in between the client
  and server. Maybe digest auth is a different flag that you can specify.
- Devise a way to prevent replay attacks, in case hashed credential string is captured in transit.

Thanks,
Cody Hanson

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -205,6 +205,8 @@ struct iperf_test
     TAILQ_HEAD(xbind_addrhead, xbind_entry) xbind_addrs; /* all -X opts */
     int       bind_port;                        /* --cport option */
     int       server_port;
+    char      *authentication;                  /* If set, client and server authentication string must match for test to run */
+    char      *client_authentication;           /* When sent from the client, the server stores token here, so comparison can be made */
     int       omit;                             /* duration of omit period (-O flag) */
     int       duration;                         /* total duration of test (-t flag) */
     char     *diskfile_name;			/* -F option */
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -210,6 +210,12 @@ iperf_get_test_server_hostname(struct iperf_test *ipt)
     return ipt->server_hostname;
 }
 
+char*
+iperf_get_test_authentication(struct iperf_test *ipt)
+{
+    return ipt->authentication;
+}
+
 int
 iperf_get_test_protocol_id(struct iperf_test *ipt)
 {
@@ -330,6 +336,12 @@ void
 iperf_set_test_server_port(struct iperf_test *ipt, int server_port)
 {
     ipt->server_port = server_port;
+}
+
+void
+iperf_set_test_authentication(struct iperf_test *ipt, char* authentication)
+{
+    ipt->authentication = strdup(authentication);
 }
 
 void
@@ -597,6 +609,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
         {"port", required_argument, NULL, 'p'},
         {"format", required_argument, NULL, 'f'},
         {"interval", required_argument, NULL, 'i'},
+        {"authentication", required_argument, NULL, 'a'},
         {"daemon", no_argument, NULL, 'D'},
         {"one-off", no_argument, NULL, '1'},
         {"verbose", no_argument, NULL, 'V'},
@@ -658,7 +671,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 
     blksize = 0;
     server_flag = client_flag = rate_flag = duration_flag = 0;
-    while ((flag = getopt_long(argc, argv, "p:f:i:D1VJvsc:ub:t:n:k:l:P:Rw:B:M:N46S:L:ZO:F:A:T:C:dI:hX:", longopts, NULL)) != -1) {
+    while ((flag = getopt_long(argc, argv, "p:f:i:D1VJvsc:ub:t:n:k:a:l:P:Rw:B:M:N46S:L:ZO:F:A:T:C:dI:hX:", longopts, NULL)) != -1) {
         switch (flag) {
             case 'p':
                 test->server_port = atoi(optarg);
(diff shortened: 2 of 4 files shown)


bmah888:
Interesting idea.  I haven't really had a chance to think about this but at least wanted to you know that someone looked at it.  It's not a part of our use case at ESnet, although it could be useful for others.

codyhanson:
Thanks for the comment. The motivation was to prevent unauthorized throughput tests happening, which could be disruptive on certain networks.

pjebs:
@codyhanson Is there any way to mask the authentication code so people watching the connection can't discover the code. Even if it is only in TLS mode?

codyhanson:
@pjebs It has been a while since I have looked at this, but I think it would be possible to hash the shared secret before it is sent across the insecure channel. I'm not sure what the level of effort would be for that.

pjebs:
I would have thought the aim was to prevent unauthorized clients from using the server in the first place.
If someone intercepts the hashed code, they can use it as if they knew the actual code. From their point of view they don't care if then know the real code. They just want to know the code that the server accepts (which is the hashed code).

codyhanson:
Now that I think about it, you are right. might require a bit higher fidelity level of obfuscation.

pjebs:
I think some way to implement SSL is required. It should be possible in TLS mode.
Not sure how the overhead of deciphering the key for SSL will effect iperf gauge of letency and bandwidth.

TheRealDJ:
Encryption on the control channel should be able to resolve this concern.

pjebs:
What is the control channel? Can you explain further how to achieve this?

TheRealDJ:
Sorry, more of a comment to the devs. This would need to be built into iperf. 

When iperf3 client starts, it opens a control channel to the server. This allows the two to coordinate the testing and share data.

ralcini:
I'm thinking about that and, if I'm not wrong, the introduction of SSL on the control channel is not so easy.
What about an alternative approach based on RSA encryption of a message containing the user credentials and a timestamp? 
I try to better explain my idea:

 - the iperf client accepts as input via command line the credentials and the file path to the RSA public key
 - using the same approach seen in this pull request, the iperf client sends the encripted message to the server
 - the server decrypts the message, validates the credentials contained and checks that the timestamp contained is near to the current server time (e.g.: +/- 30 seconds).

Requirements:
 - dependency to openssl library
 - generation of a RSA keypair to setup the service 
 - admins must send the public key to the final users with the assigned credentials

What are your thought about that?

TheRealDJ:
Makes sense to me. This would really help my project out as well!

Thanks.
DJ

ralcini:
Take a look at this https://github.com/ralcini/iperf, I tried to implement authentication using the previously described approach. Any feedback is welcome!
In the man page I've added some useful information.

TheRealDJ:
I can't seem to get authentication to work. Here is what I did:
Server: iperf/src/iperf3 -s --rsa-private-key-path server_key.pem --authorized-users-path auth_users.csv
Client: iperf/src/iperf3 -c localhost --rsa-public-key-path server_key.pub --username=testuser
Client: iperf/src/iperf3 -c localhost --rsa-public-key-path server_key.pub --username testuser

Client Output:
iperf/src/iperf3 -c localhost --rsa-public-key-path server_key.pub --username=testuser
Password: Connecting to host localhost, port 5201
iperf3: error - test authorization failed

Server Output:
Server listening on 5201
Authentication failed for user '' ts 1488403095
iperf3: error - test authorization failed

Server Output (with Debug):
Server listening on 5201
get_parameters:
{
	"tcp":	true,
	"omit":	0,
	"time":	10,
	"parallel":	1,
	"len":	131072,
	"authtoken":	"A+dcFI+FZYzMIaZW5KYQY0k50JkWhxTsBAi1xoqD9O1v3tTPMux4T6WmL1ifTsEiBeRDlH/HE41UhoI1bqsQuxIPvznFm6PEIAj23PizFhIGL9pAqC1C9EWVdVntphA89kICVBt7OG9TJWcO2Trb52zDRv2dU7ee1j67DkzJWTJ0LmnxF1X+PlVn0RfLKPyDkeLWg2aw5rbutZ7vTedIbmhUwNSob5uI+Abvy+vTvtAkO7SQMUiJgPqP6w7KYizdXOEubc1sy8A4lhGHa3sI95iBQxNICEXZXL6HF1F/sRgJsC7Bh2V1C5B2Fb81ztYpBtvZCBe1iojXFztD34qrJ/I/eWHoWPAaBe7lobLnxuC6CRQRJbab1IVCHX3GmPlFkMyQwYIuzQw/w6yJRixd37ywRKFOYEMCb944BwO3HesaP3RbdoBdEuHj0IrBcmz1BbATI3G6LojDIw/oU0jAGilOw+LuFnKxX9jWEwedblraOa10QHn4s8U73/DBVqNK3gu7UKrDe5v5JzIIsJBuOks7fMWY+Hl/C4qOx/DvQr0vmv+3J5VJOvpFfrffOFVeh3Hh0hK8rXWojyn+cYOMyUnMSpwCB+dXMXLTPbA5bpwoujsx9UrwXotQTEPHlVdvYCpNiOtW2fnA5b0uWHGMYqLjZrGK9g8D9mdTpdawtNs=",

Interesting on the server output shows "user ''" ... Seems like its not getting the client's username.

One more note: The Client and Server are the same Mac Book Pro.

Thanks.
DJ

ralcini:
..strange behaviour, not triggered on CygWin and Ubuntu 64bit. 
Can you pull the new changes, rebuild, retest in debug mode and show me the server output?

@TheRealDJ please move to https://github.com/esnet/iperf/pull/517

bmah888:
Superceded by a different implementation in #517.  We thank you for your pull request and ideas.

Issue #243, opened 2015-02-08
iperf3 retransmits reporting on FreeBSD


he32:
Hi,

while porting the TCP_INFO getsockopt() feature to NetBSD from
FreeBSD, I noticed that iperf3 has retransmit reporting turned off
for FreeBSD because "it does not work".  Looking at the kernel code
for FreeBSD's TCP_INFO, it looks like they report retransmits via
another field in the tcp_info struct, so this change is probably appropriate:

@@ -82,8 +83,9 @@ has_tcpinfo_retransmits(void)
     return 1;
 #else
 #if defined(__FreeBSD__) && __FreeBSD_version >= 600000
-    /* return 1; */
-    return 0;  /* FreeBSD retransmit reporting doesn't actually work yet */
+    return 1; /* Should work now */
@@ -114,12 +118,30 @@ get_total_retransmits(struct iperf_inter
 {
 #if defined(linux) && defined(TCP_MD5SIG)
     return irp->tcpInfo.tcpi_total_retrans;
-#else
-#if defined(__FreeBSD__) && __FreeBSD_version >= 600000
-    return irp->tcpInfo.__tcpi_retransmits;
+#elif defined(__FreeBSD__) && __FreeBSD_version >= 600000
+    return irp->tcpInfo.tcpi_snd_rexmitpack;

(The line numbers are probably off, since I have NetBSD changes
in there too.)

Regards,

- Håvard

he32:
(oh, well, the diff got mangled by the markdown, but I think the gist is still understandable)

bmah888:
This was previously fixed by the pull request #243.  Closing this as a duplicate of the pull request.

PR #244, opened 2015-02-11
Freebsd fix

he32:
The FreeBSD kernel counts retransmissions in tcpi_snd_rexmitpack.
This changes iperf accordingly, enables retransmissions reporting for FreeBSD,
and uses #elif in a few places to ease the introduction of support for
another OS.

- Havard

### DIFF
--- a/src/tcp_info.c
+++ b/src/tcp_info.c
@@ -79,13 +79,10 @@ has_tcpinfo_retransmits(void)
     ** tcpi_total_retrans field that we need; if not, not.
     */
     return 1;
-#else
-#if defined(__FreeBSD__) && __FreeBSD_version >= 600000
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the patch!  Hoping to get time to look at this soon-ish.

Does this pull request replace issue #243?

he32:
Yes, it replaces issue #243.

he32:
I should perhaps note that I've not actually tested the suggested change on FreeBSD, but I had reason to look at the kernel parts of the TCP_INFO functionality, and spotted the problem when I ported that part to NetBSD, where it's not yet committed but hopefully will be soon.

bmah888:
Superceded by #248.

PR #245, opened 2015-02-12
Two more casts for tolower() to silence compiler warnings

he32:
Add casts for the two remaining tolower() uses, since
tolower() is undefined for negative character values
(except for -1).

Also simplify the existing casts for tolower() uses.


bmah888:
Merged, thanks!

PR #248, opened 2015-02-19
Add support for NetBSD systems which have TCP_INFO implemented.

he32:
This adds support for using TCP_INFO on NetBSD if it's present.
It also enables use of TCP_INFO on FreeBSD, and fixes the "get cwnd" code
to pull info from the right field.

This pull request supersedes pull request #244

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -66,8 +66,9 @@ struct iperf_interval_results
     int       cnt_error;
 
     int omitted;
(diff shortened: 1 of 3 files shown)


bmah888:
This patch leaves an extra ```#endif``` that breaks compilation.  I'll fix it after pulling this change.

bmah888:
All done now, thanks for the patch!

Issue #257, opened 2015-03-31
Segfault in the signal handler


manojdelphix:
iperf3 server runs in a loop like this (simplified):

run():
for (;;) {
  -> iperf_run_server(test)
       -> setup signal handler
       -> iperf_server_listen
            -> bind(s)
            -> test->listener = listen(s,5)
            -> print banner
       -> cleanup_server(test);
            -> close(test->listener);
  -> iperf_reset_test(test);
}

iperf3 sets up a signal handler in a loop. The routine that sets it up returns at the end of a test. Once it returns, the stack context saved by setjmp is no longer valid (per man setjmp(3)).

Until the signal handler is reestablished in the next iteration, the saved stack context is invalid. Hence a signal received during this window will result in a segfault.

I am able to demonstrate this by adding a sleep(3) in the window.

> sybase@bbdhcp:/home/sybase/iperf-3.0.11/src-> ./iperf3 -s
> -----------------------------------------------------------
> Server listening on 5201
> -----------------------------------------------------------
> Accepted connection from 10.43.11.189, port 61818
> [  6] local 10.43.13.229 port 5201 connected to 10.43.11.189 port 32916
> [ ID] Interval           Transfer     Bandwidth
> [  6]   0.00-1.00   sec  91.5 MBytes   768 Mbits/sec
> [  6]   1.00-2.00   sec  75.6 MBytes   634 Mbits/sec
> [  6]   2.00-3.00   sec  77.8 MBytes   652 Mbits/sec
> [  6]   3.00-4.00   sec  84.8 MBytes   711 Mbits/sec
> [  6]   4.00-5.00   sec  92.0 MBytes   771 Mbits/sec
> [  6]   5.00-6.00   sec  79.0 MBytes   663 Mbits/sec
> [  6]   6.00-7.00   sec  96.1 MBytes   806 Mbits/sec
> [  6]   7.00-8.00   sec  63.4 MBytes   532 Mbits/sec
> [  6]   8.00-9.00   sec  62.1 MBytes   521 Mbits/sec
> [  6]   9.00-10.00  sec  78.9 MBytes   662 Mbits/sec
> [  6]  10.00-10.00  sec   146 KBytes   705 Mbits/sec
> - - - - - - - - - - - - - - - - - - - - - - - - -
> [ ID] Interval           Transfer     Bandwidth
> [  6]   0.00-10.00  sec   816 MBytes   684 Mbits/sec                  sender
[...]

PR #258, opened 2015-03-31
257 Segfault in the signal handler

manojdelphix:
Fix for issue #257 Segfault in the signal handler
https://github.com/esnet/iperf/issues/257

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -328,15 +328,6 @@ iperf_client_end(struct iperf_test *test)
 }
 
 
-static jmp_buf sigend_jmp_buf;
-
-static void
-sigend_handler(int sig)
-{
-    longjmp(sigend_jmp_buf, 1);
-}
-
-
 int
 iperf_run_client(struct iperf_test * test)
 {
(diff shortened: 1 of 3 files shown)


bmah888:
Thanks for the bug report and pull request.  I haven't had a chance to really think about it, but we have had weird, occasional, unexplained crashes at exit time and it'd be very nice if this was the solution.

manojdelphix:
If you have core dumps, I could take a look.

In the ones I got, there was a null pointer dereference in iperf_got_sigend. The context saved by setjmp(3) was not restored correctly and test was NULL.

This was the stack trace.

Program terminated with signal 11, Segmentation fault.
#0  iperf_got_sigend (test=0x0) at iperf_api.c:2542
(gdb) where
#0  iperf_got_sigend (test=0x0) at iperf_api.c:2542
#1  0x000000000040b8a7 in iperf_run_server (test=0x0) at iperf_server_api.c:453
#2  0x000000000040219a in run (argc=4220440, argv=<value optimized out>) at main.c:146
#3  main (argc=4220440, argv=<value optimized out>) at main.c:105
(gdb)

bmah888:
Fixed, thanks!

PR #272, opened 2015-06-06
iflush for interval results

EduardoAbinader:
In order to keep the same behavior of stdout, iflush the results
to files as they are printed, when logfile is set.


EduardoAbinader:
Yes, that behavior. Today the contents of log files (--logfile) are printed in 5 seconds base. With this flush, every iprintf is written to the logfile, behaving like default stdout.

bmah888:
OK, that sounds reasonable.  Committed, thanks for the patch!

PR #288, opened 2015-08-02
Add tmp_path in order to make the template path workable on iOS.

yeahdongcn:
Since "/tmp/iperf3.XXXXXX" is not allowed to access on iOS device (although working on simulator), I added a tmp_path variable to redirect the template to other place inside the application sandbox. The calling is shown as below.

	NSString *path = [NSTemporaryDirectory() stringByAppendingPathComponent:@"iperf3"];
	[[NSFileManager defaultManager] createDirectoryAtPath:path withIntermediateDirectories:YES attributes:nil error:nil];
	char buf[PATH_MAX];
    [path getCString:buf maxLength:PATH_MAX encoding:NSUTF8StringEncoding];
                
	struct iperf_test *test = iperf_new_test();
	if (!test) {
		iperf_errexit(NULL, "create new test error - %s", iperf_strerror(i_errno));
	}
	// ...
	iperf_set_test_tmp_path(test, buf);

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -202,7 +202,7 @@ struct iperf_test
     struct protocol *protocol;
     signed char state;
     char     *server_hostname;                  /* -c option */
-    char     *tmp_path;                         /* -c option */
+    char     *template;                         /* -c option */
     char     *bind_address;                     /* first -B option */
     TAILQ_HEAD(xbind_addrhead, xbind_entry) xbind_addrs; /* all -X opts */
     int       bind_port;                        /* --cport option */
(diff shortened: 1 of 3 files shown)


bmah888:
Thanks for the patch!  So I am guessing that on iOS you have some front-end code that calls into libiperf?  There isn't any way to set that tmp_path variable from the normal command-line.

The only thing I'd recommend is to not use strcpy(3), which you did in two places.  Maybe strncpy(3) instead, so we don't overflow tmp_path somehow.

yeahdongcn:
You're right, I will get it updated.

yeahdongcn:
I also got a repo to demonstrate how to use this in iOS project.

PR #295, opened 2015-08-13
Cygwin compatibility

Dji75:
- Rename iprintf function to iperf_iprintf because iprintf already exists in cygwin's headers files (stdio)

- Modify template path from hardcoded one:
char template[] = "/tmp/iperf3.XXXXXX";
  To local:
char template[] = "./iperf3.XXXXXX";

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -481,11 +481,11 @@ iperf_on_test_start(struct iperf_test *test)
     } else {
 	if (test->verbose) {
 	    if (test->settings->bytes)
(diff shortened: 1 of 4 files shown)


bmah888:
23ceabb concerns me.  For UNIX-like systems, there's no guarantee that you can write to the current working directory.  The original pathname is the normal UNIX way of doing things...does Cygwin not emulate / support that aspect of a UNIX-like runtime environment?

If there's no way to reconcile these, maybe there's a way to make that variable definition conditional, so on Cygwin it uses the current working directory and on any other OS, it uses the original /tmp path?

bmah888:
The parts of this pull request that make sense to include have been merged via pull request #468.  Closing this issue.

PR #305, opened 2015-09-15
Fixes a smashing stack problem as described in https://bugzilla.kerne…

teto:
…l.org/show_bug.cgi?id=104601

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -457,10 +457,11 @@ setnonblocking(int fd, int nonblocking)
 int
 getsockdomain(int sock)
 {
-    struct sockaddr sa;
+    struct sockaddr_storage sa;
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for looking into this and submitting a patch!  I'm looking at this patch now.  You leave the variable ```len``` uninitialized in the new code.  Doesn't it still make sense to initialize it to ```sizeof(sa)``` as the original code did?  Just in case there's some OS that actually does check that value?

teto:
You are right, while this posed no threat to the linux version (that ignores the input field), it may be a problem on BSD etc... so I fixed it.

bmah888:
Merged to master.  Thanks!

PR #311, opened 2015-10-11
Fix bug in iperf_new_stream leading to EINVAL

tniessen:
The commit 9a3775091b7a23c38053f7fffb8d53d49fcce02f introduced a bug in `iperf_new_stream` in `iperf_api.c`. The last character of the template is cut off due to incorrect use of `snprintf` ([here](https://github.com/esnet/iperf/commit/9a3775091b7a23c38053f7fffb8d53d49fcce02f#diff-109ce0347bb7c189bead5f3b4da4f468R2595) and [here](https://github.com/esnet/iperf/commit/9a3775091b7a23c38053f7fffb8d53d49fcce02f#diff-109ce0347bb7c189bead5f3b4da4f468R2598)). As the last six characters of the template must be `"XXXXXX"`, `mkstemp` produces the error `EINVAL` and tests fail with the message "Invalid argument".

The second argument to `snprintf` should not be the length of the data to print (`strlen(test->template)` or `strlen(buf)`, respectively) but the size of the output buffer in chars: `sizeof(template) / sizeof(char)`.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2592,10 +2592,10 @@ iperf_new_stream(struct iperf_test *test, int s)
     
     char template[1024];
     if (test->template) {
-        snprintf(template, strlen(test->template), "%s", test->template);
+        snprintf(template, sizeof(template) / sizeof(char), "%s", test->template);
     } else {
         char buf[] = "/tmp/iperf3.XXXXXX";
-        snprintf(template, strlen(buf), "%s", buf);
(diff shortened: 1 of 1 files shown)


bmah888:
Yikes!  I did read 9a37750 (for the record that came from pull request #288), but I completely missed this.  And for some reason when I run-tested on MacOS, it worked anyway.  Thanks for giving a patch.  It looks correct so I'm going to pull it in.

PR #313, opened 2015-10-23
Honor $TMPDIR

cernekee:
Some OSes, such as Android, do not have a /tmp directory.  If the user sets $TMPDIR, honor that setting, otherwise default to /tmp.

For Android, `TMPDIR=/data/local/tmp` was tested.

@Dji75 does this help with the /tmp issue you were seeing on Cygwin (bug #295)?

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2594,8 +2594,9 @@ iperf_new_stream(struct iperf_test *test, int s)
     if (test->template) {
         snprintf(template, sizeof(template) / sizeof(char), "%s", test->template);
     } else {
-        char buf[] = "/tmp/iperf3.XXXXXX";
-        snprintf(template, sizeof(template) / sizeof(char), "%s", buf);
+        char *tmpdir = getenv("TMPDIR");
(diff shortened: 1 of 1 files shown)


bmah888:
I'm wondering how this interacts with the "template" change that was the subject of #288 (which fixed a problem someone was having trying to port to iOS).

If we take the change in this pull request, there would then be two different ways to set the path to a temporary directory.  The fact that one is in support of Android and the other is in support of iOS is not lost on me.  :-)

I wonder if it's possible to make a $TMPDIR setting propagate into the tmp_template?  Layering seems a bit cleaner than having two different mechanisms to accomplish similar goals.

bmah888:
Is there still a need / desire for this functionality?  If so, what about the alternate implementation I suggested in my last comment, where if $TMPDIR is defined we use that to set the already-existing tmp_template?

bmah888:
Something along the lines of my "alternate implementation" I proposed above was done by @ralcini as as part of PR #517, although I didn't realize it until now.  Somewhat belatedly closing this pull request, thanks for your submission and apologies for sitting on it for so long.

PR #319, opened 2015-10-30
3.0 stable

latencie:


bmah888:
This looks like a mistaken attempt to sync the iperf-3.0 stable branch with the mainline.  Closing without merging.

Issue #331, opened 2015-11-26
build using musl libc


ScrumpyJack:
iperf3 version 3.1.1

building using musl libc fails (this is on Alpine Linux, so it's a follow up of #328)

iperf.h:71:21: error: field 'tcpInfo' has incomplete type
     struct tcp_info tcpInfo; /\* getsockopt(TCP_INFO) for Linux, {Free,Net}BSD */

This was fixed (probably incorrectly) with 
<diff>
-struct tcp_info tcpInfo; /\* getsockopt(TCP_INFO) for Linux, {Free,Net}BSD _/
+struct tcp_info *tcpInfo; /_ getsockopt(TCP_INFO) for Linux, {Free,Net}BSD */
</diff>

but then more errors occured

``` c
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -Os -fomit-frame-pointer -Os -fomit-frame-pointer -Wall -MT tcp_info.lo -MD -MP -MF .deps/tcp_info.Tpo -c tcp_info.c  -fPIC -DPIC -o .libs/tcp_info.o
tcp_info.c: In function 'save_tcpinfo':
tcp_info.c:100:40: error: invalid application of 'sizeof' to incomplete type 'struct tcp_info'
     socklen_t tcp_info_length = sizeof(struct tcp_info);
                                        ^
tcp_info.c:107:21: error: request for member 'tcpi_snd_cwnd' in something not a structure or union
         irp->tcpInfo.tcpi_snd_cwnd, irp->tcpInfo.tcpi_snd_mss,
                     ^
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -Os -fomit-frame-pointer -Os -fomit-frame-pointer -Wall -MT tcp_window_size.lo -MD -MP -MF .deps/tcp_window_size.Tpo -c tcp_window_size.c  -fPIC -DPIC -o .libs/tcp_window_size.o
tcp_info.c:107:49: error: request for member 'tcpi_snd_mss' in something not a structure or union
         irp->tcpInfo.tcpi_snd_cwnd, irp->tcpInfo.tcpi_snd_mss,
                                                 ^
tcp_info.c:108:21: error: request for member 'tcpi_rtt' in something not a structure or union
         irp->tcpInfo.tcpi_rtt);
                     ^
tcp_info.c: In function 'get_total_retransmits':
tcp_info.c:119:24: error: request for member 'tcpi_total_retrans' in something not a structure or union
[...]

PR #344, opened 2016-01-30
Fix build using musl libc

bkuhls:
Fixes https://github.com/esnet/iperf/issues/331

### DIFF
--- a/src/cjson.h
+++ b/src/cjson.h
@@ -23,6 +23,8 @@
 #ifndef cJSON__h
 #define cJSON__h
 
+#include <stdint.h>
+
 #ifdef __cplusplus
 extern "C"
 {
--- a/src/timer.h
+++ b/src/timer.h
@@ -30,6 +30,7 @@
 #ifndef __TIMER_H
 #define __TIMER_H
 
+#include <stdint.h>
 #include <sys/time.h>
 
 /* TimerClientData is an opaque value that tags along with a timer.  The


bmah888:
Thanks for the pull request.  In most (all?) places where we including <stdint.h> we do this conditionally like this:
```
#ifdef HAVE_STDINT_H
#include <stdint.h>
#endif
```
Do you think your change would work if the #include was done this way?  Perhaps I'm being pedantic but everyplace else we make sure that GNU autoconf actually found that header file before trying to include it.

If you're not sure, you can search for STDINT in the file ```src/iperf_config.h``` after you've run configure...we want to see something that looks like:
```
#define HAVE_STDINT_H 1
```

joerg-krause:
From [Header Portability - Autoconf](https://www.gnu.org/software/autoconf/manual/autoconf-2.61/html_node/Header-Portability.html):

> inttypes.h vs. stdint.h
The C99 standard says that inttypes.h includes stdint.h, so there's no need to include stdint.h separately in a standard environment. Some implementations have inttypes.h but not stdint.h (e.g., Solaris 7), but we don't know of any implementation that has stdint.h but not inttypes.h. 

From [Solaris Porting FAQ](http://wiki.opencsw.org/porting-faq#toc1)

> Solaris only includes stdint.h starting from version 10. For previous versions, use sys/inttypes.h instead. To make it suitable for upstream inclusion, you can shield it for Solaris.

> ```
#ifdef HAVE_STDINT_H
#  include <stdint.h>
#  else
#  ifdef HAVE_INTTYPES_H
#    include <inttypes.h>
#  endif
#endif
```

@bkuhls Any plans for an updated PR?

doughdemon:
This PR is also not enough to fix compilation with musl. In addition, _GNU_SOURCE needs to be globally defined so that musl exposes struct tcp_info in <netinet/tcp.h>, which is used in iperf.h. (Defining _GNU_SOURCE just in iperf.h is not enough because <netinet/tcp.h> is included in iperf_server_api.c before iperf.h.) See http://patchwork.ozlabs.org/patch/572187/raw/ for the patch actually used in buildroot.

bmah888:
This pull request has been replaced by #518...thanks for the submission.

PR #349, opened 2016-02-04
Adding feature to support TCP_MD5SIG option for Linux. [OLD, newer pull req #350]

critindirecx:
Signed-off-by: Tejas Wanjari <tejaswanjari@gmail.com>

PR #350, opened 2016-02-07
Adding feature to support TCP_MD5SIG option for Linux.

critindirecx:
Signed-off-by: Tejas Wanjari tejaswanjari@gmail.com

### DIFF
--- a/configure
+++ b/configure
@@ -12704,6 +12704,37 @@ $as_echo "#define HAVE_TCP_CONGESTION 1" >>confdefs.h
 
 fi
 
+# Check for TCP_MD5SIG sockopt (believed to be Linux only)
+{ $as_echo "$as_me:${as_lineno-$LINENO}: checking TCP_MD5SIG socket option" >&5
+$as_echo_n "checking TCP_MD5SIG socket option... " >&6; }
+if ${iperf3_cv_header_tcp_md5sig+:} false; then :
+  $as_echo_n "(cached) " >&6
+else
+  cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
+#include <netinet/tcp.h>
+#ifdef TCP_MD5SIG
+  yes
+#endif
+
+_ACEOF
+if (eval "$ac_cpp conftest.$ac_ext") 2>&5 |
+  $EGREP "yes" >/dev/null 2>&1; then :
+  iperf3_cv_header_tcp_md5sig=yes
+else
+  iperf3_cv_header_tcp_md5sig=no
+fi
+rm -f conftest*
+
+fi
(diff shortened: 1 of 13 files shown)


critindirecx:
This is an updated patch, with no unneeded whitespaces (unlike previous #349). Please let me know if you have suggestions/comments, or if anything else is needed.

bmah888:
Just a comment to acknowledge the updated pull request.  I'm context-switched into some other projects right now, but I look forward to finding time to really look at it.  Thanks!

bmah888:
Notes to myself as I'm looking at this change (@critindirecx, there is nothing for you to do at this point, just that I finally got around to really looking at this):  TCP_MD5SUM exists on FreeBSD, but FreeBSD uses a very different way of specifying the MD5 key / association.  I'm not even sure how to do this programmatically.  So we need (for the moment) to add a configure-time test to make sure we only try this on Linux.

The function ```__set_tcp_md5sig``` shouldn't be named with two leading underscores, because those names are reserved for C implementations.  Maybe iperf_set_tcp_md5sig?

I need to understand the change(s) made to the loops that allocate port numbers, it looks like there were a few adjustments to this logic.  Need to be sure that a client post-change can communicate with a server pre-change and vice versa.

Might need some manpage text.

critindirecx:
Sadly there are not Linux man pages for TCP_MD5SIG socket option. So I ended up investigating and writing a how-to on my blog: [TCP_MD5SIG: An Undocumented Socket Option in Linux](http://criticalindirection.com/2015/05/12/tcp_md5sig/). It also has examples.

Hope it reduces some effort on your part.

critindirecx:
@bmah888 What would be a timeline for getting this patch in mainline? It has been a long time submitting it, an update on status would be really helpful. Thanks!

bms:
It should probably be rethought for TCP-AO.

I appreciate folk use TCP-MD5 for testing against routers and so on for BGP sessions. Unfortunately, it is legacy crud.

Disclaimer: I wrote the FreeBSD implementation.

So to do it programmatically in FreeBSD, you'd have to clone some of what setkey(8) is doing [with the SPI](https://www.freebsd.org/cgi/man.cgi?query=setkey&sektion=8).

The main reason that happened is because in 2003, whenever I tried to do it through the SPD (the 'right' way -- SPI ID 0x1000 just happens to be far away from ranges which IKE daemons use for transient IPSEC associations) I hit panic city.

bmah888:
I'm (somewhat reluctantly) going to close this without merging.

The TCP-MD5 use case isn't really one that we see a lot of need for, and as noted in the discussion above it's very hard to implement this in a portable way.  Possibly because of the APIs involved, the patch contains a lot of cut-and-paste / nearly-duplicated code that would make our code more difficult to maintain going forward.

Thanks to @critindirecx for the submission and for working with me on the issues with the patch.

PR #377, opened 2016-03-06
Add option `--forceflush` to force flushing of output at the end of each interval

### LINKED ISSUE #299: Use QProcess to control iperf3 in QT
Hi

I use QProcess to control iperf3 of windows version in QT , my purpose is to realtime get the throughput
and to show on my UI. For example , if interval is 10 secs , i want to fetch the result every second . But i only got whole result after 10 secs ,  it seems that something triggered(no matter terminate(ctrl+c) or throughput finished) then send output to the process. Below is my code(just simple test)

void MainWindow::on_pushButton_clicked()
{
    QStringList l;

    connect(&p , SIGNAL(readyRead()) , this , SLOT(print_message()));
    l<<"-c"<<"127.0.0.1"<<"-i"<<"1";
    p.start("d:/iperf3" , l);

    while(true)
    {
        p.waitForFinished(1000);
    }
}

void MainWindow::print_message()
{
    QString tmp;
    tmp = QString(p.readAllStandardOutput());
    if(!tmp.isEmpty())
        qDebug()<<tmp;
}

Do you have any idea?It bother me a few days.
Thanks
Jack

quasoft:
Fixes #299

As discussed in issue #299 piping the output of iperf to another process, introduces buffering in output and delay in receiving updates by the other process.

This can be fixed by using tools like unbuffer, expect and stdbuf on UNIX/Linux, but those tools are not easy to get to work in Windows.

This pull request adds a command line option named --forceflush that allows the user to request flushing of output at the end of each interval. That way users can take advantage of this option if (and only if) buffering output (eg. when piping output) is a problem in their specific case, without the necessity of relying on external tools.

Similar problem existed when piping output to a file. The file was not updated fast enough and could not be monitored in realtime. Pull #272 fixed that by calling iflush on iprintf in interval.

Current pull request uses the same approach, but adds a command line option named --forceflush (optional) that can be used to force flushing, even if not logging output to file.


bmah888:
Merged, thanks!

bmah888:
Merged to 3.1-STABLE branch in 72ab703 per discussion in #458.

PR #389, opened 2016-03-29
fix server crash for intervals with 0 duration

### LINKED ISSUE #388: server crash in unit_snprintf
iperf3 server may crash if it ends up with 0 duration interval at the end of the test. The 0 duration interval can occur if the server and client have small skew in intervals. The server may start new interval and then immediately receive TEST_END.
When this happens the bandwidth is NaN and the unit_snprintf crashes with invalid index in label_bit (or label_byte).
Suggested fix is to skip printing intervals with 0 duration.

yanavlasov:
The 0 duration interval can occur if the server and client
have small skew in intervals. The server may start new interval and then
immediately receive TEST_END.
When this happens the bandwidth is NaN and the unit_snprintf crashes with
invalid index in label_bit (or label_byte).

Fixes #388

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2518,6 +2518,11 @@ print_interval_results(struct iperf_test *test, struct iperf_stream *sp, cJSON *
 	iperf_err(test, "print_interval_results error: interval_results is NULL");
         return;
     }
+    if (irp->interval_duration == 0) {
+        /* server may end-up with 0 duration interval if it receives TEST_END right
+           after beginning new interval */
+        return;
+    }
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the pull request!  We've fixed other problems of this nature by just letting the value be zero instead of NaN (in those cases it was loss rate instead of bandwidth).  I haven't really thought yet about what's the most appropriate way to do that for this case.

bmah888:
Thanks again for the pull request.  I'm going to fix this problem (#388) in a slightly different way to be consistent with earlier fixes of this type.

Issue #412, opened 2016-05-13
Using Omit option with UDP can produce negative loss count and loss percentage


joemce:
iperf 3.1.2

If the --omit option is used with --udp, the first interval after omit can show negative lost and percentage lost values.

``` text
% iperf3 -c 202.254.1.4 -p 19750 --parallel 1 --time 30 --affinity 2 --format m --bandwidth 0M --len 1472 --window 512M --zerocopy --udp-counters-64bit --udp --interval 1 --omit 2 --reverse

Connecting to host 202.254.1.4, port 19750
Reverse mode, remote host 202.254.1.4 is sending
[  5] local 192.168.1.100 port 53356 connected to 202.254.1.4 port 19750
[ ID] Interval           Transfer     Bandwidth       Jitter    Lost/Total Datagrams
[  5]   0.00-1.00   sec  47.1 MBytes   395 Mbits/sec  0.028 ms  85757/119304 (72%)  (omitted)
[  5]   1.00-2.00   sec  90.0 MBytes   755 Mbits/sec  0.011 ms  18323/82054 (22%)  (omitted)
[  5]   0.00-1.00   sec   114 MBytes   955 Mbits/sec  0.153 ms  -104080/81068 (-1.3e+02%)  
[  5]   1.00-2.00   sec   112 MBytes   941 Mbits/sec  0.025 ms  1140/81059 (1.4%)  
[  5]   2.00-3.00   sec   114 MBytes   956 Mbits/sec  0.012 ms  62/81276 (0.076%)  
[  5]   3.00-4.00   sec   114 MBytes   957 Mbits/sec  0.025 ms  0/81274 (0%)  
```

bmah888:
Merged to 3.1-STABLE branch.

PR #414, opened 2016-05-16
Add fix for #412

joemce:
This prevents negative loss counters with UDP when omit is used. Instead of clearing cnt_errors and outoforder_packets when the omit interval expires, this approach uses omitted_cnt_errors and omitted_outoforder_packets to track what should be excluded. This is the approach already used by the packet_count.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -152,7 +152,9 @@ struct iperf_stream
     double    jitter;
     double    prev_transit;
     int       outoforder_packets;
+    int       omitted_outoforder_packets;
     int       cnt_error;
+    int       omitted_cnt_error;
     uint64_t  target;
 
     struct sockaddr_storage local_addr;
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -87,6 +87,7 @@ struct iperf_stream_result
     iperf_size_t bytes_sent;
     iperf_size_t bytes_received_this_interval;
     iperf_size_t bytes_sent_this_interval;
+    iperf_size_t bytes_sent_omit;
     int stream_prev_total_retrans;
     int stream_retrans;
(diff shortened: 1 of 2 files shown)


joemce:
I just realized that my last pull request included changes to the bandwidth calculation that should not be included. I was experimenting with displaying received bandwidth instead of sent bandwidth.  These chunks should not be included. I'll remove them and update the pull request.

https://github.com/esnet/iperf/pull/414/commits/1f240f31f7d8a8c69f0bcbc4d8e058d6e459af0c#diff-109ce0347bb7c189bead5f3b4da4f468R2342

bmah888:
Thanks for the bug report and pull request!  I halfway understand what's going on.  Is the problem that the sender and receiver have different ideas (due to network jitter and/or clock skew) about when the omit interval finishes, and therefore the receiver sees a large backwards sequence number decrement as out-of-order and/or lost packets?

bmah888:
Just going to mention issue #412 to get cross-references in the GitHub issue tracker to work correctly.

joemce:
I think the root issue is that the stream outoforder and cnt_error packets are getting set to 0 when the omit timer fires. 

``` text
	sp->outoforder_packets = 0;
 	sp->cnt_error = 0;
```

However, when the errors per interval are then computed, the values are compared to the previous interval. This results in the negative values. The packet_count already uses an omit variable to subtract out the different. This patch just extents that same approach to the outoforder and cnt_error counters.

bmah888:
I've just merged this pull request to master.  I'd like to do some more testing on this before merging to the 3.1 release branch, but I do plan to do that at some point.  Thanks!

bmah888:
Just noting that 69b8efa merged this change to 3.1-STABLE.

PR #423, opened 2016-05-25
DEREF_AFTER_NULL in src/iperf_error.c

g-coder:
DEREF_AFTER_NULL: pointer ‘test’ at line:77 is passed as an argument to function iperf_delete_pidfile(), in which it is dereferenced at iperf_api.c:2832.
Pointer ‘test’ can be NULL and dereferencing a NULL pointer causes seg-fault.

Applied Fix: pointer ‘test’ is checked for NULL before passing it to function iperf_delete_pidfile().


bmah888:
Committed, thanks!

bmah888:
Merged to 3.1-STABLE branch.

PR #425, opened 2016-05-26
Safe programming practice. Added va_end.

g-coder:
In file iperf_util.c:
Function 'va_start' is called at line:327. But, 'va_end' is not called before returning from function 'iperf_json_printf()' at line:352 and line:355. 
The va_end performs cleanup for the argp object initialized by a call to va_start. If va_end is not called before a function that calls va_start returns, the behavior is undefined.

Applied Fix: added va_end before returning from the function.


bmah888:
Committed, thanks!

bmah888:
Merged to 3.1-STABLE branch.

PR #429, opened 2016-05-31
Double close in iperf_server_api.c

g-coder:
In file iperf_server_api.c:
Handler 'test->listener' is closed at line:547 by calling close(). But, it is double closed at line:414 by calling function cleanup_server() at line:548.
It causes double close of the handler.

Applied Fix: Placed 'close(test->listener);' after the 'if block'. It will omit the condition of double close.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -544,12 +544,12 @@ iperf_run_server(struct iperf_test *test)
                     } else { 
                         if (test->no_delay || test->settings->mss || test->settings->socket_bufsize) {
                             FD_CLR(test->listener, &test->read_set);
-                            close(test->listener);
                             if ((s = netannounce(test->settings->domain, Ptcp, test->bind_address, test->server_port)) < 0) {
 				cleanup_server(test);
                                 i_errno = IELISTEN;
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for pointing out this problem.  Unfortunately your suggested solution doesn't work.  The close(2) call that you moved is necessary where it is, in order for a call to netannounce() to succeed.  If we apply your patch, then some code near the bottom of netannounce() tries to do a bind(2) and listen(2) to the TCP port that's still being used by test->listener.  This causes the server to fail.  You can reproduce this by running the client with some set of options that sets no-delay, changes the MSS, or changes the socket buffer (window) size, for example `src/iperf3 -c localhost --window 2m`.  This causes the following error message on the server (and the test to be aborted):
```
Server listening on 5201
Accepted connection from ::1, port 64735
[ 13] local ::1 port 5201 connected to ::1 port 64736
iperf3: error - unable to start listener for connections: Address already in use
```
I've got another way in mind to solve this.  Thanks for pointing out this problem!

bmah888:
Merged to 3.1-STABLE branch in 6694f65.

Issue #404, opened 2016-04-29
TCP connections on iperf3 server port causes iperf to exit


jcribben:
I installed iperf onto a server that runs in a virtual machine hosted by Microsoft Azure. We may have triggered some sort of security feature that caused them to sporadically start sending TCP packets on our iperf ports we designated. These tcp packets look like they go through iperf and are treated as new connections without a cookie and are then interpreted as errors. This wouldn't be a problem if iperf3 didn't quit after running into 5 consecutive errors, stopping our iperf server. I believe that these random connections shouldn't cause iperf to exit since they do not harm the ability to still run tests even if they are happening often.

Error msg:

> iperf3: error - unable to receive cookie at server: Connection reset by peer

... several times, then -

> too many errors, exiting

bmah888:
This problem is believe to be fixed as a side-effect of some security-related fixes in iperf-3.1.3 and iperf-3.0.12.  iperf3 should no longer exit as a result of five consecutive random TCP connections to its control port.

PR #441, opened 2016-07-11
Fix for #404

jcribben:
#404

added leniency to rogue connections while server is running in cases of mid test as well as no test running. Issue seemed to be caused by invalid cookies & closing the test socket when calling server cleanup. The changes I made here solved my issue.


bmah888:
This problem is believe to be fixed as a side-effect of some security-related fixes in iperf-3.1.3 and iperf-3.0.12.  iperf3 should no longer exit as a result of five consecutive random TCP connections to its control port.

Issue #378, opened 2016-03-07
timeout for TCP test


pjebs:
I don't think the TCP test has a timeout. Mine keeps going without stopping and it doesn't seem to be doing anything.

pjebs:
if there is a timeout, how can I reduce it?

rpodgorny:
+1

sgon00:
+1

bmah888:
I believe this was solved by pull request #446.  Keeping this issue open to note that that code change needs to be merged.

bmah888:
#446 merged to 3.1-STABLE.  Closing.

JayanthRR:
Does version 3.2 have this merge? I am facing a similar problem. When the client is not connected to the internet, I closed the client and reopened it after the internet connection is available. I receive an error saying that the server was busy doing another test.

bmah888:
Yes 3.2 has that fix.

PR #446, opened 2016-08-05
Close control socket and data stream sockets when client is disconnected

dmdailey:
The server will hang the control socket if the client is disconnected (pull the cable) in the middle of a test.  The stream sockets are also left open indefinitely.  Use patch from user mkall to close the control socket and add the closure of the data sockets as well when the test runs for more than 5 seconds over its duration.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -312,6 +312,26 @@ iperf_test_reset(struct iperf_test *test)
 }
 
 static void
+server_timer_proc(TimerClientData client_data, struct timeval *nowP)
+{
+    struct iperf_test *test = client_data.p;
+    struct iperf_stream *sp;
+
+    test->timer = NULL;
+    if (test->done)
+        return;
+    test->done = 1;
+    /* Free streams */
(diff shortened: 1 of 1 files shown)


rpodgorny:
i'm also being hit by this. can you please merge? thank you...

bmah888:
Thanks for the pull request!  I need to go look at this.  The problem where the client just disappears is definitely one that the server needs to handle better than it does.

bmah888:
This PR might also solve #378.

bmah888:
Note to self:  To test this functionality, begin running a test and then suspend the client so it can't finish.

bmah888:
Merged to 3.1-STABLE.

inesmll:
I have installed version 3.7 on both client and server machines and the problem persists, both with TCP and UDP connections. The server just won't stop and/or restart when the client disconnects... I'm breaking the connection *before* the test duration ends and I'm using a Alpine machine (client) and a OpenWrt machine (server). I have not found any information about a possible server timeout in the iperf documentation. Am I missing something, or do other people have the same problem?

PR #459, opened 2016-09-20
Properly setting the no_fq_socket_pacing option when not compiled in

woody77:
When HAVE_SO_MAX_PACING_RATE is not defined, the test param no_fq_socket_pacing is left at 0, which implies that it's enabled (and may trip up code elsewhere that's looking at that flag, but still needs to be included in the codebase if HAVE_SO_MAX_PACING_RATE is not defined).


bmah888:
Merged to master, thanks!  Merge candidate to the 3.1-STABLE branch.

woody77:
No problem!

bmah888:
Merged ad60eee to 3.1-STABLE.

PR #460, opened 2016-09-20
Change pacing timer from 100ms to 1ms

woody77:
This changes the hardcoded timer used for pacing tcp transmission on non-fq systems and udp on all systems to a much higher-resolution.  This greatly reduces the size of the bursts, to smoothen out the transmission of packets, to reduce buffer overflows on the network path.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1200,8 +1200,8 @@ iperf_create_send_timers(struct iperf_test * test)
         sp->green_light = 1;
 	if (test->settings->rate != 0) {
 	    cd.p = sp;
-	    sp->send_timer = tmr_create((struct timeval*) 0, send_timer_proc, cd, 100000L, 1);
-	    /* (Repeat every tenth second - arbitrary often value.) */
+	    /* (Repeat every millisecond - arbitrary value to provide smooth pacing.) */
+	    sp->send_timer = tmr_create((struct timeval*) 0, send_timer_proc, cd, 1000L, 1);
 	    if (sp->send_timer == NULL) {
 		i_errno = IEINITTEST;
 		return -1;


woody77:
Note that this requires the previous pull request (#459) in order to work correctly, due to this line in iperf_send:

https://github.com/esnet/iperf/blob/master/src/iperf_api.c#L1093

Since without #459 no_fq_socket_pacing is always 0 on platforms that don't support fq pacing, which causes the main loop to transmit a packet on every flow, each time it calls iperf_send() (which at 10x a second was probably never noticed, but at 1000x a second, get's noticed at lower rates).

bmah888:
For this part I'm trying to decide if we want the timer to fire every millisecond or slightly less often (definitely more often than the 100ms of 3.1.3 and older).  Specifically I'm wondering about the added CPU overhead that comes from the timer firing 100x more often than it was originally.

I'm going to merge this as-is to master for now, and maybe based on experience with that, we can see if we need some other "arbitrary" value or we need to make it a command-line parameter or...???...  Once we figure that out we should put it in 3.1.x.

woody77:
My next testing is going to be using a timer rate based on the PPS needed
to pace the packets, but that ends up at a <1ms timer at rates over 12Mbps,
so I think it would need to be capped at a maximum rate.

That maximum rate cap would be a good candidate for being a command-line
parameter.

-Aaron

woody77:
Branch for that is here:  https://github.com/woody77/iperf/tree/pacing_timer

It seems to be working very nicely (and works well even with timers down
into 10s of microseconds).

-Aaron

bms:
Re Windows: It's been a while since I looked at hires timers on Windows. It used to be that you had to call into the multimedia extension DLLs to the base Win32 API to get hi-res timers, but these days MS's standing advice seems to be to go to QueryPerformanceCounter() directly (TSC for uni-core, with HPET as a  for multi with unsynchronized TSC).

bmah888:
Just to mention that we're still thinking about this over here at ESnet, in particular about the tradeoff between smoother pacing of packets and higher CPU utilization, and even more specifically how those play out in the high-speed paths (10/40/100 Gbps) in our environment.  We'd like to do some experiments on some of our testbed hosts, although I think that a possible outcome of that work is that we end up making a command-line parameter, along with some guidance.

Admittedly, I haven't looked at the pacing_timer branch yet.

woody77:
I'd be interested to hear what you find with the higher speed paths.  I've
tested briefly with 10GbE, and didn't see onerous CPU usage, but I haven't
spent time comparing the CPU usage vs. the timer setting on that setup,
either.

btw, I've since moved that code to my main branch, since I've been very
happy with the results up to 1GbE (where most of our testing has been).

-Aaron

wangyoucao577:
Excuse me, could I ask what's the meaning of "non-fq system"? Thanks...

woody77:
fq_pacing is feature in the Linux kernel which paces out the the TCP
packets so that they don't go out in big bursts.

https://www.ietf.org/proceedings/88/slides/slides-88-tcpm-9.pdf

But it's only available on Linux from v3.10 (?) and later.  So on older
Linux kernels, on BSD, OSX, etc, fq_pacing isn't available.

-Aaron

bmah888:
Just a note that I haven't forgotten about this, but due to various issues (including physical relocation of part of our testbed network later this week), I haven't had a chance to do any tests on our high speed (100G) hosts yet.

woody77:
No worries.

I did find some interesting behavior this last week, though, with the
fq_pacing vs. the application pacing.

The application pacing is at the tcp/udp payload layer.  Whereas fq_pacing
appears to be at L2 (or an estimate of L2 based on headers/mtu).  I see a
constant drop in the achieved rate as reported by iPerf with fq_pacing
enabled vs. disabled.

-Aaron

woody77:
@bmah888 I was looking back at this, and realized that the pr's for #459 and #460 are still only in master, not in the 3.1 release branch (I'd thought that you were going to pull those into the release branch using either a 1ms-accurate or a 10ms-accurate timer).

bmah888:
The delay is partly due to me being ultra-conservative about not wanting to impact performance by firing the timers too often.  I really want to do some measurements on this code before merging it...couple of people in my group were also interested in it as well.  Alas, there are too many interesting things to work on!

Thanks for the reminder...the plan is still to merge these at some point.

Issue #500, opened 2017-01-19
`--title` parameter is not reset after each test


jedix:
If the client sets a title using `--title "some title"`, this title is included in the server output for each subsequent request, as long as a client does not specify a new title.

To reproduce:
```
# start server
$ iperf3 -s -J -p 5201 --logfile /tmp/iperf3_5201.log -s -d

# Perform test with title option set
$ iperf3 -c 127.0.0.1 -p 5201 -t 2 --title "First test"

# Perform test without title option set
$ iperf3 -c 127.0.0.1 -p 5201 -t 2

# Perform test with different title set
$ iperf3 -c 127.0.0.1 -p 5201 -t 2 --title "Third test"

# Look at the log file
$ grep title /tmp/iperf3_5201.log 
	"title":	"First test"
	"title":	"First test"
	"title":	"Third test"
```

iperf3 version: 3.1.3, tested on Fedora and on OSX

bmah888:
Reopening to remind me to merge this to 3.1.

bmah888:
Finally merged this.  Closing.

PR #501, opened 2017-01-19
Fix title not being reset on test reset

slankes:
Also reset test->title in iperf_reset_test to make sure a successive testrun doesn't inherit the pervious testruns title.

This fixes bug #500 - tested on Centos 7.


bmah888:
Thanks for sending this!  It looks about right to me, although I'm completely ENOTIME for real testing on my part, merging the change, etc.  When I get some iperf3 time I'll look at it some more.

bmah888:
There's an issue here, in that we want to free(3) the allocated string before setting it to NULL.  Other than that, this works.  I'm going to close this pull request and commit another version of this change.  Thanks for finding this bug and doing the debugging!

PR #508, opened 2017-01-26
add support for specifying --dscp symbolically and numerically

dtaht:
Using a command line adding dscp (instead of tos) you can:

--dscp EF,CS1,etc.
--dscp 0x08
--dscp 63

These will provide the correct shifted left 2 tos value for these, and
for people that think in terms of dscp values, this is a goodness.

Having this option available lets an enduser clearly distinguish between
an old version of iperf with a non-working --tos facility, vs a
version where it works, with something saner that lets just specify
the dscp.

I did not come up with a good -? option for it, and used -5 internally.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -25,6 +25,7 @@ libiperf_la_SOURCES     = \
 	                iperf_sctp.h \
                         iperf_util.c \
                         iperf_util.h \
+			dscp.c \
                         net.c \
                         net.h \
 			portable_endian.h \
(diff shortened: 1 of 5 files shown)


dtaht:
the failing check is due to your build system not renerating autoconf.

bmah888:
Thanks for the pull request!  One question...are you saying that the `--tos` option doesn't work at all?  We actually don't use it in our main use case, so I don't have any first-hand experience with it one way or the other.

PR #517, opened 2017-02-25
Service Authentication

ralcini:
New pull request, resuming https://github.com/esnet/iperf/pull/242 with new approach

I'm thinking about that and, if I'm not wrong, the introduction of SSL on the control channel is not so easy. What about an alternative approach based on RSA encryption of a message containing the user credentials and a timestamp? I try to better explain my idea:

 * the iperf client accepts as input via command line the credentials and the file path to the RSA public key
* using the same approach seen in this pull request, the iperf client sends the encripted message to the server
* the server decrypts the message, validates the credentials contained and checks that the timestamp contained is near to the current server time (e.g.: +/- 30 seconds).

Requirements:
* dependency to openssl library
* generation of a RSA keypair to setup the service
* admins must send the public key to the final users with the assigned credentials


ralcini:
For CygWin users the binary created needs the following DLLs:
 * cygz.dll
 * cygcrypto-1.0.0.dll
 * cygwin1.dll 

Thank you BudMan@neoforum!

bmah888:
I just wanted to thank and acknowledge your pull request.  I need to make some time to dig into this.  I see some commits for CygWin compatibility...just to confirm, this change works on UNIX / Linux hosts, right?  Thanks again!

ralcini:
Hi bmah888, 
yes, the new functionality is intended to work both on *unix systems both under CygWin.
Thank you!

TheRealDJ:
Pulled down the latest. Under Linux (CentOS 6.3) still having issues.

Server Output:
# ./iperf3 -s --rsa-private-key-path priv.key --authorized-users-path users.csv -d -1
Server listening on 5201
get_parameters:
{
	"tcp":	true,
	"omit":	0,
	"time":	10,
	"parallel":	1,
	"len":	131072,
	"authtoken":	"XoqcqQut6TvwvoIhKTNE7WIESqwIGE+4WN/L3Nfwo5lcsvFccWuLJ3SYgF0cVF134mPpCoubwltYuRzkGAInH9BdiTax+qHqNfgR+rUqR6dpqpd9pecwNyKzGNbollE5P5+MSDBUFdy2zRl18qoLTYP0cYj3z0jzKZcLdzE7Fck=",
	"client_version":	"3-CURRENT"
}
Auth Token Content:
user: dj
pwd:  dj
ts:   1488550709
Auth Token Credentials:
--> dj dj
Authentication failed for user 'ff00aea778d5cf5ea4d7abdfc841d601' ts 1488550709
iperf3: error - test authorization failed

Client Output:
./iperf3 -c localhost --rsa-public-key-path pub.key --username dj -d
Password: send_parameters:
{
	"tcp":	true,
	"omit":	0,
	"time":	10,
	"parallel":	1,
	"len":	131072,
	"authtoken":	"XoqcqQut6TvwvoIhKTNE7WIESqwIGE+4WN/L3Nfwo5lcsvFccWuLJ3SYgF0cVF134mPpCoubwltYuRzkGAInH9BdiTax+qHqNfgR+rUqR6dpqpd9pecwNyKzGNbollE5P5+MSDBUFdy2zRl18qoLTYP0cYj3z0jzKZcLdzE7Fck=",
	"client_version":	"3-CURRENT"
}
Connecting to host localhost, port 5201
iperf3: error - test authorization failed

Same shows on the MAC, but the error message on the server side on the MAC has username blank:
./iperf3 -s --rsa-private-key-path server_key.pem --authorized-users-path auth_users.csv -d -1
Server listening on 5201
get_parameters:
{
	"tcp":	true,
	"omit":	0,
	"time":	10,
	"parallel":	1,
	"len":	131072,
	"authtoken":	"D9fbL+2ooIiNpvpej6NuUKXu4MWbvtIjxWzoe6KIaVgYvCo7JFT/XSWZ/r/wgOCfnmPpHDGobV9OuP3gPip/kHOIt60/mF6bXItOYQn/V4gBcLZk2ViSjRMukry3D+B3RT2hje2EiiPXPdslKbNMOGDAqNAvkkn3RJRd1ReihvP+q2THEzZZ9PY4yW/ROp+PbddCNtVyqwZVrOALx7HWYoz0CpM0du3SWftMioilFQC5dwVOrzdZKKWeJaOjAt1PJ+OlqvTU8T29oodZqVKXsDBrtvYMlrJsBaPrXW2N2ShUrLfAgVnhMtd9VrKLC/TdBqyrLLG/Y9vIsaqyY3ODHdOJwAb+7QYMFLSAbnjlhxZ6dn1fsnZg0xXcXKE1YADypKIvL2mv5ybzPMfZwqC7HzOwKTWydXc+BqzEoM0hWbvWuRy6IJ/ExRHaCGqnSVBwPlFn8Emouj77nBTvC36aU+hxoHHQspqBOUkuSzqG6umBtY7yqpkNmImoFzzmvghqLngWZ86w8+P/HlN3rLjaqXSnIeM5cKQIIppxp2sl74tIJ5PjFTNedeF+cdFbzXzuFtpLcLGen92SHTldtqUVQeH4SsE/Turz5RJ4Yx+tgUEBmv/6bWP8B+9T7zl9OpLWkqV+aGSMKTdbeuvxFEQYGrL3qPAMBEqXQX1kLreZgAk=",
	"client_version":	"3-CURRENT"
}
Auth Token Content:
user: dj
pwd:  dj
ts:   1488550824
Auth Token Credentials:
--> dj dj
Authentication failed for user '' ts 1488550824
iperf3: error - test authorization failed

Thanks.
DJ

ralcini:
@TheRealDJ I reproduced it on centos distro and the problem was related to the initialization of two strings. Can you confirm that the problem is solved for you too?

TheRealDJ:
The username is now populated, but auth is still failing.

Server:
./iperf3 -s --rsa-private-key-path server.pem --authorized-users-path auth_users.csv -d
Server listening on 5201
get_parameters:
{
	"tcp":	true,
	"omit":	0,
	"time":	10,
	"parallel":	1,
	"len":	131072,
	"authtoken":	"IAno/ERYvueF11bFYHF5hp/kXmlgtAUpo3uRq9w52BWFz4L1+aeknYqb53ekN76xP9zinUmQ32h35V9V1Pu/nlHM8FaypHe/SfOoeTrCiEIqZPRBDqicwaDu19gVEMgNkn6dVoEgtyjMwqNVvYA8Fd4f4r3MAr3ZJ8BZAwqdWIBg9fYROf4WYd9LtGC0+uDBDUI7Barhq0Pj8F2ht76fxIJOe+3Yapbx/uzstPBB5tjWNZ9IcmZKeqwP6j53ra7991Ka0psFAUnaGq9juQMiPOnWsTDNGFsfPGIhLgsmCgCY9NFtufnxmDISfX/54aGAg9MfwBuBTbDz8XbxB4R9QSQyaSL+A3eIJSZBwL4JywtbMJ7sEOL/rcbg3TGIxwsA5+tebnbLRhSV8ahLcir0nnYzaCMhsX4qdzHyqrJeWUGP50DK9IQrONiWEzTNd249zyY2aQ+4WoDs0wieZ2EZF4pRrbMsPNdr903q2arxubXeGz9n5wZ8PD1aphT3rcokMXFgOh0DjVbjbsoS4R61BfUpa0y1X2j/kGtffKYDaDWK7bCNm6rCAjI+6c+5oAAEEPbhcq8so81j5IaafPMd2XVTYARANK2tBcfWp46WcDtWvvgcys3jrfg+97OHnth7Q5IyajkWKSRHmdAWpU+xi+ly800WMcFpu0APsrbL5yw=",
	"client_version":	"3-CURRENT"
}
Auth Token Content:
user: dj
pwd:  dj
ts:   1488903456
Auth Token Credentials:
--> dj dj
Authentication failed for user 'dj' ts 1488903456
iperf3: error - test authorization failed

# cat auth_users.csv 
testuser,testpass
dj,dj

Client:
./iperf3 -c 192.168.1.111 --rsa-public-key-path server.pub --username dj
Password: Connecting to host 192.168.1.111, port 5201
iperf3: error - test authorization failed

ralcini:
@TheRealDJ in the credentials file there is the hash of selected password (not in plaintext). In the manpage there is a complete example. Let me know if you have any suggestion on documentation enhancement.

TheRealDJ:
Good news! Once I CORRECTLY hashed the {username}password, everything started working! Tested under CentOS 6.4 and MacOS 10.10.2

Any additional testing you would like me to do? I'll test it on both systems once merged as well.

Awesome work!!!

ralcini:
thank you @TheRealDJ ! 
@bmah888, what do you think? Have you had some time to review my pull request?

bmah888:
@ralcini:  I'm looking at it in my Copious Spare Time (tm).

- It looks like the manual page in `src/iperf3.1` needs to be updated with the new command-line options.  For new flags the general procedure is to update the manpage and then (post-release) update the documentation files in `docs/`.  In a perfect world you would have just updated the manpage and not the docs (which I would generate at the next release) but there's no way you could have known that.

- New files should have a license block at the top with a license that's compatible with the BSD license that we use for iperf3.  I mention this because unlike many other submissions to iperf3, yours is significant enough to (logically) require a couple of new source files (`src/iperf_auth.c` and `src/iperf_auth.h`).

- Does the password need to be input interactively?

- In your code you have C preprocessor directives that look like `# include <foo.h>`...whereas I'm used to seeing `#include <foo.h>` (no space after the hash sign).  I actually had to look up whether that was legal C (clearly it compiled), so I learned something new today.

I haven't gotten to the point of actually trying out the functionality yet, and I might have some other comments / questions after that.  But so far so good.  Honestly I'm happy to see that it doesn't seem to break existing use cases...that's always a concern of mine when adding new features.  I'll keep poking at this.

Thanks for working on this!

ralcini:
> It looks like the manual page in src/iperf3.1 needs to be updated with the new command-line options. For new flags the general procedure is to update the manpage and then (post-release) update the documentation files in docs/. In a perfect world you would have just updated the manpage and not the docs (which I would generate at the next release) but there's no way you could have known that.

I've updated the manpage in src/iperf3.1. Must I revert the changes already done on docs/*?

> New files should have a license block at the top with a license that's compatible with the BSD license that we use for iperf3. I mention this because unlike many other submissions to iperf3, yours is significant enough to (logically) require a couple of new source files (src/iperf_auth.c and src/iperf_auth.h).

I've added the standard license block.........

> Does the password need to be input interactively?

Yes, are there any other use case that you suggest to analyze?

> In your code you have C preprocessor directives that look like # include <foo.h>...whereas I'm used to seeing #include <foo.h> (no space after the hash sign). I actually had to look up whether that was legal C (clearly it compiled), so I learned something new today.

I've removed the extra space in order to mantain style coding of your project

bmah888:
Thanks...you're fast!

> I've updated the manpage in src/iperf3.1. Must I revert the changes already done on docs/*?

I'd say leave the `docs/*` stuff the way it is, just prepare for me to potentially overwrite those changes.  The documentation exists on all codelines but it's only the one on master that counts.

>> Does the password need to be input interactively?

> Yes, are there any other use case that you suggest to analyze?

I don't have a concrete use case, but I'm thinking about cases where iperf3 gets run by some other front-end program, say a scheduler, and might not be able to take a password interactively.  This pattern is not common in the Windows platform, but it's extremely common on UNIX-like systems.  iperf3 gets used as a part of the perfSONAR toolkit (and that's sort of the main use case for iperf3), in which it's run by a scheduling tool (either bwctl or pScheduler, depending on which version of perfSONAR).  Obviously this use case doesn't use authentication, but authentication would be precluded by an implementation that requires a human to type a password on the stdin stream.  Just something to think about.

ralcini:
>>>Does the password need to be input interactively?

>>Yes, are there any other use case that you suggest to analyze?

>I don't have a concrete use case, but I'm thinking about cases where iperf3 gets run by some other front-end program, say a scheduler, and might not be able to take a password interactively. This pattern is not common in the Windows platform, but it's extremely common on UNIX-like systems. iperf3 gets used as a part of the perfSONAR toolkit (and that's sort of the main use case for iperf3), in which it's run by a scheduling tool (either bwctl or pScheduler, depending on which version of perfSONAR). Obviously this use case doesn't use authentication, but authentication would be precluded by an implementation that requires a human to type a password on the stdin stream. Just something to think about.

I agree with you, batch execution needs an other method to input password. I'm thinking to a configuration file.. I think that it's better to wait an explicit request for improvement for batch execution.

bmah888:
OK, sounds fair to me.  I pulled the latest code.  Trying to run this, I generated an RSA keypair according to the instructions in the manpage.  Then created a credentials file.  Running the server like this:
```
src/iperf3 --server --rsa-private-key-path private_not_protected.pem --authorized-users-path credentials.csv
```

And the client like this:
```
src/iperf3 --username bmah --rsa-public-key-path public.pem --client localhost
```
I get a password prompt, put in the password that I'd hashed into `credentials.csv` earlier, but authentication fails.  Similar to what @TheRealDJ ran into earlier.  I'm trying to see what I might have done wrong...will report back with some more details if I can't get it working.

The manpage still mentions a `--password` option which I don't think is relevant anymore, is it?  I might want to do some other wordsmithing on the manpage after it gets merged.

Also, I'm wondering if it'd make sense to have the credentials file in some more de facto file format, such as the Apache htpasswd file (which in turn is a variation on the UNIX passwd file format).

bmah888:
OK.  I figured out the problem with `credentials.csv` after trying about 10 times to make this work.  The issue for me was that the method you give to generate a password hash only works on the bash shell, and I took me many tries to realize that.  I also got confused because the curly-brace characters you include as part of the string to be hashed look like shell metacharacters (in other words you wrote `{$S_USER}$S_PASSWORD` in your instructions, but my brain interpreted this as (initially) `${S_USER}$S_PASSWORD`.  Also I don't think the `-e` flag to the echo command is really needed (either for the executable or the bash built-in command).

Anyway, bottom line, I got this to work.

ralcini:
> The manpage still mentions a --password option which I don't think is relevant anymore, is it? I might want to do some other wordsmithing on the manpage after it gets merged.

I don't find any reference to `--password` in the manpage (iperf3.1 file). Are you referring to an other file? 

> Also,  I'm wondering if it'd make sense to have the credentials file in some more de facto file format, such as the Apache htpasswd file (which in turn is a variation on the UNIX passwd file format).

Interesting idea the apache htpasswd, but the only crypto method valuable in my optinion is bcrypt (https://httpd.apache.org/docs/2.4/misc/password_encryptions.html).. I will check the complexity to implement it.

> Also I don't think the -e flag to the echo command is really needed (either for the executable or the bash built-in command).

Right! It is a mistake, `-e` flag can be removed

ralcini:
@bmah888: in order to use password hashing compatible with htaccess files, I need to import apr library in your project if you agree. In order to keep things simple I propose to implement only bcrypt method.

Is it better to merge this pull request and open a new one for this enhancement?

bmah888:
Hrm.  I somehow thought this would be simpler than it is.  Why don't you hold off on any APR related changes for now.  The only reason I suggested doing either the htpasswd or UNIX passwd file format was that I wasn't super-keen on requiring a new authentication file format that's used (as far as I know) by no other program, and requiring a procedure to add users that's specific to the iperf3 implementation.  The passwd and htpasswd formats are well-known (at least by us UNIX types).  Anyway.  Doing one of these file formats almost sounds like it's more work than it's worth.  Let me think this over a bit and if I can't come up with anything else we'll just use what you've got.

TheRealDJ:
@bmah888, what if we create a script (something like iperf_passwd) that take the username and password from the user, and creates the line to be added to the creds file ... Would this be acceptable?

ralcini:
@bmah888: any new idea or proposal?

bmah888:
Sorry, haven't forgotten about you.  We're trying to prune down a lot of existing issues in the issue tracker.  Also we've got an impending perfSONAR release that's taking up some attention.

I was going to ask about replay attacks, but it looks because the tokens have timestamps that are checked to be within 30 seconds, that reduces the window of vulnerability.  I guess we could live with that.  That's the price we pay for not having an entirely encrypted control connection.  Did I get that right?

The timestamp checking adds a requirement for synchronized clocks (but only if using authentication).  That's probably reasonable.

ralcini:
Hi @bmah888 , yes replay attacks are avoided with timestamp validation and it require time syncronization between client & server (but I think that it is not a probleam today with the use of ntp servers). A similar technique is used with time-limited hmac tokens. 

If an SSL/TLS channel was available, I would have choosen a simpler approach..

bmah888:
D'oh.  I was all set to merge this pull request but some changes I did on master (merged pull request #508) now conflict with it.  Nobody's fault...I think I can fix this pretty easily.

ralcini:
Thank you all!

bmah888:
Thanks **you** @ralcini for your submission!

kathampy:
I have concerns about running an iPerf server in a sensitive environment.
Isn't it better to encrypt the credentials with an ephemeral key instead of padding them with a timestamp? DHE could be used directly since the server's public key is known.

Issue #503, opened 2017-01-23
./configure says nanosleep not required, then quits because of it


ShaunCurrier:
While running `./configure --host=i686-w64-mingw32`, the following is printed to console and it quits early:
```
<snip>
checking for library containing nanosleep... none required
<snip>
nanosleep() required for timing operations.
```

So, it says a nanosleep library is not required. Then says it is required.  Then it quits.

bmah888:
mingw is not an explicitly supported platform.  That having been said I'm not sure why it's behaving the way it is.  What version of iperf3 are you trying to build?

scurrier03:
I have `iperf-3.1.4`, which I got from the following link.
http://downloads.es.net/pub/iperf/iperf-3-current.tar.gz

scurrier03:
Related thread in the mingw-w64 discussion forum.
https://sourceforge.net/p/mingw-w64/discussion/723798/thread/7f8ca21d/?limit=25

scurrier03:
I've never dealt with shell scripts like `configure` before, but I noticed something that seems odd.  There are two places in the script where the error message `nanosleep() required for timing operations` appears.

The first occurrence, it's in a block of code that seems to be heavily related to `nanosleep` because it's surrounded by variables named after `nanosleep` and the command at the top of the block is `# Solaris puts nanosleep in -lrt`.  This seems totally normal.

The second occurrence, it's in a block of code that is similar to the first, but has replaced `nanosleep` with `hstrerror` - all except for the line where a reference to `nanosleep` remains: `echo "nanosleep() required for timing operations."`.

In my novice view, it looks like some code was copied and all references to `nanosleep` replaced, but the error message mentioning `nanosleep` was not replaced.

I see the two occurrences of the text `nanosleep() required for timing operations` on line 11827 and 11890 of the `configure` file.  This file doesn't appear in the repo.  I'm guessing it's generated before being offered at the link I referred to above?

I'm not sure how exactly it relates, but perhaps this ultimately traces back to [line 74](https://github.com/esnet/iperf/blob/01fb3e6dcfe91d5870c9b2187d1deff7aca95341/configure.ac#L74) in [configure.ac](https://github.com/esnet/iperf/blob/01fb3e6dcfe91d5870c9b2187d1deff7aca95341/configure.ac).

So maybe this is all just a case of a copy/paste error leading to an incorrect error message?

Borim7:
[...]

PR #523, opened 2017-03-07
Remove unused hstrerror(), bad nanosleep() message in configure.ac (#503)

ShaunCurrier:
Per comments of #503, `hstrerror()` is not used, but is preventing `./configure --host=i686-w64-mingw32` command from finishing.

### DIFF
--- a/configure
+++ b/configure
@@ -12311,69 +12311,6 @@ exit 1
 fi
 
 
-# Solaris puts hstrerror in -lresolv
-{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for library containing hstrerror" >&5
-$as_echo_n "checking for library containing hstrerror... " >&6; }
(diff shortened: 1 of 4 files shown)


Borim7:
Just removing hstrerror() from the configure file is not enough. You also have to remove the dead code.

ShaunCurrier:
Whoops, I misunderstood your comment before and didn't know enough to know better, but it makes sense now.  I will remove the dead code.  Thanks for the help and sorry for the confusion.

ShaunCurrier:
I took another stab at it by removing the dead code.  It builds fine for Linux and the tests in `test_commands.sh` work in Ubuntu, so it doesn't appear that I broke anything.  But please take a look as this is not my strong suit.

bmah888:
Thanks!  Merged to mainline.  Should be merged to 3.1-STABLE branch, will track that in #503.

Issue #525, opened 2017-03-08
add tcpi_rttvar from tcp_info to JSON output


bltierney:
tcpi_rttvar should be useful to determine if the path is congested or under-buffered.

It would be good to include that in the output.

bmah888:
Reopening to make sure I merge the fix for this.  Set target milestone.

bmah888:
Merged to 3.1.  Closing.

PR #534, opened 2017-03-24
add tcp rttvar to stream info

hoang-tranviet:
This commit adds tcp_info rttvar to JSON output, as discussed in #525


hoang-tranviet:
There are tcp_info stuffs in build_tcpinfo_message(), but I could not find any reference/usage of this function so I left them untouched.

bmah888:
Thanks for the pull request!  We need to do a little testing on this...but just by reading the code I'd say it looks good so far.

bmah888:
Merged to master (and eventually to 3.1-stable).  Thanks again!

PR #538, opened 2017-03-30
3.1 stable

harshascb:


bmah888:
I'm going to go out on a limb and say that this pull request, which looks like it merges all the changes on master to the 3.1-stable branch, is a mistake.  In any case we're not going to do that.

Issue #249, opened 2015-03-06
Report actual send window as well?


he32:
Hi,

I wonder ... would it be interesting to report the actual send window in use, i.e. the
one the receiver announces to the sender?  The problem is that the default display
on platforms doing TCP_INFO is quite wide, so there's no room for an additional coloumn,
within 80 characters, ref.

```
[ ID] Interval           Transfer     Bandwidth       Retr  Cwnd
[  6]   0.00-1.00   sec  3.46 MBytes  29.0 Mbits/sec    0    512 KBytes       
[  6]   1.00-2.01   sec  3.38 MBytes  28.2 Mbits/sec    0    512 KBytes       
[  6]   2.01-3.00   sec  3.44 MBytes  29.0 Mbits/sec    0    512 KBytes       
[  6]   3.00-4.00   sec  3.46 MBytes  29.0 Mbits/sec    0    512 KBytes       
```

I have some possibly hackish changes which changes this into

```
[ ID] Interval        Transfer  Bandwidth   Retr  Cwnd      Swnd
[  6]   0.00-1.00   s  3.45 MB  28.9 Mb/s    0    512 KB   32.5 KB   
[  6]   1.00-2.00   s  3.44 MB  28.8 Mb/s    0    512 KB   31.1 KB   
[  6]   2.00-3.00   s  3.46 MB  29.1 Mb/s    0    512 KB   32.5 KB   
[  6]   3.00-4.00   s  3.45 MB  28.9 Mb/s    0    512 KB   31.1 KB   
```

(and here we can see that it's the receiver's window which is limiting the performance.)

If someone wants take a peek, I can create a branch in my forked repository and commit
the change there for inspection.

Regards,

Håvard

bltierney:
check and see if this is in the JSON output. If so, that should be good enough, as not that many folks will need this.

bmah888:
We don't seem to have this in JSON, and I think this is probably the right place to put that information.

Just so I understand, the variable we're talking about is `tcpi_rcv_space` in `struct tcp_info`?

he32:
> We don't seem to have this in JSON, and I think this is
> probably the right place to put that information.
> 
> Just so I understand, the variable we're talking about is
> `tcpi_rcv_space` in `struct tcp_info`?

My diff uses `tcpi_snd_wnd` from `struct tcp_info`, and since the
Linux API apparently expresses this in number of segments, you
need to multiply by `tcpi_snd_mss` to get the window in bytes.

[...]

PR #542, opened 2017-04-01
Add printing of the actual send window.

he32:
To fit within 80 coloumns, shorten the printed units.
This is a suggested fix for issue #249.


src/units.c
```diff
@@ -330,6 +331,26 @@ extern    "C"
...
+    void short_unit_snprintf(char *s, int inLen,
+                               double inNum, char inFormat)
+    {
+       char *cp;
+
+       unit_snprintf(s, inLen, inNum, inFormat);
+       if ((cp = index(s, 'i')) != NULL) {
+           *cp = 0;
+       }
+       if ((cp = index(s, 'y')) != NULL) {
```
pprindeville: Should this be an `else if` instead?

pprindeville:
Is this PR being maintained?  Can it be rebased?

src/units.c
he32: Could be, as a minor optimization.

he32:
I just updated/merged what is in my branch.
I just got the impression there was little or no interest since the "send window"
information isn't available on Linux(?)

bmah888:
I am going to, somewhat reluctantly, close this pull request as-is.  Thanks for working on this and submitting it.

There are three reasons for not taking it:

1.  Usefulness of the semantics of `tcp_info.tcpi_snd_wnd`.  I think that when we looked into this, it was the window advertisement that was being advertised to the other side, so the number on the client side only makes sense in the --reverse case where the server is sending.  At this point I don't think we know how to get the number originally requested in #249.

2.  We'd prefer not to make any changes to the human-readable output at this point, and only have the window size output, if we can get it, in the JSON.

3.  Not working on Linux (or mac OS), thus limiting its applicability.  Of our three main development platforms, this only worked on FreeBSD.

PR #556, opened 2017-04-26
fix invalid sizeof on pointer instead of type

GabrielGanne:
This is only a coherency fix since sizeof(char*) >> sizeof(char)
There should be no functional nor stability impact whatsoever


bmah888:
Merged, thanks for the pull request!

PR #551, opened 2017-04-17
warning fixes

GabrielGanne:
The first commit removes some generated files, because it bothered me for working, but you _might_ not want to merge it.

The other commits are mainly stab and warning fixes. They should not change iperf's behavior in any way.

PR #564, opened 2017-05-04
Fix a heap corruption issue leading to a server-side crash.

bmah888:
Commit 5ab2132c (PR #551) fixed, among other things, a memory
leak.  The solution, however, causes a hazard where a free() of
an invalid pointer can corrupt the heap.  We've observed this
fairly repeatably while running the test_commands.sh script on
CentOS 7.

To remedy this, we NULL out a pointer after the object it
pointed to has been free-d, just like a number of other similar
objects.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -440,6 +440,7 @@ cleanup_server(struct iperf_test *test)
     }
     if (test->congestion_used != NULL) {
         free(test->congestion_used);
+	test->congestion_used = NULL;
     }
     if (test->timer != NULL) {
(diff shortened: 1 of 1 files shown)


bmah888:
Just for kicks I tried the `test-commands.sh` script on FreeBSD 11.0, with `MALLOC_OPTIONS` set to `junk:true`.  As expected, a bunch of runs fell over (because iperf3 built on FreeBSD doesn't support all the features that Linux does).  I saw some weird server-side behavior with master branch code, but things seemed to behave much more normally with the patch in this PR.

PR #577, opened 2017-05-12
fix free on uninitialized var found by clang

GabrielGanne:
the two calls to freeaddrinfo(res) are made on an uninitialized var
if you do not pass the test on socket options

### DIFF
--- a/src/iperf_tcp.c
+++ b/src/iperf_tcp.c
@@ -143,6 +143,7 @@ iperf_tcp_listen(struct iperf_test *test)
     int saved_errno;
     int rcvbuf_actual, sndbuf_actual;
 
+    res = NULL;
     s = test->listener;
 
     /*


bmah888:
Ah, I see the problem here but I'm going to fix it differently.  There are actually multiple calls of the form `freeaddrinfo(res)` in `tcp_listen()`...I'm assuming you're referring to the two near the bottom of the function.  Really I should not have made the error handling for those two calls invoke freeaddrinfo().  In fact none of the variables involved in the call to getaddrinfo() need to be in scope outside of the block beginning with the line:
```
    if (test->no_delay || test->settings->mss || test->settings->socket_bufsize) {
```
so I'm going to move their declarations inside that block to prevent future problems of this sort.

Thanks for pointing this out!

Issue #576, opened 2017-05-12
Unnecessary machine identification in cookie


Safari77:
# Context

* Version of iperf3:
git

* Hardware:
x86_64 SandyBridge

* Operating system (and distribution, if any):
Fedora 25

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report

* Expected Behavior
No identifying information about the system is needed when doing a bandwidth test.

* Actual Behavior
Spy cookie is sent on each test, which includes hostname and system's time in microsecond precision.

* Steps to Reproduce
Just run it.

* Possible Solution
Cookie can be ascii-string  with entropy from `/dev/urandom` or `getrandom(2)` and equivalent system calls.

In addition to the cookie, also `sp->buffer` is initialized with random data. On slow systems reading from `/dev/urandom` might be, well, slow. I implemented the random stuff by reading first a seed, then filling the buffers with `arc4random_buf` (libressl-portable).  If you disagree, nice, code would be a lot less simpler.

I haven't published the code as of yet, but it's working.

bmah888:
I think the term "spy cookie" is a bit of an exaggeration.  But I'm doing a little research to try to figure out why we put this information in the cookie, instead of using, say, a random UUID.

How big of a problem is the initialization of `sp->buffer`?  We only do this once per stream at the start of the test.  I don't object in principle to initializing in a different way, but I'd like to not introduce any new dependencies into iperf3.

Safari77:
`MAX_BLOCKSIZE` is 1048576. Some systems have read speeds below 10 MiB/s from `/dev/urandom`... For example, my Android device OnePlus 3 has 12 MiB/s. So if it needs on some slow system for example 0.25s to init the buffer, I don't think that's very nice.

Using `srandom(time(NULL))` and `random` to fill the buffer also leaks info about the system time.
You wouldn't need to add a new dependency, if you include the needed module directly into iperf repo...

bmah888:
@Safari77 : OK after a bit of discussion, I learned that there's no particular reason, historically, for the hostname or timestamp to be in the cookie.  In fact the cookie appears to have been UUID-based originally (which I think we agree would make sense, among different alternatives) but it was removed for the sake of "portability".  I'm not sure what to make of that decision, which happened about five years ago.  An entropy-generated ASCII string, as you suggested, would be fine with me too.

[...]

PR #582, opened 2017-05-18
Return random ascii-string in make_cookie.

Safari77:

```
```

* Issues fixed (if any):
#576 

Having hostname and microsecond timestamp in the cookie is not
necessary.  Also fill test buffer with data from /dev/urandom
instead of using random().

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -3169,9 +3169,6 @@ iperf_new_stream(struct iperf_test *test, int s)
         free(sp);
         return NULL;
     }
-    srandom(time(NULL));
-    for (i = 0; i < test->settings->blksize; ++i)
-        sp->buffer[i] = random();
 
     /* Set socket */
     sp->socket = s;
@@ -3196,7 +3193,8 @@ iperf_new_stream(struct iperf_test *test, int s)
         sp->diskfile_fd = -1;
 
     /* Initialize stream */
-    if (iperf_init_stream(sp, test) < 0) {
+    if ((readentropy(sp->buffer, test->settings->blksize) < 0) ||
+        (iperf_init_stream(sp, test) < 0)) {
         close(sp->buffer_fd);
         munmap(sp->buffer, sp->test->settings->blksize);
         free(sp->result);
--- a/src/iperf_util.c
+++ b/src/iperf_util.c
@@ -45,6 +45,37 @@
 #include <errno.h>
 
 #include "cjson.h"
+#include "iperf.h"
+#include "iperf_api.h"
+
+/*
+ * Read entropy from /dev/urandom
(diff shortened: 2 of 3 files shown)


bmah888:
Thanks for the pull request!   Looks good so far, although I might not get to really play with it for a few days due to travel.

One question which I guess I should have asked earlier.  Do we think that all the various platforms that might be running iperf3 all support `/dev/urandom`?  The officially supported platforms (FreeBSD, Linux, and macOS) all do, and a quick round of Google searching and Wikipedia imply that pretty much everything in the iperf3 universe does (including Cygwin, iOS, and various other platforms that we don't officially support).  Just wanted to get some confirmation and a warm fuzzy feeling that this change won't be gratuitously breaking something.

Reference:  https://en.wikipedia.org/wiki//dev/random

Safari77:
For example, in Fedora Linux, packagers test the new versions before unleashing them to the public. Also, "make check" is often done; if it fails, rpm package does not get built (t_uuid test would catch if there is trouble accessing /dev/urandom).

Similarly, other OS'es should have similar testing implemented for the packages. I mean, it exits with error if /dev/urandom is not available—how hard can it be to miss that?

I'd also add note about /dev/urandom requirement in `RELEASE_NOTES`.

bmah888:
Well I wasn't so worried about pushing broken packages to users.  You're right that various distros' package-building infrastructure would catch that.  It's more like I didn't want to surprise any maintainers, so they're not saying, "Oh cr*p, Bruce merged something into master that I have to fix for my platform."  (Having maintained various softwares for FreeBSD myself, I've been in this situation.)  Anyway, I've kind of convinced myself this isn't likely to be a problem.

I ran with this change a couple times last night, looked good so far.  I'm planning to play with it a little more and then merging it when I get a little uninterrupted time.  Thanks!

bmah888:
I guess one last comment I have on the change is that with the transformation you are doing to printable characters you are basically getting 5 bits of entropy per random byte.  If you expand the set of characters in `rndchars`, you can easily get another bit of entropy per byte...specifically I'm thinking of the set of characters that's used in the base64 encoding (basically it's all upper and lower-case letters, all digits, "+" and "/", for a total of 64 characters, or 6 bits of entropy when you fold the random bytes into those characters).

I am not suggesting that you need to make this change.  It think the commit is fine as-is, because 1) we don't need cryptographically strong entropy in this case and 2) your code change probably (insert wavy hands) gives more entropy in the cookie string than there was to begin with.

bmah888:
Merged...thanks for the pull request!

(Ironically I'm merging this C language code change from the exhibit floor at PyCon.)

Safari77:
Yeah I was thinking that 185 bits of entropy is enough.

PR #587, opened 2017-05-25
C++ compatiblity

KennethWong15:

* Issues fixed (if any):

I would like to be able to use libiperf as an external library to my C++ application. 
During compilation, there are problems with iperf_api.h not being able to find FILE*
and uint64 because they aren't included in the header. 

Making the functions extern because it's not accessible as an API, as well as including 
stdio/stdint in iperf_api.h

### DIFF
--- a/src/iperf_api.h
+++ b/src/iperf_api.h
@@ -29,6 +29,14 @@
 
 #include <sys/time.h>
 #include <setjmp.h>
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the pull request!  This looks pretty good to me.  I intend to merge this as a part of the 3.2 release.

Issue #457, opened 2016-09-20
strange mismatch behavior UDP OUT OF ORDER 


bbordereau:
Hello, i have got a strange behavior
On productio network , i decide to test bandwidth and error. So i perform an end to end between a client and a server in DC.
When i perform test the iperf3 with the server as server ( iperf3 -s ) and client as client in UDP , with 10Mb/s everything is ok  but 100Mb/s is being worse and i see in the server side, a lot of messages as OUT OF ORDER 
I do not see errors or drops on my network 
When i reverse side, the lost is worts more than 72%
If i did the test with TCP , everything looks fine 
is anyone has got an idea , i would like to be sure that my issue is not iperf3 itself ?
the command for client i used : iperf3 -c <@ip server) - b 100M -u 
any help will be great ?
what does it mean out of order in iperf3, and what are the possible root cause 
Best regards
Bert

bmah888:
This means that the iperf3 server is receiving the UDP test packets in a different order than they were sent from the iperf3 client.  There could be a variety of explanations for this phenomenon, ranging from packets taking different paths through the network to bugs in network drivers.

More information would be required for someone to help you diagnose the problem.  If you take a packet trace (say using tcpdump) at the destination, you could actually decode the iperf3 sequence numbers in the packets and see if the packets had arrived intact, out of order, etc.

bltierney:
might there be a bonded link or bonded NIC? That often leads to out of order packets.

Also, please see if 3.1.6 fixes it.

AndrewJDR:
@bmah888 :

Given that UDP as a protocol makes no guarantee that packets will arrive in the same order they're sent, why do out of order packets get counted as a "lost" packets in the stats output?

For example, in the UDP test output below using the latest f1415a0d98b commit of iperf on both ends and using the command `iperf3 -R -u --get-server-output -b 4M -c <my_server_hostname>`, there are several "OUT OF ORDER" messages, then immediately afterward, those out of order packets are reported as lost packets (e.g. `2/342`). Why are they counted as lost? This is UDP, so out-of-order packets are something that can happen under totally normal circumstances.
```
Connecting to host xxx, port 5201
Reverse mode, remote host xxx is sending
[  4] local 192.168.1.110 port 56316 connected to yyy port 5201
[ ID] Interval           Transfer     Bandwidth       Jitter    Lost/Total Datagrams
[...]

PR #589, opened 2017-05-31
Fix out-of-order / loss counting issues with UDP (#457)

bmah888:
  `3.1-STABLE`) to which this pull request applies:  master

* Issues fixed (if any):  #457 

Try to compensate for gaps in sequence number space caused by packet reordering, so we don't falsely report packets as missing when they've just been reordered.

Adjust some of the debugging output for UDP to make a little more sense.


sadamchandra:
same issue seen iperf 2.0.10 and 2.0.12 ,  so is  this patch relevant for 2.0.10/2.0.12 ?

PR #591, opened 2017-06-05
fix server-side memory leaks

GabrielGanne:
this should fix leaks: 
* timer free list was not cleansed
* frandom file was not closed

Adding a function to close the frandom file might not be the best way to go. I did it to try and stay within the iperf_util context: it opens, so it should be the one to close.
Do tell me if you wish this to be handled another way.

test command:
```bash
valgrind --leak-check=full --show-leak-kinds=all --track-origins=yes \
         --read-var-info=yes -- ./src/.libs/iperf3 -s --json -1
```

Valgrind memcheck's trace:
```
72 bytes in 1 blocks are indirectly lost in loss record 1 of 3
   at 0x4C27AD1: malloc (vg_replace_malloc.c:299)
   by 0x4E552E5: tmr_create (timer.c:151)
   by 0x4E4FC58: create_server_timers (iperf_server_api.c:357)
   by 0x4E50BB9: iperf_run_server (iperf_server_api.c:631)
   by 0x401031: run (main.c:151)
   by 0x400EDD: main (main.c:107)
144 (72 direct, 72 indirect) bytes in 1 blocks are definitely lost in loss record 2 of 3
   at 0x4C27AD1: malloc (vg_replace_malloc.c:299)
   by 0x4E552E5: tmr_create (timer.c:151)
   by 0x4E4FCE7: create_server_timers (iperf_server_api.c:364)
   by 0x4E50BB9: iperf_run_server (iperf_server_api.c:631)
   by 0x401031: run (main.c:151)
   by 0x400EDD: main (main.c:107)
568 bytes in 1 blocks are still reachable in loss record 1 of 1
   at 0x4C27AD1: malloc (vg_replace_malloc.c:299)
   by 0x5A2654C: __fopen_internal (in /usr/lib64/libc-2.17.so)
   by 0x4E53351: readentropy (iperf_util.c:64)
   by 0x4E4ABCB: iperf_new_stream (iperf_api.c:3203)
   by 0x4E50701: iperf_run_server (iperf_server_api.c:573)
   by 0x401031: run (main.c:151)
   by 0x400EDD: main (main.c:107)
```

Signed-off-by: Gabriel Ganne <gabriel.ganne@enea.com>

### DIFF
--- a/src/iperf_util.c
+++ b/src/iperf_util.c
@@ -48,6 +48,7 @@
 #include "iperf.h"
 #include "iperf_api.h"
 
+static FILE *frandom = NULL;
 /*
  * Read entropy from /dev/urandom
  * Errors are fatal.
@@ -55,7 +56,6 @@
  */
 int readentropy(void *out, size_t outsize)
 {
-    static FILE *frandom;
     static const char rndfile[] = "/dev/urandom";
 
     if (!outsize) return 0;
@@ -74,6 +74,13 @@ int readentropy(void *out, size_t outsize)
                       feof(frandom) ? "EOF" : strerror(errno));
     }
     return 0;
+}
+void entropycleanup()
+{
+    if (frandom != NULL) {
+        fclose(frandom);
+        frandom = NULL;
+    }
 }
 
 
--- a/src/iperf_util.h
+++ b/src/iperf_util.h
@@ -31,7 +31,10 @@
 #include <sys/select.h>
 #include <stddef.h>
 
+extern FILE *frandom;
+
 int readentropy(void *out, size_t outsize);
+void entropycleanup();
 
 void make_cookie(char *);
 
--- a/src/main.c
+++ b/src/main.c
@@ -44,6 +44,7 @@
 
 #include "iperf.h"
 #include "iperf_api.h"
+#include "iperf_util.h"
 #include "units.h"
 #include "iperf_locale.h"
 #include "net.h"
@@ -159,6 +160,8 @@ run(struct iperf_test *test)
                 if (iperf_get_test_one_off(test))
                     break;
             }
+	    entropycleanup();
+	    tmr_destroy();
 	    iperf_delete_pidfile(test);
             break;
 	case 'c':


bmah888:
Thanks for the pull request...really appreciate your hunting down all of these various warnings and errors.  So I see there are basically two sets of changes here:

o Timer free list leaks.  My interpretation of the timer free list (which I didn't write) is that it's supposed to hold onto unused/deactivated timers to save the work of reallocating those again.  So it isn't really necessary to get rid of the remaining timers in `server_cleanup()`.  In fact, if we do that, that defeats the point of having the free list, because the next time that `iperf_run_server()` gets called, its timers will need to be reallocated from scratch, rather than just re-used from the free list.  (For the small number of timers used by iperf3, it might not make a huge difference one way or the other.)  I wonder if a better place to do `tmr_destroy()` might be near the bottom of `run()`.

o Similarly with the `frandom` cleanup...I don't think it needs to be cleaned up after every test (server-side)...I think the intent is to keep this file open for a long time (in the case of the server).  It's probably appropriate to do `entropycleanup()` closer to process exit time, near the bottom of `run()`...that also caused it to be explicitly closed on the client.

I'm not hugely worried about these memory leaks given that they're small and (probably) bounded...unless I'm very much mistaken (tell me if I am) these don't cause continual memory growth.

Adding @Safari77 to this discussion, as the author of the `readentropy()` function...your thoughts would be welcome as well.  Thanks all!

Safari77:
`exit` closes the stream, so I don't see the point, unless it is to silence valgrind.

GabrielGanne:
My understanding of these two leaks is that they indeed cannot grow.
Cleaning at the end of `iperf_run_server` or `run` seems to ends up the same, isn't it ?
If not, I'll move this as indicated.

I did have an iperf server that I left on a small device which happened to run out of memory (for no reason I could find), and so I tested iperf for leaks. Once I could see them, fixing was straightforward, and so I did.

I believe that silencing memcheck has value in itself even if, as you say, there is no stability issue here.

Safari77:
Maybe you want to make `entropycleanup` also do `frandom = NULL;` so that it's actually _cleanup_, not just _silence valgrind_.

Otherwise I have nothing against this pull request. When closing `frandom` manually instead of using `exit` it just releases the allocated memory one microsecond earlier, or something like that.

bmah888:
@GabrielGanne : 

> My understanding of these two leaks is that they indeed cannot grow.
> Cleaning at the end of iperf_run_server or run seems to ends up the same, isn't it ?
> If not, I'll move this as indicated.

If you do the cleanup at the end of `iperf_run_server()` then the cleanup gets executed on every iteration through the (infinite) loop in `run()`.  The timer free list and random file pointer don't need to be cleaned up that often, only on process exit.  Also, in your suggested change, the cleanup only happens on the server side, not the client side.  (If we think the client side isn't important to clean up because the client is going to exit, then it really isn't that important for the server side either, right?)

Although now that I'm looking at it, the server usually terminates via `iperf_got_sigend()`, not by going down to the end of `run()` so it might be important to do the cleanup there too.

Basically I don't have a strong opinion whether these memory leaks get fixed or not, but if we're going to try to fix them, let's make sure we do a complete job with it.

Thanks!

GabrielGanne:
Hi,

I'm sorry, I took some holidays and left this behind.
Looking back, I believe this small commit is optional (right without being good) and that the last version of my commit answers your comments.
What do you want to do with it ?

Best regards

bmah888:
As I'm looking at this commit, it still seems to have the issues in my last comment on 7 June...notably that the timer and entropy file get cleaned up after every pass through the server loop (but not on the client side), also the cleanup doesn't happen in the most common server exit case via `iperf_got_sigend()`.  Thank you for the PR, although I'm going to close it at this point, because (as I see things) it doesn't really address a significant problem.  I'd be happy to look at later iterations on this however.  Thanks again!

PR #597, opened 2017-06-09
ctrl_sck not closed in iperf_client_end()

p0intR:
This should fix leaks on test->ctrl_sck

I think iperf_client_end() is the right position for closing ctrl_sck.
This fixed the socket fd leaks I'm getting while running repeated client tests using this library.

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -407,6 +407,10 @@ iperf_client_end(struct iperf_test *test)
     /* show final summary */
     test->reporter_callback(test);
 
+    /* Close control socket */
+    if (test->ctrl_sck)
+        close(test->ctrl_sck);
+
     if (iperf_set_send_state(test, IPERF_DONE) != 0)
         return -1;
 


bmah888:
Good catch!  I haven't really played with this yet but it looks good so far.

On another note:  Do you think iperf3 will still Do The Right Thing with cleaning up the control connection in various error conditions in the client (or server for that matter)?  I'm wondering if there are other places that might need to close / clean up `test->ctrl_sck` as well, for example `iperf_reset_test` or `iperf_free_test`.  I am not asking you to change your pull request, this is more for discussion.

Thanks!

p0intR:
Well, test->ctrl_sck and stream->sockets are opened by functions in the context of either iperf_client_api or iperf_server_api, therefore should also be closed in this context.

I would expect iperf_run_client to close all opened sockets if it returns -1, otherwise iperf_client_end must be called.

bmah888:
OK, I think you convinced me.  Anyway the original pull request was good so I'm going to merge that shortly.  Thanks!

PR #611, opened 2017-07-02
Add getline() fallback for OSX <= 10.6

### LINKED ISSUE #607: getline() not available on OSX pre-10.7
# Context

* Version of iperf3:
4f3a7a5403b61f49ab277d0756e87c651a45bdfe

* Hardware:
Mac

Snow Leopard (10.6) and prior

built with MacPorts (gcc 4.2)

# Bug Report

* Expected Behavior
successful build

* Actual Behavior
```
Undefined symbols:
  "_getline", referenced from:
      _iperf_getpass in iperf3_profile-iperf_auth.o
ld: symbol(s) not found
```

* Steps to Reproduce
N/A

* Possible Solution
Replace the missing getline() function in [src/iperf_auth.c](https://github.com/esnet/iperf/blob/master/src/iperf_auth.c) for OSX <= 10.6. This was [proposed to libgss](http://lists.gnu.org/archive/html/help-gss/2017-04/msg00001.html) and that patch can be used here as well.

aque:
Fixes #607. Verified on 10.6.8 x64 running llvm-gcc-4.2.

* Issues fixed (if any): #607 

Implements a fallback to the `getline()` function on OSX <= 10.6. Adapted from the mailutils forum (https://lists.gnu.org/archive/html/bug-mailutils/2002-10/msg00038.html).

### DIFF
--- a/src/iperf_auth.c
+++ b/src/iperf_auth.c
@@ -34,6 +34,8 @@
 #include <stdio.h>
 #include <termios.h>
 
+#include "iperf_util.h"
+
 #if defined(HAVE_SSL)
 
 #include <openssl/bio.h>
--- a/src/iperf_util.c
+++ b/src/iperf_util.c
@@ -432,3 +432,87 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
     }
     fprintf(fp, "]\n");
 }
+
+/* Fallback getline() for OSX <= 10.6 */
+#ifdef __APPLE__
+#include <Availability.h>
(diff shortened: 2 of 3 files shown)


l2dy:
Since iperf3 is under BSD license, I'd suggest to avoid adding code under the LGPL license and use code from https://github.com/freebsd/freebsd/blob/master/contrib/file/src/getline.c instead. I didn't test it on OS X <= 10.6 though.

aque:
Thanks @l2dy. @bmah888, let me know if you would like me to test as I have a VM built with 10.6.8. Until then, I'm leaving this issue to the more qualified developers. 😄

PR #622, opened 2017-07-31
Add required RSA header for OpenSSL

pprindeville:
Fixes cross-compilation issues under certain buildroot configurations (like LEDE).

### DIFF
--- a/src/iperf_auth.c
+++ b/src/iperf_auth.c
@@ -36,6 +36,7 @@
 
 #if defined(HAVE_SSL)
 
+#include <openssl/rsa.h>
 #include <openssl/bio.h>
 #include <openssl/pem.h>
 #include <openssl/sha.h>


bmah888:
Just asking for my own information...  What's broken before this change?  Like under what circumstances does the build not work?  Just cross-compiling?  Or something else?

pprindeville:
I've only done cross-compiling so I can't say if it builds natively or not.

bmah888:
OK fair enough.  Native compiling works fine on all the supported platforms, which is why your PR was a bit of a surprise.  I'll merge this PR a little later today (have to go to meetings) but just for my reference if you got an error message when trying cross-compile, could you add that as as comment here?  Thanks!

pprindeville:
> [...] but just for my reference if you got an error message when trying cross-compile, could you add that as as comment here? Thanks!

Sure, will do that in the future.

bmah888:
Cool.  Thanks for the PR!

Issue #624, opened 2017-07-31
configure can't be told to build without openssl if it's detectable on the build system


pprindeville:
# Context

* Version of iperf3:

3.2

* Hardware:

x86_64

* Operating system (and distribution, if any):

Cross building under Ubuntu 16.04 LTS for LEDE Head

# Bug Report

* Expected Behavior

You might have libopenssl on your system but still want to build iperf3 without iperf-auth (or any OpenSSL) encumbrance.

* Actual Behavior

If you specify `--disable-openssl` or `--without-openssl` that is ignored, and if you don't put anything, it autodetects openssl anyway, as shown by:

```
configure: WARNING: unrecognized options: --disable-openssl
...
checking whether compiling and linking against OpenSSL works... yes
...
/bin/bash ../libtool  --tag=CC   --mode=link x86_64-openwrt-linux-musl-gcc -g -Os -pipe -fno-caller-saves -fno-plt -fhonour-copts -Wno-error=unused-but-set-variable -Wno-error=unused-result -iremap/home/philipp/bertram/lede/build_dir/target-x86_64_musl_powercode-bmu/iperf-3.2:iperf-3.2 -fstack-protector -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro -D_GNU_SOURCE  -Wall -g -L/home/philipp/bertram/lede/staging_dir/target-x86_64_musl_powercode-bmu/usr/lib -L/home/philipp/bertram/lede/staging_dir/target-x86_64_musl_powercode-bmu/lib -L/home/philipp/bertram/lede/staging_dir/toolchain-x86_64_gcc-5.4.0_musl/usr/lib -L/home/philipp/bertram/lede/staging_dir/toolchain-x86_64_gcc-5.4.0_musl/lib -znow -zrelro  -L/home/philipp/bertram/lede/staging_dir/target-x86_64_musl_powercode-bmu/usr/lib -o iperf3 iperf3-main.o libiperf.la -lssl -lcrypto 
```

* Steps to Reproduce

Install `openssl-devel` (or whatever) but try to build `iperf3` without it.

* Possible Solution

None so far.

bmah888:
[...]

PR #631, opened 2017-08-11
fix for issue #624 - force build without openssl

ralcini:
As requested by issue #624 , with this fix we are able to force the build of iperf skipping openssl library check. 

At the same time, I think it's better to raise an exception if openssl is enabled but not found into the system.

please run autoconf after pull changes!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -101,13 +101,16 @@ AC_CHECK_HEADERS([netinet/sctp.h],
 #endif
 ])
 
-# Check for OPENSSL support
-AX_CHECK_OPENSSL(
-	AC_DEFINE([HAVE_SSL], [1], [OpenSSL Is Available])
-)
-LDFLAGS="$LDFLAGS $OPENSSL_LDFLAGS"
-LIBS="$OPENSSL_LIBS $LIBS"
-CPPFLAGS="$OPENSSL_INCLUDES $CPPFLAGS"
+if test "x$with_openssl" != "xno"; then
+    # Check for OPENSSL support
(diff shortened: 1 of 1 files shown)


pprindeville:
@ralcini BTW, I wanted to email you about something related (I think Bruce did too) but your email isn't exposed on your profile...

bmah888:
Wow, those big red "X"s in the GitHub UI look pretty daunting.  I think the PR is great in general, and it's actually a pretty small change I'm requesting in my review.

configure.ac
```diff
@@ -101,13 +101,16 @@ AC_CHECK_HEADERS([netinet/sctp.h],
...
 ])
 
-# Check for OPENSSL support
-AX_CHECK_OPENSSL(
-	AC_DEFINE([HAVE_SSL], [1], [OpenSSL Is Available])
-)
-LDFLAGS="$LDFLAGS $OPENSSL_LDFLAGS"
-LIBS="$OPENSSL_LIBS $LIBS"
-CPPFLAGS="$OPENSSL_INCLUDES $CPPFLAGS"
+if test "x$with_openssl" != "xno"; then
```
pprindeville: Per Bruce's request, you could do:

```
if test "x$with_openssl" = "xno"; then
  AC_MSG_WARN([Building without OpenSSL; disabling iperf_auth functionality.])
else
```

ralcini:
Wow, those big red "X"!!! =)
I've just committed the warning message that you suggested.

bmah888:
Merged to master and the 3.2-STABLE branch, thanks!

Issue #608, opened 2017-06-25
Windows 10 Network Performance Test --4 and --6 unknown 


ericbassett:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the iperf3 mailing list at iperf-dev@google-groups.com or
question sites such as Stack Overflow
(http://www.stackoverflow.com/).  A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3:
Windows 10 Network Performance Test
* Hardware:
HP Spectre x360
* Operating system (and distribution, if any):
Windows 10
_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report

* Expected Behavior
Selecting IPV4 or IPV6 adds "-4" or "-6" to arguments
* Actual Behavior
Selecting IPV4 or IPV6 adds "--4" and "--6" (extra dash)
* Steps to Reproduce
Run a client mode iperf3 set to TCP and IPV4 or IPV6
* Possible Solution
Remove extra dash in arguments
_Please submit patches or code changes as a pull request._

bmah888:
It sounds like you are using some sort of Windows based GUI that calls an iperf3 implementation underneath?  I am inferring that from your description of expected and actual behavior...iperf3 is a command-line application (for which Windows is not officially supported)...users of iperf3 don't "select" anything, they add the command-line arguments (such as `-4` or `-6`) directly.

If my guess is correct, then you really need to file an issue / bug with whatever project created and/or distributed the GUI front-end you are using.  (I'd be interested to see a pointer to that as well, just for my edification.)

bmah888:
No other feedback.  I'm going to close this as an invalid bug because based on the best information I have, this isn't really an iperf3 issue.

PR #640, opened 2017-09-18
Two fixes for a warning for possibly-too-large UDP packets.

### LINKED ISSUE #604: iperf 3.1.6 UDP block size exceeds TCP MSS 0
Iperf 3.1.6
Win10 TH2

I am running iperf to generate packets between my host PC to a VM.
Following command works fine but I receive a warning and I do not know exactly if I can ignore it or not:
```
iperf3.exe -c 10.10.10.2 -t 60 -O 1 -fm  -V  -u  -b 1000m  -R  -l 5000.

Warning:  UDP block size 5000 exceeds TCP MSS 0, may result in fragmentation / drops
Summary Results:
[ ID] Interval           Transfer     Bandwidth       Jitter    Lost/Total Datagrams
[  4]   0.00-60.00  sec  6.90 GBytes   987 Mbits/sec  148.369 ms  153028/1486702 (10%)  
[  4] Sent 1486702 datagrams
[SUM]  0.0-60.0 sec  35886 datagrams received out-of-order
CPU Utilization: local/receiver 27.2% (9.4%u/17.9%s), remote/sender 41.4% (0.9%u/40.5%s)
```
I also tried without the -l option and I receive the same warning but the Bandwidth is 548Mbit/sec and the warning is:  UDP block size 1460 exceeds TCP MSS 0, may result in fragmentation.

EDIT:  Reformatted for command output legibility. --bmah

bmah888:
1.  Make sure we have a valid TCP MSS on the path when comparing
the UDP block size.

2.  Fix a redundant "warning".

This should fix a bug observed on Windows but not (so far) on any
UNIX-like platforms.  Fixes #604.

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -382,10 +382,11 @@ iperf_connect(struct iperf_test *test)
 	 * Regardless of whether explicitly or implicitly set, if the
 	 * block size is larger than the MSS, print a warning.
 	 */
-	if (test->settings->blksize > test->ctrl_sck_mss) {
+	if (test->settings->blksize > test->ctrl_sck_mss &&
+	    test->ctrl_sck_mss > 0) {
 	    char str[128];
 	    snprintf(str, sizeof(str),
-		     "Warning:  UDP block size %d exceeds TCP MSS %d, may result in fragmentation / drops", test->settings->blksize, test->ctrl_sck_mss);
+		     "UDP block size %d exceeds TCP MSS %d, may result in fragmentation / drops", test->settings->blksize, test->ctrl_sck_mss);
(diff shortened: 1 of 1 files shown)


bmah888:
Note:  Need to double-check the final commit message when merging this, in at least one place we mistakenly linked this to issue #608.

bmah888:
@pprindeville :  Good point about reversing the order of tests, thank you.  Will fix this.  Yes, the second fix is getting rid of "Warning: ", which is redundant given that the warning() function adds "warning: " at the start of its output.

bmah888:
@pprindeville : Thanks for the review/approval!

PR #641, opened 2017-09-19
Fixed the problme that automatically selected format will caused the …

shingchuang:
This fixed the bug that automatically selected format conversion unit will caused a segmentation fault, due the index go beyond the range of array!

### DIFF
--- a/src/units.c
+++ b/src/units.c
@@ -316,6 +316,7 @@ extern    "C"
 			conv++;
 		    }
 		}
+		if (conv > TERA_CONV) conv = TERA_CONV;
 		break;
 	    }
 	}


bmah888:
Good catch.  Would it be possible to get the same effect as your fix by just changing `conv <= TERA_CONV` to `conv < TERA_CONV` in the two `while()` statements above?  In the case where we walk off the end of the enum, this prevents `conv` from ever taking on an invalid value, as opposing to letting it become invalid and then fixing it up after the loop.

shingchuang:
I had thought about your version and the following version:
                float     f=isupper((int) inFormant)) ? 1024.0 : 1000.0;
                while (tmpNum >= f && conv <TERA_CONV)
                {
                    tmpNum /= f;
                    conv++;
                }
I just thought my version only modified one line, only one line makes everyone quickly to understand what I talked about, that's all! Anyway,  just fixed it, I would accepted any version!! And, thanks for your works, it helps a lot!

bmah888:
I'm going to close this pull request because it's basically been superceded by #642.  Thanks to @shingchuang for pointing out the problem!

Issue #650, opened 2017-10-11
authentication token not reset between tests


ralcini:
The problem occurrs between the execution of authenticated and unauthenticated tests and only if the iperf server is not restarted between attempts. 

The authentication credentials between the attempts seems to be cached.

First step - client side:
```
$> src/iperf3 -c localhost --rsa-public-key-path public.pem --username mario
Password: Connecting to host localhost, port 5201
iperf3: error - test authorization failed
```
First step - server side:
```
$> src/iperf3 -s --rsa-private-key-path private_not_protected.pem --authorized-users-path credentials.csv
-----------------------------------------------------------
Server listening on 5201
-----------------------------------------------------------
Authentication failed for user 'mario' ts 1507739247
iperf3: error - test authorization failed
-----------------------------------------------------------
Server listening on 5201
-----------------------------------------------------------
```

Second step - client side:
```
$> src/iperf3 -c localhost
Password: Connecting to host localhost, port 5201
iperf3: error - test authorization failed
```

Second step - server side - we can see the user mario used in the first attempt and the ts used.
```
Authentication failed for user 'mario' ts 1507739247
iperf3: error - test authorization failed
-----------------------------------------------------------
Server listening on 5201
-----------------------------------------------------------
```

If the first attempt is successful, the second attempt without credential will succeed for if executed during the ts validity (30 seconds).

The problem is due to a lack of authtoken variable reinitialization between tests executions.

PR #651, opened 2017-10-11
fix for authentication token not reset between tests #650

ralcini:
fix for #650

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2305,6 +2305,7 @@ iperf_reset_test(struct iperf_test *test)
     test->settings->rate = 0;
     test->settings->burst = 0;
     test->settings->mss = 0;
+    test->settings->authtoken = NULL;
     memset(test->cookie, 0, COOKIE_SIZE);
     test->multisend = 10;	/* arbitrary */
     test->udp_counters_64bit = 0;


src/iperf_api.c
```diff
@@ -2305,6 +2305,7 @@ iperf_reset_test(struct iperf_test *test)
     test->settings->burst = 0;
     test->settings->mss = 0;
     test->settings->tos = 0;
+    test->settings->authtoken = NULL;
```
pprindeville: Isn't this potentially a leak?  I'd do:

```
free(test->settings->authtoken);
test->settings->authtoken = NULL;
```

remembering that `free(NULL);` is a safe (but inefficient) operation.

bmah888: Yep, you're right...I should have caught that.  (The hazards of working on both C and Java code at the same time.)  Thanks for the correction.  I usually do something like:

```
if (test->settings->authtoken) {
    free(test->settings->authtoken);
    test->settings->authtoken = NULL;
}
```
I'm guessing there's no easy way short of unmerging to tell GitHub to re-open a pull request, right?

pprindeville: Don't know if there is or not.  What happens if you click on Revert?

bmah888: ```Sorry, this pull request couldn’t be reverted automatically. It may have already been reverted, or the content may have changed since it was merged.```

I know why...a line of code I changed just above this is preventing the reverse diff from applying.  I'll just fix this with a new PR...I'd like you to look that over if you don't mind.  Thanks!

ralcini: sorry for the issue! So, could I help with the new PR?

bmah888: @ralcini: No worries.  The commit you just mentioned is (I think) the same as what I did for PR #652.  Want to give that a review?  Thanks!

ralcini: @bmah888 yes, it is!

PR #652, opened 2017-10-12
Fix memory leak introduced in 1460aa3433769db1f5660e7819b669e5414e9b42.

bmah888:
Pointed out by:  @pprindeville
Should have been found by:  @bmah888


bmah888:
Awesome, thanks!  I'm going to let the Travis CI builds for that pull request finish first before merging.  travis.org seems to get unhappy when I work faster than it does.

Issue #369, opened 2016-02-28
daemon() is not portable


canavan:
daemon() is not portable; add a check to configure.ac, use own implementation if it is not available

diff --git a/configure.ac b/configure.ac
index c85b907..3d8b825 100644
--- a/configure.ac
+++ b/configure.ac
@@ -141,4 +141,11 @@ AC_CHECK_FUNCS([cpuset_setaffinity sched_setaffinity],
 # it needs and what arguments it expects.
 AC_CHECK_FUNCS([sendfile])

+# AC_DEFINE([HAVE_DAEMON], [], [Description])
+AC_CHECK_FUNC(daemon, [ac_cv_func_daemon=yes; AC_DEFINE(HAVE_DAEMON, 1, [Have daemon()])])
+
+AC_CONFIG_FILES([
+lib/Makefile
+])
+
 AC_OUTPUT([Makefile src/Makefile src/version.h examples/Makefile iperf3.spec])
diff --git a/src/main.c b/src/main.c
index 4bd5a32..150a227 100644
--- a/src/main.c
+++ b/src/main.c
@@ -125,6 +125,65 @@ sigend_handler(int sig)
     longjmp(sigend_jmp_buf, 1);
 }

+#ifndef HAVE_DAEMON
+#include <sys/stat.h>
+#include <fcntl.h>
+
+int daemon(int nochdir, int noclose)
+{
-    pid_t pid = 0;
-    pid_t sid = 0;
-   
-    pid = fork();
-    if (pid < 0)
-    {
-   return -1;
-    }
-   
-    if (pid > 0)
-    {
-   exit(0);
-    }
-   
-    umask(0);
-   
-    sid = setsid();
-    if(sid < 0)
-    {
-   return -1;
-    }
  +
-    signal(SIGHUP, SIG_IGN);
-    pid=fork();
-    if (pid == -1) {
-   iperf_errexit(NULL, "failed to fork while daemonising (errno=%d)", iperf_strerror(errno));
-    } else if (pid != 0) {
-   exit(0);
-    }
  +
-    if (nochdir == 0) {
-   chdir("/");
-    }
-   
-    if (noclose == 0) {
-   close(STDIN_FILENO);
-   close(STDOUT_FILENO);
-   close(STDERR_FILENO);
  +
-   if (open("/dev/null", O_RDONLY) == -1) {
-       return -1;
-   }
-   if (open("/dev/null", O_WRONLY) == -1) {
-       return -1;
-   }
[...]

PR #655, opened 2017-10-17
Issue 369 (implement daemon(3) for systems that don't have it)

bmah888:
Implement daemon(3) for systems that don't have it.


src/missing_daemon.c
```diff
@@ -0,0 +1,56 @@
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <signal.h>
+
+int missing_daemon(int nochdir, int noclose)
```
pprindeville: A few comments about this function in general.

1. the usual order is `chdir()`, then `open()/dup2()`, then `fork()`, `setsid()`, and `fork()` again
2. changing the `umask` is non-standard behavior
3. the `signal()` is also non-standard behavior
4. I would open `/dev/null` as `O_RDWR` and then `dup2()` it onto `0`, `1`, and `2`, before closing the original `fd`. Note the `dup2()` saves you from having to close the fd's 0-2.

Not sure I see the point of using an alternate name if the original name is absent, since there's no chance of a clash by definition.

I would, however, protect the entire function with `#ifndef HAVE_DAEMON` so it compiles down to zero bytes and doesn't pollute the namespace if `HAVE_DAEMON` is defined.

bmah888: Thanks for the review comments!  For the record I cobbled this PR together out of the patch that was submitted in issue #369.

After having looked more carefully at the patch plus your review, I wonder if it'd be better to:

1.  Fix the actual `missing_daemon()` implementation as per your comments.  I had assumed that there was a single canonical order of operations, and that was it.

2.  Rename to the original function name.

3.  Move the function definition to `iperf_util.c` (doesn't need to be in a seperate file) and wrap it with `#ifndef HAVE_DAEMON`.

4.  Maybe add a function declaration to `iperf_util.h`, wrapped by `#ifndef HAVE_DAEMON`.  This might require that `iperf_util.h` include `iperf_config.h`, which it doesn't currently, and I think that might be an issue for the `getline()` definition recently added to the file.

pprindeville: Yes to all.

src/iperf_util.c
```diff
@@ -432,3 +432,57 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
     }
     fprintf(fp, "]\n");
 }
+
+/* daemon(3) implementation for systems lacking one */
+#ifndef HAVE_DAEMON
+int daemon(int nochdir, int noclose)
+{
+    pid_t pid = 0;
```
pprindeville: I'd rewrite this implementation to do the order of operations we previously described, use `dup2()`, etc.

bmah888: Yeah I just didn't want to deal with all those changes at the same time.  :-)

It's also interesting that what's in the patch very roughly corresponds to the most readily-available source for daemon(3) that I have (FreeBSD).  The order of operations you gave, is that based on glibc or something else?  I was going to dig through various sources to see what to do...still trying to wrap my mind around the idea that there are multiple ways to do daemon(3) (and yeah I know it hasn't really been standardized so that's probably to be expected).

pprindeville:
Yes, from [glibc](https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=misc/daemon.c;hb=HEAD)...

bmah888:
@pprindeville :  OK got it, thanks for the pointer.  Philosophically I'm not sure whether to follow the BSD or Linux way but I'm going to read through the glibc implementation first before deciding anything.

pprindeville:
> Philosophically I'm not sure whether to follow the BSD or Linux way but I'm going to read through the glibc implementation first before deciding anything.

Well, the glibc implementation is elegant and concise.

bmah888:
So...you're gonna hate me for this.

I actually sat down and compared the daemon(3) function definition between Linux/glibc, FreeBSD, and what was submitted in #369.  Linux/glibc and FreeBSD are actually much closer than I thought (so my earlier comments on this were invalid).  They both do fork(), setsid(), chdir(), dup2().  Notably they both only do one fork() call.  FreeBSD also does some signal-handling stuff as well.  Neither resets umask().

glibc daemon (from your earlier link):

https://sourceware.org/git/?p=glibc.git;a=blob_plain;f=misc/daemon.c;hb=HEAD

FreeBSD daemon:

https://svnweb.freebsd.org/base/head/lib/libc/gen/daemon.c?view=co

The code submitted to iperf3 in #369 does fork(), umask(), setsid(), signal, fork() again, chdir(), and open(), which is totally different.  I can't figure out where this order of operations comes from, although it's somewhat similar to what I see in R. Stevens' "Advanced Programming in the UNIX Environment".

All of these are different from what you listed in comments above, which is "`chdir()`, then `open()/dup2()`, then `fork()`, `setsid()`, and `fork()` again".

I'm still trying to wrap my mind around all this, but I'm leaning towards what Linux and FreeBSD do (they're both pretty close and they're used quite a lot).

src/iperf_util.c
```diff
@@ -433,56 +434,65 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
...
+    int fd;
+
+    /*
+     * Ignore any possible SIGHUP when the parent process exits.
+     * Note that the iperf3 server process will eventually install
+     * its own signal handler for SIGHUP, so we can be a little
+     * sloppy about not restoring the prior value.  This does not
+     * generalize.
+     */
+    signal(SIGHUP, SIG_IGN);
```
pprindeville: If this is non-standard behavior AND iperf3 doesn't rely on it anyway because it ends up doing its own thing... then KISS.

src/iperf_util.c
```diff
@@ -433,56 +434,65 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
...
-
     sid = setsid();
     if (sid < 0) {
 	return -1;
     }
 
-    signal(SIGHUP, SIG_IGN);
+    /*
+     * Fork again to avoid becoming a session leader.
+     * This might only matter on old SVr4-derived OSs.
```
pprindeville: Anyone still running SVR4 should be led through the streets with people shouting "Shame!" at them, as they clearly don't care about being a beachhead for botnet attacks, etc...

src/iperf_util.c
bmah888: No, it's not non-standard at all, any more than any other daemon(3) behavior since it seems to vary between implementations.  The idea is to make sure the child isn't affected by a SIGHUP when the parent dies.  iperf3 installs a signal handler later so we don't need to worry about restoring the original value, but the effect is still there.  FreeBSD does this BTW.

src/iperf_util.c
bmah888: SVr4-*derived* OSs.  I don't mind getting rid of this except that the person who originally filed a bug was running IRIX, which is very much based on SVr4.  So I don't want to go through all this trouble just to have something that won't work for them.

bmah888: (Whether we should be expending this amount of effort on such old platforms is a different question, but I figure we've put this much effort into it, we may as well finish it off.)

pprindeville: > So I don't want to go through all this trouble just to have something that won't work for them.

Right, I just thought you said something about iperf3 installing its own handler anyway.  In any case, nothing major...

bmah888: OK.  I really appreciate your feedback and discussion.  At this point I think I want to merge this code to get it out into the world for more testing (I've done about all I call given that I don't have a platform that would actually use this daemon() implementation, so I need to fake out some stuff just to test it).  If something doesn't work, presumably somebody will yell about it and I'll fix it.  Thanks again!

Issue #176, opened 2014-05-19
3.0.3: Solaris missing getopt


marksolaris:
Pre Solaris 10, getopt didn't exist. This code adds it. It'll be compiled for every platform, but only called on Solaris machines (#ifdef sun). This code works on Solaris machines with getopt as well, the patch will install this instead of the OS routine.

It could be standardized for other OS's by checking for /usr/include/getopt.h in configure and defining NEED_GETOPT, then activating this code for that.

```
diff -rubN iperf-3.0.3.orig/src/Makefile.am iperf-3.0.3/src/Makefile.am
--- iperf-3.0.3.orig/src/Makefile.am    2014-03-27 05:06:38.000000000 +1100
+++ iperf-3.0.3/src/Makefile.am 2014-05-19 11:03:39.433413684 +1000
@@ -29,6 +29,8 @@
                         tcp_info.c \
                         tcp_window_size.c \
                         tcp_window_size.h \
+                        missing_getopt.c \
+                        missing_getopt.h \
                         timer.c \
                         timer.h \
                         units.c \
diff -rubN iperf-3.0.3.orig/src/Makefile.in iperf-3.0.3/src/Makefile.in
--- iperf-3.0.3.orig/src/Makefile.in    2014-03-27 05:06:38.000000000 +1100
+++ iperf-3.0.3/src/Makefile.in 2014-05-19 11:04:27.740578504 +1000
@@ -81,7 +81,7 @@
 am_libiperf_la_OBJECTS = cjson.lo iperf_api.lo iperf_error.lo \
        iperf_client_api.lo iperf_server_api.lo iperf_tcp.lo \
        iperf_udp.lo iperf_util.lo locale.lo net.lo tcp_info.lo \
-       tcp_window_size.lo timer.lo units.lo
+       tcp_window_size.lo missing_getopt.lo timer.lo units.lo
 libiperf_la_OBJECTS = $(am_libiperf_la_OBJECTS)
 PROGRAMS = $(bin_PROGRAMS) $(noinst_PROGRAMS)
 am_iperf3_OBJECTS = iperf3-main.$(OBJEXT)
@@ -101,7 +101,7 @@
        iperf3_profile-locale.$(OBJEXT) iperf3_profile-net.$(OBJEXT) \
        iperf3_profile-tcp_info.$(OBJEXT) \
        iperf3_profile-tcp_window_size.$(OBJEXT) \
[...]

PR #656, opened 2017-10-17
Issue 176 (improve compatibility on platforms without getopt_long())

bmah888:
Towards #176, which basically improves compatibility with Solaris
and other systems that don't have getopt_long().

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -32,7 +32,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef HAVE_GETOPT_H
 #include <getopt.h>
+#else
+#include "missing_getopt.h"
+#endif /* HAVE_GETOPT_H */
 #include <errno.h>
 #include <signal.h>
 #include <unistd.h>
@@ -705,7 +709,12 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
     char *client_username = NULL, *client_rsa_public_key = NULL;
 #endif /* HAVE_SSL */
 
-    while ((flag = getopt_long(argc, argv, "p:f:i:D1VJvsc:ub:t:n:k:l:P:Rw:B:M:N46S:L:ZO:F:A:T:C:dI:hX:", longopts, NULL)) != -1) {
+    char optstring[] = "p:f:i:D1VJvsc:ub:t:n:k:l:P:Rw:B:M:N46S:L:ZO:F:A:T:C:dI:hX:";
+#ifdef HAVE_GETOPT_H
+    while ((flag = getopt_long(argc, argv, optstring, longopts, NULL)) != -1) {
+#else
+    while ((flag = netbsd_getopt_long(argc, argv, optstring, longopts, NULL)) != -1) {
+#endif /* HAVE_GETOPT_H */
         switch (flag) {
             case 'p':
                 test->server_port = atoi(optarg);
--- a/src/missing_getopt.c
+++ b/src/missing_getopt.c
@@ -1,0 +1,409 @@
+/*	$NetBSD: getopt_long.c,v 1.11 2009/04/14 17:34:41 joerg Exp $	*/
+
+/*-
+ * Copyright (c) 2000 The NetBSD Foundation, Inc.
+ * All rights reserved.
+ *
+ * This code is derived from software contributed to The NetBSD Foundation
+ * by Dieter Baron and Thomas Klausner.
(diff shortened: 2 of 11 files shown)


src/missing_getopt.c
```diff
@@ -0,0 +1,409 @@
...
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "missing_getopt.h"
```
pprindeville: Not sure it's worth naming the file differently since one is in the standard include path and the other is in-tree.

src/missing_getopt.c
```diff
@@ -0,0 +1,409 @@
...
+
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdarg.h>
+
+#undef __UNCONST
```
pprindeville: I would protect the whole rest of the file with `#ifndef HAVE_GETOPT`.

src/missing_getopt.c
```diff
@@ -0,0 +1,409 @@
...
+
+	/*
+	 * compute lengths of blocks and number and size of cycles
+	 */
+	nnonopts = panonopt_end - panonopt_start;
+	nopts = opt_end - panonopt_end;
+	ncycle = gcd(nnonopts, nopts);
+	cyclelen = (opt_end - panonopt_start) / ncycle;
+
+	for (i = 0; i < ncycle; i++) {
```
pprindeville: Not a fan of using `int` for ordinal counters.

src/missing_getopt.c
```diff
@@ -0,0 +1,409 @@
...
+		cstart = panonopt_end+i;
+		pos = cstart;
+		for (j = 0; j < cyclelen; j++) {
```
pprindeville: Ditto.

Then again, you're using canned code which supposedly works, so maybe it's not worth changing...

src/iperf_api.c
```diff
@@ -32,7 +32,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef HAVE_GETOPT_H
 #include <getopt.h>
+#else
+#include "missing_getopt.h"
```
pprindeville: Is it called `missing_getopt.h` upstream?  If not, I wouldn't bother with the name change since clashing isn't really a problem if (a) the file is non-existent and (b) we're talking about standard includes versus in-tree.

src/iperf_api.c
```diff
@@ -705,7 +709,12 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
     char *client_username = NULL, *client_rsa_public_key = NULL;
 #endif /* HAVE_SSL */
 
-    while ((flag = getopt_long(argc, argv, "p:f:i:D1VJvsc:ub:t:n:k:l:P:Rw:B:M:N46S:L:ZO:F:A:T:C:dI:hX:", longopts, NULL)) != -1) {
+    char optstring[] = "p:f:i:D1VJvsc:ub:t:n:k:l:P:Rw:B:M:N46S:L:ZO:F:A:T:C:dI:hX:";
+#ifdef HAVE_GETOPT_H
+    while ((flag = getopt_long(argc, argv, optstring, longopts, NULL)) != -1) {
+#else
+    while ((flag = netbsd_getopt_long(argc, argv, optstring, longopts, NULL)) != -1) {
```
pprindeville: The other thing is, why use 2 different prefixes of `netbsd_` versus `missing_`?  Seems confusing.

bmah888:
Closing this...too much time has passed, particularly given the platform that motivated this work.

Thank you to @pprindeville for the comments...they will be useful if and when we ever need to revive this PR.

Issue #659, opened 2017-10-24
Windows UDP test. warning: Block size


LikHait:
# Context

* Version of iperf3:
iperf 3.2 STABLE/ master
* Operating system (and distribution, if any):
windows 10 x64(cygwin)

# Bug Report

* Expected Behavior
Successful execution with default value

* Actual Behavior
warning: Block size 401152 > sending socket buffer size 65536
warning: Block size 401152 > receiving socket buffer size 65536
Increasing socket buffer size to 402176
[ 5] local 192.168.0.101 port 52569 connected to 82.144.193.18 port 5201
iperf3: error - unable to write to stream socket: Message too long
* Steps to Reproduce
iperf3 -c xxx.xxx.xxx.xxx -u

ralcini:
in some cases the issue degenerate in #658

bmah888:
OK.  #658 and this bug seem related so I'm going to discuss both of them in this issue (@ralcini I hope that's OK with you).  As I wrote on #660, the underlying issue seems to be that `test->ctrl_sck_mss` got some strange, large value.  The code change in #660 keeps us from using that to set a block size that's too large, which is good, but that doesn't help the fact that in this situation it should have just used `DEFAULT_UDP_BLKSIZE` as the default.

Going back to this code in `iperf_connect`:

```
    int opt;
    socklen_t len;

    len = sizeof(opt);
    if (getsockopt(test->ctrl_sck, IPPROTO_TCP, TCP_MAXSEG, &opt, &len) < 0) {
	test->ctrl_sck_mss = 0;
    }
    else {
	test->ctrl_sck_mss = opt;
    }
```

If the getsockopt(3) call fails (which presumably it does on Windows earlier than Windows 10), then `test->ctrl_sck_mss` gets set to 0.  In that case, if `tcp->settings->blksize` wasn't set already, it gets `DEFAULT_UDP_BLKSIZE`.

If the getsockopt(3) call succeeds (Windows 10 and later), then `test->ctrl_sck_mss` gets some value that's reasonable, and that becomes the value of `tcp->settings->blksize` if it wasn't already set.

So I'm trying to understand how an invalid, large number could ever get into `test->settings->blksize` in the first place.  It doesn't seem to me there's a way for this to happen, unless on Windows systems it's possible for the getsockopt(3) call to return successfully _and_ put an invalid value in the `opt` variable.

OK.  Let's look at this code (same function, a little lower down):
```
    if (test->protocol->id == Pudp) {
[...]

PR #660, opened 2017-10-24
Issue #659 

LikHait:
* Version of iperf3 : 
master/ 3.2-STABLE
* Issues fixed (if any):
#659 
Fixed UDP blksize default value

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -339,12 +339,22 @@ iperf_connect(struct iperf_test *test)
     socklen_t len;
 
     len = sizeof(opt);
-    if (getsockopt(test->ctrl_sck, IPPROTO_TCP, TCP_MAXSEG, &opt, &len) < 0) {
-	test->ctrl_sck_mss = 0;
-    }
-    else {
-	test->ctrl_sck_mss = opt;
-    }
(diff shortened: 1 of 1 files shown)


bmah888:
After thinking about this some more, let's try something slightly different.  Let's move your sanity check test to right after we try to retrieve the value, rather than just doing the sanity check where we try to set`test->settings->blksize`.  The reason is that we use `test->ctrl_sck_mss` later on the code, and if it has a weird value that we decided not to use it's going to print a confusing warning.  So maybe something like this:
```
    int opt;
    socklen_t len;

    len = sizeof(opt);
    if (getsockopt(test->ctrl_sck, IPPROTO_TCP, TCP_MAXSEG, &opt, &len) < 0) {
	test->ctrl_sck_mss = 0;
    }
    else {
	if (opt > 0 && opt <= MAX_UDP_BLOCKSIZE) {
	    test->ctrl_sck_mss = opt;
	} 
	else {
	    char str[128];
	    snprintf(str, sizeof(str),
		     "Ignoring nonsense TCP MSS %d", opt);
	    warning(str);    

	    test->ctrl_sck_mss = 0;
	}
    }
```
(And then remove the sanity check from the place where you added it in your PR.)

How's that look to you?  (And does it work correctly?)  If you agree, do you want to update your PR?  Thanks!

LikHait:
Ok. It looks good

bmah888:
Merged, thanks!

PR #663, opened 2017-10-31
Provide a getline(3) implementation for systems without it.

### LINKED ISSUE #607: getline() not available on OSX pre-10.7

bmah888:
The getline(3) implementation comes from NetBSD's file(1) command.
This should make compiling work on macOS pre 10.7.  Fixes #607.

### DIFF
--- a/src/iperf_util.c
+++ b/src/iperf_util.c
@@ -432,3 +432,86 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
     }
     fprintf(fp, "]\n");
 }
+
+/* Compatibility version of getline(3) for systems that don't have it.. */
+#ifndef HAVE_GETLINE
+/* The following code adopted from NetBSD's getline.c, which is: */
+
+/*-
+ * Copyright (c) 2011 The NetBSD Foundation, Inc.
+ * All rights reserved.
+ *
+ * This code is derived from software contributed to The NetBSD Foundation
+ * by Christos Zoulas.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE NETBSD FOUNDATION, INC. AND CONTRIBUTORS
+ * ``AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED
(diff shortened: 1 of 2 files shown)


src/iperf_util.c
```diff
@@ -432,3 +432,86 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
     }
     fprintf(fp, "]\n");
 }
+
+/* Compatibility version of getline(3) for systems that don't have it.. */
+#ifndef HAVE_GETLINE
```
pprindeville: Now you're talking... that's what I like to see!

src/iperf_util.c
```diff
@@ -432,3 +432,86 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
...
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL THE FOUNDATION OR CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+ssize_t
```
pprindeville: If this weren't poached code, I'd suggest a `static` here to not pollute the namespace, but since it is pulled from somewhere else I'd say "leave it as-is".

src/iperf_util.c
```diff
@@ -432,3 +432,86 @@ iperf_dump_fdset(FILE *fp, char *str, int nfds, fd_set *fds)
...
+getdelim(char **buf, size_t *bufsiz, int delimiter, FILE *fp)
+{
+	char *ptr, *eptr;
+
```
pprindeville: Pet peeve... multiple blank lines...

PR #664, opened 2017-10-31
Fix 3.3 build warnings

pprindeville:
To be applied to `master`.

Fixes compilation warnings against the MUSL library, and simplifies the detection of the correct endianness (via header file when present) without explicitly dependencies on detecting the type of C run-time being built against.

Fixes type mismatches on 64-bit machines where `size_t` or `uint64_t` values are being formatted as `%llu`.

Fixes warnings when `_GNU_SOURCE` is redefined.

### DIFF
--- a/src/iperf_tcp.c
+++ b/src/iperf_tcp.c
@@ -24,8 +24,6 @@
  * This code is distributed under a BSD style license, see the LICENSE
  * file for complete information.
  */
-#include "iperf_config.h"
-
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -43,6 +41,12 @@
 #include "iperf_tcp.h"
 #include "net.h"
 #include "cjson.h"
+
+#if defined(HAVE_INTTYPES_H)
+# include <inttypes.h>
+#else
+# define PRIu64		"llu"
+#endif
 
 #if defined(HAVE_FLOWLABEL)
 #include "flowlabel.h"
@@ -90,7 +94,7 @@ iperf_tcp_send(struct iperf_stream *sp)
     sp->result->bytes_sent_this_interval += r;
 
     if (sp->test->debug)
-	printf("sent %d bytes of %d, total %llu\n", r, sp->settings->blksize, sp->result->bytes_sent);
+	printf("sent %d bytes of %d, total %" PRIu64 "\n", r, sp->settings->blksize, sp->result->bytes_sent);
 
     return r;
 }
--- a/src/portable_endian.h
+++ b/src/portable_endian.h
@@ -10,17 +10,36 @@
 
 #endif
 
-// GLIBC / Linux with endian(3) support, which was added in glibc 2.9.
-// Intended to support CentOS 6 and newer.
-#if defined(__linux__) && \
-    ((__GLIBC__ > 3) || \
-     (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 9))
+#if defined(__CYGWIN__)
 
 #	include <endian.h>
 
-#elif defined(__CYGWIN__)
-
+#elif defined(HAVE_ENDIAN_H)
 #	include <endian.h>
+
+#elif defined(HAVE_SYS_ENDIAN_H)
+#	include <sys/endian.h>
+
+#	if defined(__OpenBSD__)
+
+#		define be16toh(x) betoh16(x)
+#		define le16toh(x) letoh16(x)
+
+#		define be32toh(x) betoh32(x)
+#		define le32toh(x) letoh32(x)
+
+#		define be64toh(x) betoh64(x)
+#		define le64toh(x) letoh64(x)
+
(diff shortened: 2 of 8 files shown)


pprindeville:
Hm.... `configure` will need to be refreshed after `autoconf` is run, but I have a later versions so when I run it all sorts of things change due to updates in the library files.

bmah888:
Sigh.  Bear with me, I'm still pretty new to the GitHub pull request model.

I was trying to figure out how to comment on a particular part of a pull request, but obviously failing in a confusing manner.

pprindeville:
> I was trying to figure out how to comment on a particular part of a pull request, but obviously failing in a confusing manner.

@bmah888 Easiest way is to click on the "Files changed" tab, scroll down the line in interest, and then mouse-over the line number... when a "+" appears click on it to add your comment.

src/portable_endian.h
```diff
@@ -10,14 +10,33 @@
 
 #endif
 
-// GLIBC / Linux with endian(3) support, which was added in glibc 2.9.
-// Intended to support CentOS 6 and newer.
-#if defined(__linux__) && \
-    ((__GLIBC__ > 3) || \
-     (__GLIBC__ == 2 && __GLIBC_MINOR__ >= 9))
-
+#if defined(HAVE_ENDIAN_H)
```
bmah888: I think I get the idea here...rather than having specific cases for every OS (which admittedly seems to have been the original intent), the new idea is to include either `<endian.h>` or `<sys/endian.h>` as needed (or in some cases neither one), and then add OS-specific definitions as needed.  This looks like a reasonable and somewhat more flexible approach.  But...it looks like this might break the Cygwin case.  Cygwin appears to have `<endian.h>`, but the only way to get to `#elif defined(__CYGWIN__)` is to have neither `<endian.h>` nor `<sys/endian.h>`.  Let me know if I read this wrongly.

src/iperf_tcp.c
```diff
@@ -24,8 +24,6 @@
  * This code is distributed under a BSD style license, see the LICENSE
  * file for complete information.
  */
-#include "iperf_config.h"
```
bmah888: Why get rid of this?  I admit the code isn't very consistent in this area, but I tend to lean in the direction of "when in doubt, add the autoconf-generated header", particularly if it's not immediately obvious how that header would get included in the compile (through `src/iperf.h` apparently).  Is this a style change or is there something functional that it fixes?

src/iperf_tcp.c
```diff
@@ -44,6 +42,12 @@
 #include "net.h"
 #include "cjson.h"
 
+#if defined(HAVE_INTTYPES_H)
+# include <inttypes.h>
```
bmah888: Nice to know there's a real solution to this problem!  I need to go read about the usage of `<inttypes.h>`.  And thanks for providing a fallback for non-C99-compliant people.

bmah888:
@pprindeville : Thanks for the GitHub help, think I got it now.  The TL;DR is that I think this is all good, except maybe for the Cygwin issue.  Many thanks for helping me clean up this stuff.

src/portable_endian.h
pprindeville: Moving the `CYGWIN` test to the top.  Will rebuild and try a new binary.

src/iperf_tcp.c
pprindeville: It's for consistency: none of the system include files should be affected by anything in that file (or any files it includes in-tree).  The first in-tree header file that gets included is `"iperf.h"` which promptly includes `"iperf_config.h"`.  This is what `src/iperf_udp.c` does.

src/iperf_tcp.c
pprindeville: Well, in 2017 if people don't have C99 compliance (which was discussed in draft form for 2+ years before approval), then they're almost 20 years out-of-date...  Hopefully this is adequate.

src/portable_endian.h
pprindeville: Built and installed.  Seems to work, though I'm not sure about all the places where/how endianess is used...

pprindeville:
The Cygwin concern has been addressed and now `__CYGWIN__` is tested ahead of `HAVE_ENDIAN_H` or `HAVE_SYS_ENDIAN_H`.

bmah888:
@pprindeville:  Once again, thanks for your help!  I hope to be able to look at the latest version today, but please feel free to poke me if I don't respond in a reasonable time.

@ralcini:  I seem to remember you did some development on Cygwin.  Would you be able to take a look at this pull request (or `master`, if I merge it before you get a chance to look at it) and be sure that Cygwin building still works?  Thanks!

pprindeville:
Pleasure to do it!

bmah888:
Merged, thanks again!

pprindeville:
Awesome, thanks!

PR #667, opened 2017-11-08
Remove a few unnecessary headers and unused functions.

millert:
While working on a port of iperf to mingw64 I noticed some unneeded headers and functions.

The following headers are not needed, iperf uses no definitions from them:

sysexits.h -- EXIT_SUCCESS and EXIT_FAILURE come from stdlib.h, not sysexits.h
pthread.h -- nothing is used from pthreads.h either

The following functions are not referenced anywhere:

get_tcp_windowsize() -- iperf now uses get/setsockopt() with SO_SNDBUF SO_RCVBUF directly
set_tcp_windowsize() -- same as above
delay() -- not used and removing this drops a dependency on nanosleep()


bmah888:
Thanks for the PR!  That definitely cleans up some cruft that we no longer need.  Hoping to do a few tests with it soon.

bmah888:
Merged, thanks for the PR.

On an unrelated note I'd also like to take the opportunity to thank you for your work on sudo!

PR #669, opened 2017-11-14
Add missing saved_errno dance around some close(2) calls.

millert:
Many of the close() calls in the error path preserve the original errno of the failed system call but not all.  This commit preserves the original errno around calls to close() and freeaddrinfo() so that things like iperf_strerror() will report the correct error string.

### DIFF
--- a/src/iperf_sctp.c
+++ b/src/iperf_sctp.c
@@ -437,6 +437,7 @@ iperf_sctp_bindx(struct iperf_test *test, int s, int is_server)
     int nxaddrs;
     int retval;
     int domain;
+    int saved_errno;
 
     domain = test->settings->domain;
     xbe0 = NULL;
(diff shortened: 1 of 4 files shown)


src/iperf_sctp.c
```diff
@@ -525,8 +545,10 @@ iperf_sctp_bindx(struct iperf_test *test, int s, int is_server)
     }
 
     if (sctp_bindx(s, xaddrs, nxaddrs, SCTP_BINDX_ADD_ADDR) == -1) {
+        saved_errno = errno;
```
bmah888: We need a declaration of `int saved_errno` at the top of this function (`iperf_sctp_bindx()`) or compilation blows up here (and again three lines lower).

millert: Looks like I missed one, sorry about that.

bmah888: No worries, thanks for fixing!  Admittedly I could have just gone and fixed this myself post-merge, but I'm also using this to learn about the GitHub pull request review workflow, which is still new to me.

millert: Verified last change after installing the sctp packages on Linux.

bmah888: It also fixes the build on FreeBSD.  Looks like we're good to go.  Thanks again!

PR #670, opened 2017-11-15
Make help and manpage a bit clearer about TOS values

klausman:

master

Make help and manpage a bit clearler about TOS values on the cmdline

Since iperf3 uses strtol() to parse these values, the user can specify
them with 0 prefix for octals and 0x for hex values.


klausman:
I've added similar information/reference to --dscp in the the manpage and --help output.

Issue #712, opened 2018-03-09
Authentication functions added to libiperf


FuzzyStatic:
# Context

* Version of iperf3: 3.5

* Hardware: Any

# Enhancement Request

* Current behavior

There are no authentication functions included in libiperf

* Desired behavior

I would like to see authentication functions included in libiperf.

1. A function to input the username
2. A function to input the password
3. A function to input the key in a single line (base64 encoded)

FuzzyStatic:
@ralcini can you help with this?

ralcini:
@FuzzyStatic can you explain me better how you expect to use these new function?

FuzzyStatic:
There are getters and setters for all the parameters:

```
/* Getter routines for some fields inside iperf_test. */
int	iperf_get_verbose( struct iperf_test* ipt );
int	iperf_get_control_socket( struct iperf_test* ipt );
int	iperf_get_test_omit( struct iperf_test* ipt );
int	iperf_get_test_duration( struct iperf_test* ipt );
char	iperf_get_test_role( struct iperf_test* ipt );
int	iperf_get_test_reverse( struct iperf_test* ipt );
int	iperf_get_test_blksize( struct iperf_test* ipt );
FILE*	iperf_get_test_outfile( struct iperf_test* ipt );
uint64_t iperf_get_test_rate( struct iperf_test* ipt );
int     iperf_get_test_burst( struct iperf_test* ipt );
int	iperf_get_test_socket_bufsize( struct iperf_test* ipt );
double	iperf_get_test_reporter_interval( struct iperf_test* ipt );
double	iperf_get_test_stats_interval( struct iperf_test* ipt );
int	iperf_get_test_num_streams( struct iperf_test* ipt );
int	iperf_get_test_server_port( struct iperf_test* ipt );
char*	iperf_get_test_server_hostname( struct iperf_test* ipt );
char*	iperf_get_test_template( struct iperf_test* ipt );
int	iperf_get_test_protocol_id( struct iperf_test* ipt );
int	iperf_get_test_json_output( struct iperf_test* ipt );
char*	iperf_get_test_json_output_string ( struct iperf_test* ipt );
int	iperf_get_test_zerocopy( struct iperf_test* ipt );
int	iperf_get_test_get_server_output( struct iperf_test* ipt );
char*	iperf_get_test_bind_address ( struct iperf_test* ipt );
int	iperf_get_test_udp_counters_64bit( struct iperf_test* ipt );
[...]

PR #713, opened 2018-03-11
Issue #712 - Auth support on libiperf

ralcini:
Added the following methods on libiperf as requested on #712 
```
iperf_set_test_client_username(struct iperf_test *ipt, char *client_username)
iperf_set_test_client_password(struct iperf_test *ipt, char *client_password)
iperf_set_test_client_rsa_pubkey(struct iperf_test *ipt, char *client_rsa_pubkey_base64)
```

* A small refactor was needed
* A direct feedback is requested on the iperf_reset_test behaviour. In the previous implementation is obvious that authtoken must be reset: but what about username, password and publickey setup? I supposed to reset them accordingly.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -62,6 +62,10 @@
 #include "queue.h"
 #include "cjson.h"
 
+#if defined(HAVE_SSL)
+#include <openssl/bio.h>
+#endif // HAVE_SSL
+
 typedef uint64_t iperf_size_t;
 
 struct iperf_interval_results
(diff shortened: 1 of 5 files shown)


bmah888:
Thanks @ralcini for the pull request.  I need to catch up on this, plus the other comments in #712.

bmah888:
@FuzzyStatic:  Do these API changes export enough functionality for your use case?  I'm just reviewing the code by looking at it, but not actually trying to write something to use it.

FuzzyStatic:
@bmah888 Yes these API changes have everything I need.

src/iperf.h
```diff
@@ -62,6 +62,10 @@
 #include "queue.h"
 #include "cjson.h"
 
+#if defined(HAVE_SSL)
+#include <openssl/bio.h>
```
fortyandtwo: Adding `#include <openssl/evp.h>` here will prevent the compiler from warning about the implicit declaration of [EVP_PKEY_free](https://travis-ci.org/esnet/iperf/jobs/351984688#L574)

bmah888: I agree with this comment.

ralcini: ok, just added. Sorry for the delay

PR #737, opened 2018-05-17
Now the it can output a valid JSON string.

K9A2:

```
```

Merge to branch master.

* Issues fixed (if any):

NO.

No it can output an valid JSON string when using '-J' and '-V' at the same time.

Current version of iperf3 will ouput the string: 'Control connection MSS ...'  when using JSON output at client side. When users redirect the output to a JSON file, by using command such as 'iperf3 -c 127.0.0.1 -J -V 1> sample.json', and parse the sample.json later, it failed because that string is unparseable. Therefore, I added an condiction at the 'if' to judge whether it is using JSON output before printing verbose output. Moreover, I also modified the help information to clearify it.

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -363,8 +363,9 @@ iperf_connect(struct iperf_test *test)
 		}
 	}
 
-    if (test->verbose) {
-	printf("Control connection MSS %d\n", test->ctrl_sck_mss);
+    // Now it will not print control connection message when using JSON output
+    if (test->verbose && !test->json_output) {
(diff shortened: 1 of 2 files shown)


bmah888:
I'm not sure how I feel about this.  I agree there's an interaction between `-v` and `-J` that needs to be fixed, but it's not clear to me what the best solution is (it would never have occurred to me to do verbose output and expect valid JSON output).  It's very cumbersome to add a check for JSON output for every debug or verbose output, maybe we can just make that combination an error?  Or make `-J` automatically disable `-v`?

bmah888:
I'm going to close this and fix it a different way.  Thanks for bringing the point up!

PR #775, opened 2018-07-24
Fixed function implementation

LikHait:
Fixed implementation iperf_set_test_reverse.
Example of wrong behavior:
```
test->role = 'c';
iperf_set_test_reverse(test, 1);
*****
iperf_set_test_reverse(test, 1);
```
Expected value:
```
test->sender == 0;
test-> reverse == 1;
```
Current value:
```
test->sender == 1;
test-> reverse == 1;
```
------------------------------------
Also corrected unnecessary checks:
`test->sender && test->sender_has_retransmits => test->sender_has_retransmits == 1`

If !sender => ` test->sender_has_retransmits == -1`

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -428,8 +428,10 @@ iperf_set_test_num_streams(struct iperf_test *ipt, int num_streams)
 static void
 check_sender_has_retransmits(struct iperf_test *ipt)
 {
-    if (ipt->sender && ipt->protocol->id == Ptcp)
-	ipt->sender_has_retransmits = has_tcpinfo_retransmits();
+    if (ipt->sender && ipt->protocol->id == Ptcp && has_tcpinfo_retransmits())
+	ipt->sender_has_retransmits = 1;
+    else
+	ipt->sender_has_retransmits = 0;
 }
 
 void
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2492,7 +2492,7 @@ iperf_reset_stats(struct iperf_test *test)
         rp->bytes_sent_omit = rp->bytes_sent;
         rp->bytes_received = 0;
         rp->bytes_sent_this_interval = rp->bytes_received_this_interval = 0;
-	if (test->sender && test->sender_has_retransmits) {
+	if (test->sender_has_retransmits == 1) {
 	    struct iperf_interval_results ir; /* temporary results structure */
 	    save_tcpinfo(sp, &ir);
 	    rp->stream_prev_total_retrans = get_total_retransmits(&ir);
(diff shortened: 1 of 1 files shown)


LikHait:
> 2. Why test->sender_has_retransmits == 1 and not just test->sender_has_retransmits?

test->sender_has_retransmits is NOT Boolean. It can take the values 1, 0 and -1

test->sender_has_retransmits == -1 is mean that current side is receiver.

> 1. You're saying we can remove test->sender because that's implied by test->sender_has_retransmits?

I think that in this case - yes. We can check it if add debug output as 
```
if (test->sender_has_retransmits == 1) {
	    printf("point1\n");
```
> Do we need to set ipt->sender_has_retransmits = 0 at the top of this function to make sure it has some defined value by time the function exits?`

We already initialize it, calling `check_sender_has_retransmits`, at the beginning of test

bmah888:
Thanks for replying to my questions.  I think I didn't point you to the right part of the code for that last question/feedback (probably due to my inexperience with GitHub)...I'm going to try that again in a moment.  Also, would you say that this change is independent of your pull request #780 for the bidirectional tests?

src/iperf_api.c
```diff
@@ -428,10 +428,8 @@ iperf_set_test_num_streams(struct iperf_test *ipt, int num_streams)
 static void
 check_sender_has_retransmits(struct iperf_test *ipt)
 {
-    if (ipt->sender && ipt->protocol->id == Ptcp && has_tcpinfo_retransmits())
-	ipt->sender_has_retransmits = 1;
-    else
-	ipt->sender_has_retransmits = 0;
+    if (ipt->sender && ipt->protocol->id == Ptcp)
```
bmah888: This is the right place.  I'm wondering if we should initialize `ipt->sender_has_retransmit` to zero just above this line?  Otherwise it's possible to run through the function without setting it...the original version of the code always set this value one way or another.

LikHait: Oh, you are looking at outdated code (I reset it cause saw undefined behaviour).

LikHait:
> Also, would you say that this change is independent of your pull request #780 for the bidirectional tests?

These changes were aimed at correcting the implementation iperf_set_test_reverse. Bidirectional contains  these commits, but most have been changed.

LikHait:
> What I'm wondering is that if check_sender_has_retransmits() should explicitly set ipt->sender_has_retransmits to zero before the conditional? The prior implementation always set some value one way or another, but the new code might leave the value unset.

I can't find what you're talking about. Perhaps you are looking at the outdated code? I canceled the changes of `check_sender_has_retransmits ()` in the 399fe5f

bmah888:
>I can't find what you're talking about. Perhaps you are looking at the outdated code? I canceled the changes of check_sender_has_retransmits () in the 399fe5f

Yep, I'm an idiot.  I was looking at old code.


Sigh.  I'm an idiot.  Yes I was looking at the old code.

Issue #778, opened 2018-08-03
Segmentation fault.


LikHait:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the iperf3 mailing list at iperf-dev@google-groups.com or
question sites such as Stack Overflow
(http://www.stackoverflow.com/).  A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3: esnet/master

* Operating system (and distribution, if any): Linux, Ubuntu 18.04 LTS

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report

* Expected Behavior

* Actual Behavior
Segmentation fault on server side

* Steps to Reproduce
iperf3 -s -V
iperf3 -c localhost -P 2 -R

* Possible Solution
https://github.com/Aquantia/iperf/pull/8/commits/f1778271d2c6bed1dae09dccd332511b263f6871

LikHait:
Yep, it looks good. #780 also contains these changes

PR #784, opened 2018-08-22
Fix a segmentation fault in verbose mode.

bmah888:
Porting the fix for #778 from Aquantia/iperf.


bmah888:
Cool, thanks for the review.  Merging.

Issue #768, opened 2018-07-06
iperf3 3.5 TCP option -n not working


paaguti:
# Context

* Version of iperf3:
iperf3.5

* Hardware:
 amd64 laptop with Intel i5 CPU,

* Operating system (and distribution, if any):
Running on QEMU based VMs
Host OS: Ubuntu 16.04.4LTS
Guest OS: Alpine Linux 3.7+testing

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report

* Expected Behavior
iperf3 -c <IP> -n 1.8M should produce a data transaction with 1.8M bytes

* Actual Behavior
Dumping from the PCAP file and filtering the FIN/ACK packet:
```
       time       sport      dport          seq              ack      flags
     0.19117      56562       5201               38          2904690  FA
```
Expecting something around 1887436 and not 2904690 in the ack number

* Steps to Reproduce
VM1 as server, `iperf3 -s` ,capture packets with tshark
VM2 as client, `iperf3 -c <IP_VM1> -n 1.8M`
Analyse TCP SEQ and ACK numbers

* Possible Solution
Yet N/A

paaguti:
Confirming that it doesn't work for `--reverse` mode either.

thomas-fossati:
I have observed the same anomalous behaviour as @paaguti running `iperf 3.5 (cJSON 1.5.2)` on `Darwin 17.6.0`.

I have attached a plot that shows the distribution of the observed transfers (in green) around the expected value (in red-ish), in my case 1Mbyte. See in particular the odd spike around 1.25M.

bmah888:
1.8MB is a pretty small transfer compared to what iperf3 was designed for.  At that size you'll run into some effects where the default sending length (`-l` flag) will be significant, because iperf3 only does writes (at the socket layer) of that particular length...it's 128K by default for TCP transfers.

Still I see some non-intuitive behavior even on my local macOS laptop sending to itself, and it feels like there's something not quite right here.  I'm particularly puzzled by the fact that the number of bytes sent is different on different runs...if we ran for a certain length of time (which is the default mode of operation) I'd expect that, but not with `-n` specified.

bmah888:
This line (around line 2205 in `src/iperf_api.c`) appears to be the culprit:
```
    testp->multisend = 10;	/* arbitrary */
```
[...]

PR #797, opened 2018-09-24
CLI backward compatible fix for issue #768

paaguti:

```
```

  master

* Issues fixed (if any):
  #768 

Provide a backwards compatible approach to using the -n option.
-n burst leaves multisend untouched
-n burst/multisend sets the value for multisend.
It's like with the -b flag

This work is done in the context of the MAMI Project, which is receiving funding from the European Union’s Horizon 2020 research and innovation programme under grant agreement No 688421.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -920,6 +920,22 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 		client_flag = 1;
                 break;
             case 'n':
+				/* take the same approach as with -b to make it backwards compatible */
+				/* -b <burst_size> as legacy
+				   -b <burst_size>/<multisend> for new behaviour
+				   -b <burst_size>/1 should make iperf3 repeatable */
+						slash = strchr(optarg, '/');
+		if (slash) {
+		    *slash = '\0';
+		    ++slash;
+		    test->settings->burst = atoi(slash);
+		    if (test->settings->burst <= 0 ||
+		        test->settings->burst > MAX_BURST) {
+			i_errno = IEBURST;
+			return -1;
+		    }
+		}
+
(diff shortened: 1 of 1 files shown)


paaguti:
The second alternative directly writes the multisend field in the test. This is what I did originally and it seems to be more reliable than setting the burst

bmah888:
Hi @paaguti...thanks for the pull request.  I didn't quite understand your last comment...are you saying that you need to do some more work on this code change?

paaguti:
Hi,

first I tried setting test->settings->burst, it looked more intuitive for me. But I discovered that it doesn't  prevent multiple bursts. Therefore I went back to my original rationale and set test->multisend and everything works as expected (no more changes needed) . 

I have left both steps in the GitHub, because I think it would be nice to understand and document the behaviour depending on both variables. 

They seem interdependent, because when test->settings->burst not zero, it will control the send loop the same way as test->multisend does, but there are more implications I don't quite understand yet.

Maybe I'm just confusing you with my thoughts, rumblings and ways to arrive to knowledge... sorry if it is the case.

bmah888:
I haven't forgotten about this.  The PR is actually larger than it has to be, because there are a number of whitespace-only changes in the code...all of the significant changes are really in the argument-processing section of `iperf_api.c`.  Because there were merge conflicts, I started playing with a change that's just the argument processing.  But I noticed I still wasn't getting exactly deterministic results...they seem "better" in an informal way.  I did try rearranging some of the logic in send loop to see if that makes a difference, and I'm not sure.

The result of my changes are on the issue-768 branch.  There's some diagnostic code in there as well that would eventually go away.

bmah888:
Stumbled across #982 while trying to test this.

PR #804, opened 2018-10-25
Add delay parameter. This change adds a -z/delay parameter which spec…

JanJSokolowski:
Add delay parameter. This change adds a -z/delay parameter which specifies how many seconds to sleep after connecting to server. This is useful if running many simultaneous tests, as problems establishing new connections if a test is already running have been found.

Signed-off-by: Jan Sokolowski <jan.sokolowski@intel.com>


JanJSokolowski:
Hi!
What would be the preference in submitting patches? If the patch adds a new functionality, should we add it in the bugtracker (similar to bug reports) or just submit a patch?

Thanks for any input,

Regards
Jan Sokolowski  
jan.sokolowski@intel.com

PR #813, opened 2018-11-07
Increase maximum concurrent streams

leorosen:
Instead of the hard-coded MAX_STREAMS macro, discover the maximum file
descriptor count available to the process, and leave some descriptors in
reserve;

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -312,6 +312,7 @@ struct iperf_test
     double remote_cpu_util[3];                     /* cpu utilization for the remote host/client - total, user, system */
 
     int       num_streams;                      /* total streams in the test (-P) */
+    int       max_streams;			/* maximum number of streams */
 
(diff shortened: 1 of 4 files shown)


leorosen:
Fair enough. Now, how do you feel about WASTING half of your file descriptions on mmap, just because you forget to close them after mmap has returned?
If you really feel FDs are so precious, I think you would be ashamed to ware them like that. 

{ Leo }

bmah888:
Actually we can't close the file descriptors used for the mmap(2) calls, because they're needed for the `--zerocopy` option (because sendfile(2) requires an open file descriptor to send the payload as a "file").  One could validly make an argument that we could close them in the non-`--zerocopy` case.  I'll have to think about whether that's a good thing to do  It might be.  But as far as I can tell or recall, nobody "forgot" to close them.

PR #827, opened 2018-12-07
tcp sctp: Let the system pick the largest possible backlog value for `listen()`

lenormf:
Hi,

* Issues fixed (if any): Passing 5 as a value for the backlog parameter of the `listen()` call is insufficient when testing large amounts of sessions, so we let the system truncate `INT_MAX` down to whatever the maximum value is supported.

* Brief description of code changes (suitable for use as a commit message): This commit applies the same changes made by b481169 (#693), to the TCP and SCTP server sockets.

### DIFF
--- a/src/iperf_sctp.c
+++ b/src/iperf_sctp.c
@@ -37,6 +37,7 @@
 #include <netdb.h>
 #include <sys/time.h>
 #include <sys/select.h>
+#include <limits.h>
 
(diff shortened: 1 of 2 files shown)


lenormf:
Not sure how to fix this `timer_t` failed test, it didn't fail with the first commit.

bmah888:
I'm not sure what's the deal with the `timer_t` test failure...I'm not able to reproduce it.

Issue #831, opened 2018-12-26
Iperf3 server with listening port 0 does not print its listening port


jamofer:
# Context

* Version of iperf3: iperf 3.1.3

* Hardware: X86_64

* Operating system (and distribution, if any): Kubuntu 18.04

# Bug Report

* Expected Behavior
```
iperf3 -B 127.0.0.1 -s -p 0
-----------------------------------------------------------
Server listening on 39653
-----------------------------------------------------------
```
* Actual Behavior
```
-----------------------------------------------------------
Server listening on 0
-----------------------------------------------------------
```
* Steps to Reproduce
iperf3 -B 127.0.0.1 -s -p 0

* Possible Solution
#832

bmah888:
The fact that `--port 0` (or `-p 0`) is accepted is actually a bug.  There's no reason to allow the iperf3 server to bind itself to a random port. I'm going to file an issue for that.

MrMino:
@bmah888 Port 0 is a standard way of saying "kernel, please give me an unused port". It is expected by users as it's something standard to Linux.

First use-case that comes to mind would be automation of iperf on remote hosts. I need to use port 0 for that, as I'm never sure what port would be open for me.

This would also make it possible to make simple scripts that run performance tests without the need of specifying a port.

Port 0 is expected and solves a valid use-case. Why reject it? Why _defend_ against it?

jamofer:
In fact that was my use case. It's really strange to put a new feature limiting the application usage in my opinion.

PR #832, opened 2018-12-26
print current listening port when specified port is 0

jamofer:

* Issues fixed (if any): #831

print current listening port when specified port is 0


bmah888:
The case where this can happen is that the user does something like `iperf3 --server --port 0`.  This forces the iperf3 server to allocate an ephemeral TCP port.  I consider that to be a bug, in other words, `--port 0` shouldn't be allowed in the first place.  There's no reason to let the server bind itself to a random port.

MrMino:
@bmah888 This isn't just "random" port. It's guaranteed to be an unused port. That's why it's useful.

PR #837, opened 2019-01-11
Fix duplicate sum JSON objects in bidirectional mode reports

mmenshchikov:
master

* Issues fixed (if any):
In bidirectional mode, "sum" / "sum_received" / "sum_sent" objects are duplicated, and most JSON parsers don't allow duplicate keys. Hence it is impossible to parse report fully.

Introduce "sums" array for bidirectional mode which acts as a container for "sum" / "sum_received" / "sum_sent" objects. Add it to both intervals and end objects.
For each interval in 'intervals', new JSON structure is:
"sums": [ <sum1>, <sum2> ]
For 'end', new structure is:
"sums': [ { "sum_sent", "sum_received" }, { "sum_sent", "sum_received" } ]

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -347,6 +347,7 @@ struct iperf_test
     cJSON *json_connected;
     cJSON *json_intervals;
     cJSON *json_end;
+    cJSON *json_end_sums;
 
     /* Server output (use on client side only) */
     char *server_output_text;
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the pull request.  I see the problem you pointed out with `--json` output, that's a real issue. It looks like your PR fixes the problem but all I did was glance at the code briefly. Hope to have some time to dig into this more soon.

omrisaba:
Hi
Is it right that in version 3.8 this issue is still not fixed?
Any ETA when the json issue will be fixed for --bidir option?

PR #854, opened 2019-03-30
Fix configure when openssl is absent.

hhb:
openssl flags was appended to normal cpp/ld flags even if it is absent. They should really be appended if openssl is detected. This change fixes configure.ac and regenerates configure.


bmah888:
Thanks for the pull request.  This looks like it fixes a real bug, albeit not one we see very often.

PR #855, opened 2019-03-30
Fix warnings for ILP32

hhb:


bmah888:
I wonder if it'd be a good idea for me to import a new cjson, which would have the side effect of taking care (I think) of the first part of the problem.

hhb:
It would always be a good idea to follow upstream. For this particular case, [upstream](https://github.com/DaveGamble/cJSON/blob/master/cJSON.c#L389) is using INT_MAX. This is a conservative setting. But I guess it works until we need a buffer larger than 2G?

hhb:
Wait... Seems upstream never used LLONG_MAX. No matter in [1.5.2](https://github.com/DaveGamble/cJSON/blob/de5df3e56f2d3edb077937107c7c17adee5aa1ab/cJSON.c#L342) (the version used here) or latest. Am I looking at the right upstream? Or this is a local change?

bmah888:
Yeah I wasn't thinking very clearly when I wrote my last comment.  Updating to a new cjson might still be a good idea for a number of reasons, but we have some local changes.  In particular, we need to be able to encode and decode 64-bit integers in JSON numeric values.  

I put a comment on DaveGamble/cJSON#300, which is (someone else's) request for that functionality in the upstream cjson sources.

bmah888:
I'm in the process of importing a new version of cJSON.  The process of adapting those sources to run as a part of iperf3 will include a modified version of the changes from this PR, so I'm going to close this PR at this point.  Thanks for the submission and for keeping us honest with respect to ILP32!

Issue #751, opened 2018-06-08
v3.5 error - control socket has closed unexpectedly


bcronje:
# Context
I use iperf3 as part of an automated test environment using ansible. We run various tests over different network conditions.

* Version of iperf3: iperf 3.5 (cJSON 1.5.2)

* Hardware: lxd containers

* Operating system (and distribution, if any): Ubuntu 16.04.3

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

# Bug Report
**Randomly** I get the following error in the json report:
`"error":	"error - control socket has closed unexpectedly"`
When it happens, it seems to always occur after the second last interval, e.g. using `-t 60` the json report will contain the interval report for interval 58 - 59, and then ends with the error. So the 60th interval is never reported on.

The server is run with `iperf3 -s -D`
The client is run with `iperf3 -c 1.1.1.1 -t 60 -P 8 -J`

Unfortunately I do not have access to the output of the server.

Note - I never experienced this when I was still using iper3 version that ships with Ubuntu, think it was v3.1.x ?? This might only be coincidence or a regression.

* Expected Behavior
Expects to complete the test without any errors.

* Actual Behavior
Test ends with `"error":	"error - control socket has closed unexpectedly"`

* Steps to Reproduce
It occurs randomly, so not easy to reproduce.

bmah888:
Increased the timeout on the control channel as per the first commit in PR #859, which hopefully will solve this issue.

A further enhancement could be to allow setting this as a command-line parameter, although note that the timer that is affected by this has no effect for the `--bytes` or ``--blockcount` cases.

Closing.

PR #859, opened 2019-04-19
Issue 751

acooks:
* Fixes issue #751

* Reduce likelihood of triggering server cleanup timeout before results can be exchanged, which happens when the round trip time is hundreds to thousands of milliseconds.

---
This change is [Reviewable](https://reviewable.io/reviews/esnet/iperf/859)

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -270,6 +270,9 @@ create_server_timers(struct iperf_test * test)
 {
     struct iperf_time now;
     TimerClientData cd;
+    int max_rtt = 4; /* seconds */
+    int state_transitions = 10; /* number of state transitions in iperf3 */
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR!  I hope to get around to studying the code change sometime soon...your description of the problem and solution make sense given the symptoms that have been reported.

bmah888:
Because I'm trying to finish off some things, I've just gone ahead and cherry-picked the first of the two commits from this pull-request to the iperf3 master branch.  Thanks for the pull-request!

acooks:
Ok. I believe the second commit is essential, but haven't had time to trace
the path that demonstrates that point again. It should be somewhat obvious
that the server takes longer to close the connection and cleanup between
multiple runs in a simple manual test.

PR #868, opened 2019-05-16
Fix getting json server output in bidir mode

srgnk:
In bidirectional mode, if option --get-server-output is set and if
both client and server have --json set to true, client would still print
the json output of server to stdout as a separate piece instead of
including it into client's json output.

This patch fixes this problem, the server's json output would be
appended to client's json field 'server_output_json' as it should be.

There is probably better approach to fix this but that also works and the fix is pretty small.

To reproduce the issue, you can start the server as follows: 
`iperf3 -s --json`

and client:
`iperf3 -c <server> --json --get-server-output --bidir`

You will see that server's JSON output would be printed as below before the client's output. However it should be included into client's JSON output.
`
Server JSON output:
{
        "start":        {
                "connected":    [{
<...>
`

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -3325,7 +3325,7 @@ iperf_print_results(struct iperf_test *test)
             }
 
             /* Print server output if we're on the client and it was requested/provided */
-            if (test->role == 'c' && iperf_get_test_get_server_output(test)) {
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR.  Actually whether the server has `--json` or not doesn't really matter, either way the client ends up printing stuff it shouldn't.

PR #881, opened 2019-06-11
Fix off by one overwrite crash and memory leaks.

acmay:
The base64 decode will crash on musl c-library builds for OpenWRT
due to the write of the '\0' past the end of the allocated buffer.
Fix other various memory leaks on the authentication code paths.

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1757,7 +1757,7 @@ get_parameters(struct iperf_test *test)
 	if (test->debug) {
             char *str;
             str = cJSON_Print(j);
-	    printf("get_parameters:\n%s\n", str );
+            printf("get_parameters:\n%s\n", str );
             free(str);
 	}
 
@@ -2400,8 +2400,6 @@ iperf_free_test(struct iperf_test *test)
         SLIST_REMOVE_HEAD(&test->streams, streams);
         iperf_free_stream(sp);
     }
-    free(test->remote_congestion_used);
-    test->remote_congestion_used = NULL;
     if (test->server_hostname)
 	free(test->server_hostname);
     if (test->tmp_template)
@@ -2538,7 +2536,8 @@ iperf_reset_test(struct iperf_test *test)
 
     SLIST_INIT(&test->streams);
 
-    free(test->remote_congestion_used);
+    if (test->remote_congestion_used)
+        free(test->remote_congestion_used);
     test->remote_congestion_used = NULL;
     test->role = 's';
     test->mode = RECEIVER;
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1246,8 +1246,10 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 
         char *client_password = NULL;
         size_t s;
-        if ((client_password = getenv("IPERF3_PASSWORD")) == NULL &&
-            iperf_getpass(&client_password, &s, stdin) < 0){
+        /* Need to copy env var, so we can do a common free */
+        if ((client_password = getenv("IPERF3_PASSWORD")) != NULL)
+             client_password = strdup(client_password);
+        else if (iperf_getpass(&client_password, &s, stdin) < 0){
             return -1;
         } 
 
@@ -1265,6 +1267,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
         test->settings->client_password = client_password;
         test->settings->client_rsa_pubkey = load_pubkey_from_file(client_rsa_public_key);
     }
+    free(client_rsa_public_key);
 
     if (test->role == 'c' && (server_rsa_private_key || test->server_authorized_users)){
         i_errno = IESERVERONLY;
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the pull request.  I admittedly have not studied it yet.  Paging @ralcini for comments.

src/iperf_api.c
```diff
@@ -1545,11 +1551,17 @@ int test_is_authorized(struct iperf_test *test){
...
+            free(username);
+            free(password);
             return 0;
         } else {
             iperf_printf(test, report_authetication_failed, username, ts);
+            free(username);
+            free(password);
             return -1;
         }
+        free(username);
```
ralcini: I think that is not needed (lines 1563 and 1564).. redundant to the "else"

src/iperf_api.c
```diff
@@ -1743,7 +1755,10 @@ get_parameters(struct iperf_test *test)
         r = -1;
     } else {
 	if (test->debug) {
-	    printf("get_parameters:\n%s\n", cJSON_Print(j));
+            char *str;
+            str = cJSON_Print(j);
+	    printf("get_parameters:\n%s\n", str );
+            free(str);
```
ralcini: suggested in order to avoid any memory leak con cJSON_Print function?

src/iperf_api.c
```diff
@@ -1908,7 +1923,9 @@ send_results(struct iperf_test *test)
 		}
 	    }
 	    if (r == 0 && test->debug) {
-		printf("send_results\n%s\n", cJSON_Print(j));
+                char *str = cJSON_Print(j);
+		printf("send_results\n%s\n", str);
+                free(str);
```
ralcini: same doubt of block 1758-1761

src/iperf_api.c
```diff
@@ -2381,7 +2400,8 @@ iperf_free_test(struct iperf_test *test)
         SLIST_REMOVE_HEAD(&test->streams, streams);
         iperf_free_stream(sp);
     }
-
+    free(test->remote_congestion_used);
+    test->remote_congestion_used = NULL;
```
ralcini: in actual master this is already handled with lines 2413-2414
`if (test->remote_congestion_used)`
`	free(test->remote_congestion_used);`

src/iperf_api.c
```diff
@@ -2498,6 +2538,8 @@ iperf_reset_test(struct iperf_test *test)
 
     SLIST_INIT(&test->streams);
 
+    free(test->remote_congestion_used);
+    test->remote_congestion_used = NULL;
```
ralcini: I suggest:
`if (test->remote_congestion_used)`
`free(test->remote_congestion_used);`

ralcini:
thank you @acmay for the fixes!

src/iperf_api.c
bmah888: I did a little research and the memory from `cJSON_Print()` actually does need to be explicitly freed (see the description of cJSON_Print() in cJSON's README file).  In other parts of iperf3, we use cJSON_Delete() but I think that's wrong...I think cJSON_Delete() is intended for use on JSON objects, not arbitrary strings.  I believe that cJSON_free() is intended for this use, per https://github.com/DaveGamble/cJSON/issues/5#issuecomment-298469697.  I'm going to open an issue for the use of cJSON_free() (which apparently didn't exist when iperf3 was originally written).

src/iperf_api.c
bmah888: Same comment from me about `cJSON_free()`.

src/iperf_api.c
bmah888: I agree @ralcini:  Lines 1563 and 1564 will never be reached.

src/iperf_api.c
acmay: free(NULL) is always safe to do as per the man pages. I don't see the benefit of doing the if for the extra lines and risk of a typo between the 2.

bmah888: You have a perfectly valid point about `free(NULL)` being safe to do.  However the rest of iperf3 does this sort of check...I want to try to keep some kind of consistency in the code idioms within iperf3, even if they might be suboptimal.

bmah888:
As a general comment, thanks for working up these fixes and for taking the time to make a pull request.  Just a few details to be ironed out.  I'd like to get these fixes merged soon-ish for inclusion in a 3.7 release.

src/iperf_api.c
```diff
@@ -1273,12 +1276,15 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
         !(server_rsa_private_key && test->server_authorized_users)) {
          i_errno = IESETSERVERAUTH;
         return -1;
-    } else if (test->role == 's' && server_rsa_private_key && test_load_private_key_from_file(server_rsa_private_key) < 0){
-        i_errno = IESETSERVERAUTH;
-        return -1;
-    } else {
+    } else if (test->role == 's' && server_rsa_private_key == NULL) {
```
bmah888: The check for `server_rsa_private_key` got inverted here. The net effect is that authentication succeeds even if the wrong password is given.

bmah888:
I pulled this PR into a new branch, `pr-881`, so I could fix the logic inversion in the test for `server_rsa_private_key` and test it, as well as a few stylistic changes.

I'm hoping to get the result merged soon-ish after some more testing.

bmah888:
Merged a modified version of this pull request as #887.  Thanks for the fixes!

PR #898, opened 2019-07-16
fix burst mode throttle checking

olichtne:
  `3.1-STABLE`) to which this pull request applies: **master**

* Issues fixed (if any):

While investigating a different iperf issue with udp stream tests I noticed that using -b 0/1000 results in iperf sending 1000 packets and then just idle for the remainder of the test duration.
The full command I was using to reproduce the issue:
```iperf3 -c 192.168.101.2 -B 192.168.101.1 -b0/1000 -u -t 10 -l 1400```

When burst mode is configured for unlimited rate (-b0) but with a
specific packet burst value, iperf only sends packets once, after that
the iperf_check_throttle function gets called and sets green_light=0
due to the rate value being 0 and average calculated rate always being
higher than that.

This can be fixed by moving the "don't throttle" condition
directly inside the iperf_check_throttle function.

Signed-off-by: Ondrej Lichtner <olichtne@redhat.com>

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1397,7 +1397,7 @@ iperf_check_throttle(struct iperf_stream *sp, struct iperf_time *nowP)
     double seconds;
     uint64_t bits_per_second;
 
-    if (sp->test->done)
+    if (sp->test->done || sp->test->settings->rate == 0 || sp->test->settings->burst != 0)
         return;
     iperf_time_diff(&sp->result->start_time_fixed, nowP, &temp_time);
     seconds = iperf_time_in_secs(&temp_time);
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR.  I'm trying to remember what are the semantics exactly of doing burst mode on an unlimited-rate UDP transfer?  In other words what's the difference between `-b0` and `-b0/1000`?  Without having thought about it a lot, it feels like both of them are supposed to spew packets out as fast as they can, so the idea of a burst size doesn't really mean anything.

olichtne:
Yes, I also thing that both are supposed to "spew packets out as fast as they can" which is the reason for my pull request. Which I assume means that the ```iperf_check_throttle``` function shouldn't be getting called with this combination of parameters.

On master, the condition for "should I do throttling" is only in one place - before calling the throttle function and in the send burst loop. However, the throttling function is also called after this loop at which point it calculates that the target rate '0' (as configured through ```-b 0```) has been exceeded and red-lights the generation stream after the first burst.

This means that the ```-b 0/1000``` combination basically only sends 1000 packets in a single burst and then does nothing until an end-condition is met, in case of an end condition for number of packets sent that just means it'll loop until infinity; in case a duration it'll wait until the expiration.

olichtne:
I'll just repeat that this is related to the issue #899 where I can workaround slow udp generation by using the ```-b 0/1000``` arguments together if I have this fix. I reported the issue separately as it might mean that a different fix is required for that, but if not then this does work as a workaround.

bmah888:
Got it.  I'm going to have to find some time to stare at this code for a little while (as well as your bug report in #899).  Thanks.

PR #908, opened 2019-08-28
Force type check on JSON objects that are blindly used as string type

marian-cingel:
The commit tries to prevent a situation when iperf_server treats
a received JSON object as a 'valuestring' without additional check,
while client can send something else - a number, null, ...
Passing the 'object->valuestring' into 'strdup' then ends with fault.

Add 'cJSON_GetObjectItemOfType', 'cJSON_GetObjectItemCaseSensitiveOfType' wrappers.

### DIFF
--- a/src/cjson.c
+++ b/src/cjson.c
@@ -1738,9 +1738,21 @@ CJSON_PUBLIC(cJSON *) cJSON_GetObjectItem(const cJSON * const object, const char
     return get_object_item(object, string, false);
 }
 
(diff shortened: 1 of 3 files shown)


bmah888:
Thanks for the pull request!  Type-checking seems like a useful thing to have in the upstream cJSON (we import a modified version of that library for use in iperf3).  I would recommend that you submit the type-check functions as a pull request to the DaveGamble/cJSON project on GitHub, that way they'd be available to everyone.  If that shows up, then we can see about using those functions in iperf3.

I'm going to close this for now, because I think this is really more of a cJSON issue.  But I'd be happy to use those functions if they were provided upstream.  Thanks again!

PR #913, opened 2019-09-19
iperf3.7: Compatibility for Android

khubaibumer:
Add "--stream-path" for user-defined temporary stream path

Signed-off-by: khubaibumer <khubaibumer@gmail.com>

```
```

3.7+ (cJSON 1.5.2)

* Issues fixed (if any):

Set Stream Path to user-defined path
 To work on Android as /tmp is not available on Android
 also access to a static /data/local/tmp/ is also restricted for Applications
 Only native application path is accessable with complete permissions

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -126,6 +126,7 @@ struct iperf_interval_results
     long rtt;
     long rttvar;
     long pmtu;
+    long reorder;
 };
 
 struct iperf_stream_result
@@ -137,6 +138,7 @@ struct iperf_stream_result
     atomic_iperf_size_t bytes_sent_omit;
     long stream_prev_total_retrans;
     long stream_retrans;
+    long stream_reorder;
     long stream_max_rtt;
     long stream_min_rtt;
     long stream_sum_rtt;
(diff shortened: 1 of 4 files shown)


khubaibumer:
> Thanks for the pull request! I'm wondering what the advantage is of setting `--stream-path` on the command line, when iperf3 already recognizes the `TEMPDIR`, `TEMP`, and `TMP` environment variables for setting that temporary directory.

As I've mentioned in my commit that "TEMPDIR" in Android's case is "/data/local/tmp" and its access is restricted for zygote processes. By using "--stream-path" any application can use its "native directory" instead. Also resetting TMPDIR/TMP/TEMP can break their systems as iperf can be just a part of their whole project. I believe --stream-path is much better option than using TEMPDIR/TMP/TEMP for all restricted systems

hardikbhalodi:
I build IPerf3 shared library for Android and its working but when i do this command
"iperf3 -c ping-90ms.online.net" It's Blocking until complete output does not fetch and printing all lines at once but In Ubuntu i installed IPerf3 and its working asynchronously.

any help on this?

hugoinfante83:
This week i went through all the versions of iperf 3, and since iperf 3.2 it stop working in my application cache directory,  any  version in 3.1.x with the temp changed to private cache will work.
I confirm that i can run on iperf3.2 with temp = data/local/tmp using adb shell only.

PR #921, opened 2019-10-21
Added Generic Segmentation Offload support for UDP protocol.

yazimut:
Generic Segmentation Offload allows you to reduce the load on the CPU by transferring routine work from the kernel to the network card. This option is activated using the key "--udp-gso N", where "N' is the segment size.
Linux kernel began supporting UDP GSO from version 4.18.0. Also it could be software or hardware supporting. Hardware supporting could be only if network card and its driver are supporting UDP GSO. You may check it using "ethtool -k <your interface>" - find "tx-udp-segmentation" parameter. If it equals "on" - hardware supporting is enabled, if "off" - disabled, if "off [fixed] - your interface doesn't support GSO".
e.g. On the AQC107 10Gbps card, that supports UDP GSO, CPU loading is falling from 100% to 25% (about these values). Bitrate is increasing from 300Mbps to 2.6Gbps (about these values) - comparing "-l 100" and "-l 6400 --udp-gso 100".

```
```

  `3.1-STABLE`) to which this pull request applies: latest


gegles:
Please, take a look at https://github.com/esnet/iperf/pull/1925, hopefully this can become the definitive PR for this feature

bmah888:
PR #1925 was committed with this functionality. Thanks for your submission!

PR #923, opened 2019-10-24
implement forceflush for json

ikreb7:
Hello,
I use iperf to test stuff and don't want wait too long until the json output returned. There are only few changes at the code. I use the ``forceflush``-flag and print immediately the current json content. I look forward to your feedback. 

```
```

  `3.1-STABLE`) to which this pull request applies: latest

There are only few changes at the code. I use the ``forceflush``-Flag and print immediately the current json content.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -672,6 +672,10 @@ iperf_on_test_start(struct iperf_test *test)
 {
     if (test->json_output) {
 	cJSON_AddItemToObject(test->json_start, "test_start", iperf_json_printf("protocol: %s  num_streams: %d  blksize: %d  omit: %d  duration: %d  bytes: %d  blocks: %d  reverse: %d  tos: %d", test->protocol->name, (int64_t) test->num_streams, (int64_t) test->settings->blksize, (int64_t) test->omit, (int64_t) test->duration, (int64_t) test->settings->bytes, (int64_t) test->settings->blocks, test->reverse?(int64_t)1:(int64_t)0, (int64_t) test->settings->tos));
+        if (test->forceflush) {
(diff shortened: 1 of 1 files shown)


h0ax-ki1l3r:
@ikreb7 
I was trying to test this feature. I tried to test with `forceflush` with `-J` flag did not output the file.
E.g. 
`iperf3 -s -J --forceflush 10 --logfile test.json`

It printed something on the screen, however the results where not getting flushed into test.json every 10 seconds. Could you please elaborate more on what this PR tries to accomplish. 

> print immediately the current json content.

More specifically, I am confused and not sure what you mean by print immediately.

bmah888:
I'm trying to understand the value of this change.  For better or for worse, iperf3 currently emits its output with the `--json` flag as a single JSON object at the end of the run.  If we flush the output in the middle of the run (what your change does), the output won't be valid JSON.  So any consumer of this output still won't be able to parse it until the end of the run anyway.

rchromatic:
Hello,
I found this thread while I was looking for the same feature. I tested the patch but **unfortunately these changes do not cover the server side. The server still prints "intervals" only at the end of the session.**

@bmah888 
I need the output from iperf3 during the speed test, to show it on UI. It allows users to see the progress and be sure that speed test is working. Right now I use non JSON output, but wanted to switch to JSON to make parsing more clean.

> So any consumer of this output still won't be able to parse it until the end of the run anyway.

iperf3 should print valid JSON, which is done in this patch. It adds extra {} before printing "intervals".

@hoax-killer 
>I am confused and not sure what you mean by print immediately.

Iperf3 should print result once the data is available. It should not wait until the end of the speed test session.

Issue #153, opened 2014-03-21
ldconfig needed in make install?


bmah888:
@lomaxfrog just ran into a problem on an Ubuntu Linux system where a manual invocation of ldconfig was needed after make install in order for the iperf3 binary to find its shared library.  This was on iperf 3.0.2.

This issue is to investigate what steps we should be doing in this case.  Might be a little tricky because so much of the Makefile stuff is auto-generated.

bmah888:
This really is a generic software installation problem.  We're not going to solve it here, given the difficulties in doing this across multiple platforms.  Closing without a fix.

kriswebdev:
If you want to stick to `/usr/local/lib` instead of `/usr/lib32` on Unbuntu distro, then just run `ldconfig /usr/local/lib` (requires superuser) at the end of `make install`.

See: http://stackoverflow.com/questions/4743233/is-usr-local-lib-searched-for-shared-libraries

Chadster766:
Thanks running ldconfig after "make install" solved the issue for me.

nzjc:
Running Debian 7.8 64Bit (wheezy). Had the issue as prescribed in the title, ran 'sudo ldconfig' post make install, working fine now. Thanks for the help in this thread.

tomturner123:
Had to use this on latest version (3.1.3) with ubuntu. If this is to be a necessary requirement for all ubuntu versions I suggest its worthwhile making it a bit more obvious so that the user wont need to search google / github to get going

jangwhan:
same here with Ubuntu 16.04 (Xenial Xerus). 
I understand this is caused by cross-platform issue(to which I have no clues as to why), but as tomturner123 said, ubuntu better make warning signs about this, unless someone like me get lost again....

jmack51:
Does anyone have a fix for the iperf3 error on Mac OSX?  With 3.2, I'm seeing libiperf.so.0 is not found.

bmah888:
@jmack51:  On mac OS, there won't be a `*.so.0` shared library...mac OS uses a `*.dynlib` extension for its shared libraries.  If "the Python wrapper" refers to iperf3-python, that's a separate project and you probably ought to take this up with them (I'm not sure if there's a bug here or not).

bmah888:
@jmack51:  Ah, disregard, I see where you already opened an issue with thiezn/iperf3-python.

jmack51:
[...]

PR #934, opened 2019-11-23
Fix runtime linking when installed outside default prefix (#153)

hpjansson:
When libiperf was installed in a directory not searched by default (e.g. /usr/local/lib on some systems), iperf3 would fail to find it at runtime. Pass -rpath when linking iperf3 to remedy this. The problem was originally discovered on Ubuntu Linux.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -49,7 +49,7 @@ libiperf_la_SOURCES     = \
 iperf3_SOURCES          = main.c
 iperf3_CFLAGS           = -g
 iperf3_LDADD            = libiperf.la
-iperf3_LDFLAGS          = -g
+iperf3_LDFLAGS          = -g -rpath $(libdir)
 
 if ENABLE_PROFILING
 # If the iperf-profiled-binary is enabled (and this condition is true by default)


hpjansson:
I ran into this while looking for the best practice solution for a similar situation in one of my projects. This is the solution I ended up with. It's a small change, so I thought maybe it'd work for you too.

bmah888:
Thanks for the PR...that's very helpful.  Wondering if anyone who packages iperf3 for their OS/distro has any feedback/comments on this change?

hpjansson:
I don't package iperf3 myself, but I have experience with packaging for openSUSE. With my packager hat on, I'd probably patch out the -rpath to make linking consistent with the rest of the platform and because the package installs to a standard location.

A possible compromise would be to make it optional (on by default) and add a --without-rpath flag to configure.

Historically, using -rpath to link with libraries in common paths has been frowned upon, because the DT_RPATH entry makes binaries impossible to relocate -- i.e. it would always look for the library at the full path first and if found, would ignore LD_LIBRARY_PATH and other attempts to override it. Newer ld (since ~2017) adds DT_RUNPATH which has a lower priority than LD_LIBRARY_PATH, so it's less of a problem than it used to be.

The Debian wiki has a [useful summary](https://wiki.debian.org/RpathIssue). It's probably representative of how packagers think about the situation.

bmah888:
After a really long time, we decided to give up on #153 and not try to solve it, so I'm going to respectfully decline this pull request.  Thank you for the submission!

Issue #940, opened 2019-12-11
iperf_set_test_client_rsa_pubkey() API call always fails to set public key


oleh-yudin-ring:
# Context

* Version of iperf3:
3.7

# Bug Report

* Expected Behavior
When using iperf_set_test_client_ username/password/rsa_pubkey C API it is expected these data to be passed to server

This is not happening due to load_pubkey_from_base64() doesn't work as expected.
In addition there is a leak of bio pointer allocated data.
Also the same is related to load_privkey_from_base64().

* Actual Behavior
Authentication data is not sent to server (test->settings->authtoken is empty) for the test during test run

* Steps to Reproduce

1. Via C API setup test with iperf_set_test_client_ username/password/rsa_pubkey
2. Run test and see auth failure error

* Possible Solution
There is no need to do Base64Decode() cause PEM_read_bio_PUBKEY() already requires base64 key string in PEM format with corresponding header and footer to work correctly.
Or find a different way to set raw RSA key data into OpenSSL EVP_PKEY
The following solution worked for me:
```
EVP_PKEY *load_pubkey_from_base64(const char *buffer) {
    BIO* bio = BIO_new(BIO_s_mem());
    BIO_puts(bio, "-----BEGIN PUBLIC KEY-----\n");
    BIO_puts(bio, buffer);
    BIO_puts(bio, "\n-----END PUBLIC KEY-----\n");
    EVP_PKEY *pkey = PEM_read_bio_PUBKEY(bio, NULL, NULL, NULL);
    BIO_free(bio);
    return (pkey);
}
```

bmah888:
I don't do any OpenSSL programming myself.  I am trying to figure out what the semantics of `load_pubkey_from_base64()` are, or rather what the desired functionality is.

I would have thought that the input to this function (whatever its name is) would just have been the same as the file contents that `load_pubkey_from_file()` takes.  Instead it looks like it's trying to take the base64-encoded part of the file minus the beginning and ending lines.

@ralcini When you implemented these API call, what did you have in mind?  What is supposed to be in the block of memory pointed to by the `buffer` parameter?

@FuzzyStatic Have you tried using the new API calls?  Did this include `load_pubkey_from_base64()` and `load_privkey_from_base64()`?

bmah888:
@oleh-yudin-ring I think that your comment about leaking memory is referring to the fact that the `bio` buffer is not deallocated after `pkey` is created.  Is that right?

oleh-yudin-ring:
[...]

PR #949, opened 2020-01-04
Fix memory leaks for RSA key setters

oleh-yudin-ring:
* Version of iperf3: `master` branch

* Issues fixed (if any): #940(partially) memory leaks while using API calls to set RSA keys


bmah888:
Thanks for the fix!

PR #950, opened 2020-01-06
build: Unbreak builds on macOS by not building profiled binaries by default

### LINKED ISSUE #933: iPerf3 v3.7 no longer compiling under macOS with './configure; make'; Now requires '--disable-profiling' flag
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the iperf3 mailing list at iperf-dev@googlegroups.com or
question sites such as Stack Overflow
(http://www.stackoverflow.com/).  A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3:

**v3.7**

* Hardware:

**iMac (Retina 5K, 27-inch, 2019)**
3 GHz Intel Core i5
32 GB 2667 MHz DDR4

**macos 10.14.6 with all patches applied**

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

**Xcode 11.2.1**
```
$ clang --version
Apple clang version 11.0.0 (clang-1100.0.33.12)
Target: x86_64-apple-darwin18.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report

* Expected Behavior

When running usual `./configure; make` expected code to compile as it always has.

* Actual Behavior

Compilation fails.  Relevant pieces in output is as follows:
```
main.c:140:8: warning: 'daemon' is deprecated: first deprecated in macOS 10.5 - Use posix_spawn APIs instead. [-Wdeprecated-declarations]
                rc = daemon(0, 0);
                     ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/stdlib.h:293:6: note: 'daemon' has been explicitly marked deprecated here
int      daemon(int, int) __DARWIN_1050(daemon) __OSX_AVAILABLE_BUT_DEPRECATED_MSG(__MAC_10_0, __MAC_10_5, __IPHONE_2_0, __IPHONE_2_0, "Use po...
         ^
1 warning generated.
```
...
```
libtool: link: gcc -pg -g -g -O2 -Wall -pg -g -o .libs/iperf3_profile iperf3_profile-main.o iperf3_profile-cjson.o iperf3_profile-iperf_api.o iperf3_profile-iperf_error.o iperf3_profile-iperf_auth.o iperf3_profile-iperf_client_api.o iperf3_profile-iperf_locale.o iperf3_profile-iperf_server_api.o iperf3_profile-iperf_tcp.o iperf3_profile-iperf_udp.o iperf3_profile-iperf_sctp.o iperf3_profile-iperf_util.o iperf3_profile-iperf_time.o iperf3_profile-dscp.o iperf3_profile-net.o iperf3_profile-tcp_info.o iperf3_profile-timer.o iperf3_profile-units.o  ./.libs/libiperf.dylib
clang: error: the clang compiler does not support -pg option on versions of OS X 10.9 and later
make[2]: *** [iperf3_profile] Error 1
make[1]: *** [all] Error 2
make: *** [all-recursive] Error 1

```
* Steps to Reproduce

1.  Download iperf source from here:  https://github.com/esnet/iperf/releases/tag/3.7
2.  Decompress .tar.gz into respective `iperf-3.7` directory.
3.  `cd iperf-3.7`
4.  `./configure; make`
5.  Boom!

* Possible Solution

Searching through the source code using `grep -R "\-pg" *`, I found that this flag is set in two files:
  * src/Makefile.am
  * src/Makefile.in

In those files just before the flags are set, it reads in the comments
```
if ENABLE_PROFILING
# If the iperf-profiled-binary is enabled (and this condition is true by default)
# Specify the sources and various flags for the profiled iperf binary. This
# binary recompiles all the source files to make sure they are all profiled.
iperf3_profile_SOURCES  = main.c \
                          $(libiperf_la_SOURCES)

iperf3_profile_CFLAGS   = -pg -g
iperf3_profile_LDADD    = libiperf.la
iperf3_profile_LDFLAGS  = -pg -g
endif
```

Doing more digging, I found there is a flag for `./configure` which is `--disable-profiling`.  And yes, if you do `./configure --disable-profiling; make`, the binary compiles just fine.

So as the error message indicates that the clang compiler doesn't support the `-pg` flag anyway, possibly make it that for macOS, the default is to disable profiling.  I believe this likely would simply require modifying the ./configure script, but as there are multiple references to `darwin` in that script, I have to tap out at this point as only those who code this would know the potential side effects/etc. or where the code change would need to go.

_Please submit patches or code changes as a pull request._

# Enhancement Request

Not sure if I would classify this as an enhancement request, but as the typical `./configure; make; make install` behavior, which worked up until sometime this summer at least (the last time I compiled iPerf3 from source on some new systems), no longer works as it did before, maybe this is more an enhancement than anything?

* Current behavior

* Desired behavior

* Implementation notes

_If submitting a proposed implementation of an enhancement request,
please use the pull request mechanism._

bmah888:
Fixes #933.

### DIFF
--- a/configure
+++ b/configure
@@ -1416,7 +1416,7 @@ Optional Features:
   --enable-fast-install[=PKGS]
                           optimize for fast installation [default=yes]
   --disable-libtool-lock  avoid locking (might break parallel builds)
-  --disable-profiling     Disable iperf profiling binary
+  --enable-profiling      Enable iperf profiling binary
 
 Optional Packages:
   --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
@@ -12135,12 +12135,12 @@ if test "x$GCC" = "xyes"; then
   CFLAGS="$CFLAGS -Wall"
 fi
 
-# Check if profiling must be disabled
+# Check if enable profiling
 # Check whether --enable-profiling was given.
 if test "${enable_profiling+set}" = set; then :
-  enableval=$enable_profiling; :
+  enableval=$enable_profiling; enable_profiling=yes
 else
-  enable_profiling=yes
+  :
 fi
 
  if test x$enable_profiling = xyes; then
--- a/configure.ac
+++ b/configure.ac
@@ -52,11 +52,11 @@ if test "x$GCC" = "xyes"; then
   CFLAGS="$CFLAGS -Wall"
 fi
 
-# Check if profiling must be disabled
+# Check if enable profiling
 AC_ARG_ENABLE([profiling],
-    AS_HELP_STRING([--disable-profiling], [Disable iperf profiling binary]),
-    [:],
-    [enable_profiling=yes])
+    AS_HELP_STRING([--enable-profiling], [Enable iperf profiling binary]),
+    [enable_profiling=yes],
+    [:])
 AM_CONDITIONAL([ENABLE_PROFILING], [test x$enable_profiling = xyes])
 
 # Checks for header files.
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -52,7 +52,7 @@ iperf3_LDADD            = libiperf.la
 iperf3_LDFLAGS          = -g
 
 if ENABLE_PROFILING
-# If the iperf-profiled-binary is enabled (and this condition is true by default)
+# If the iperf-profiled-binary is enabled
 # Specify the sources and various flags for the profiled iperf binary. This
 # binary recompiles all the source files to make sure they are all profiled.
 iperf3_profile_SOURCES  = main.c \
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -664,7 +664,7 @@ iperf3_CFLAGS = -g
 iperf3_LDADD = libiperf.la
 iperf3_LDFLAGS = -g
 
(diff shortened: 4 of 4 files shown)


asy972:
I built 3.7 with "configure --disable-profiling". This option (--disable-profiling) suddenly enabled the iperf3_profile build in 3.8.1.

bmah888:
> I built 3.7 with "configure --disable-profiling". This option (--disable-profiling) suddenly enabled the iperf3_profile build in 3.8.1.

Sorry about that..this was fixed (after iperf-3.8.1) with pull request #1018.  iperf-3.9 (when released) will contain this fix.

PR #954, opened 2020-01-17
include iperf_util.h in main.c

tamird:
73b02f9 implemented `daemon(3)` for systems that don't have it, but the
function prototype was never included in `main.c` where it is
referenced.

@bmah888


tamird:
@bmah888 ping.

PR #957, opened 2020-01-23
fix urandom being left open

chris-hawley:

* Issues fixed (if any): This file gets left open throughout both the c-api and iperf command program lifespan. Neither seem to close the file properly leading to a memory leak. The fix is more important for the c-api because the leak won't go away at the end of each iperf iteration. I wasn't exactly sure how often this function is used or if there is a better place to stick the file closing code. However, this change does eliminate the memory leak related to urandom being left open. 

Side note: There is another leak related to json which i'm currently trying to find a fix for. This looks like a cJSON issue but I'll continue investigating. 

* Brief description of code changes (suitable for use as a commit message): Fixed urandom being left open

### DIFF
--- a/src/iperf_util.c
+++ b/src/iperf_util.c
@@ -75,6 +75,8 @@ int readentropy(void *out, size_t outsize)
                       rndfile,
                       feof(frandom) ? "EOF" : strerror(errno));
     }
+    fclose(frandom);
+    frandom = NULL;
     return 0;
 }
 


bmah888:
Thanks for the pull request, and apologies for the delay in responding to you.

The code opening `/dev/urandom` predates me, but I think what's happening is that the file object is declared as `static` so that there's only one instance of that variable.   The idea is that the file only gets opened one time, on the first time the function `readentropy()` is called.  Repeated calls to `readentropy()` will continue to use the same open file.  It's true that the file never gets closed, and technically this is a leak, but it's only one file object and file descriptor.  It's done this way as a performance optimization, to avoid having to allocate and deallocate `FILE` objects and opening/closing the file multiple times.

I'm going to leave this code the way it is.  I would be interested in the other memory issue you found; the cJSON we're using is rather old and it probably needs an update.

Thanks again!

chris-hawley:
I agree that this fix would impact performance negatively. Ideally, it would be freed at the end of the library execution with a dedicated destruct function. 

It's been a while since I looked at this but from what I remember the current version of cJSON had the same issue. I could be wrong. I'll look at it again and see where it ended up.

Issue #914, opened 2019-09-29
Negative values in loss report with UDP


danielaCasasv:
**# Context**
* Version of iperf3: 3.7+
* Hardware: Intel
* Operating system: Ubuntu14.04.6 server running on VM (virtualBox 5.2.28)

**# Bug Report**
Got a report with negative values in the lost_packets and losr_percent counters at server side logfile.

* Expected Behavior:
`{
			"streams":	[{
					"socket":	6,
					"start":	19.000102,
					"end":	20.000105,
					"seconds":	1.0000029802322388,
					"bytes":	166520,
					"bits_per_second":	1332156.0298656528,
					"jitter_ms":	2.8512847190910904,
					"lost_packets":	0,
					"packets":	115,
					"lost_percent":	0,
					"omitted":	false,
					"sender":	false
				}],
			"sum":	{
				"start":	19.000102,
				"end":	20.000105,
				"seconds":	1.0000029802322388,
				"bytes":	166520,
				"bits_per_second":	1332156.0298656528,
				"jitter_ms":	2.8512847190910904,
				"lost_packets":	0,
				"packets":	115,
				"lost_percent":	0,
				"omitted":	false,
				"sender":	false
			}`
The above stream report just appeared as expected in the same logfile than the next stream report (the one with negative values) 
* Actual Behavior
`{
			"streams":	[{
					"socket":	6,
					"start":	21.000194,
					"end":	22.000068,
					"seconds":	0.99987399578094482,
					"bytes":	372136,
					"bits_per_second":	2977463.1729218694,
					"jitter_ms":	2110.7105786449633,
					"lost_packets":	-111,
					"packets":	146,
					"lost_percent":	-76.027397260273972,
					"omitted":	false,
					"sender":	false
				}],
			"sum":	{
				"start":	21.000194,
				"end":	22.000068,
				"seconds":	0.99987399578094482,
				"bytes":	372136,
				"bits_per_second":	2977463.1729218694,
				"jitter_ms":	2110.7105786449633,
				"lost_packets":	-111,
				"packets":	146,
				"lost_percent":	-76.027397260273972,
				"omitted":	false,
				"sender":	false
			}`

* Steps to Reproduce
command server side: iperf3 -s -p 11002 -1 -J --logfile serverOut_11_to_2.json
command client side: iperf3 -c 10.0.0.2 -p 11002 -u -b 1675.289k -t 40 -J --logfile clientOut_11_to_2.json

[...]

PR #962, opened 2020-02-25
Issue #914 - workaround  to protect from garbled packet counters

davidBar-On:

* Issues fixed (if any): #914

As described in issue #914 discussion, it seems that sometimes the first UDP packet is received truncated or garbled, so the packet counter is garbled.
The changes include:
1. Adding a guard well known text before and after pcount in the packet header and verifying in the receiving side that this text was received.  Only then pcount is regarded as valid.
2. Clear the header part of the receiving buffer before each receive, to make sure that guards will no be valid in case a truncated packet is received.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -357,6 +357,10 @@ struct iperf_test
 
 };
 
+/* Guard text before Packet Counter in UDP packets */
+#define PCOUNT_GURAD_BEFORE "BFOR"
+#define PCOUNT_GUARD_AFTER "AFTR"
+
 /* default settings */
 #define PORT 5201  /* default port to listen on (don't use the same port as iperf2) */
 #define uS_TO_NS 1000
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the pull request!  Do we know if this actually solves the problem originally reported in #914?  (I agree that if the cause is caused by corrupted test packets, this change should help the situation.)

One other consideration is this change is not backward compatible with older versions of iperf3 (because it's unconditionally changing the format of the UDP payload).  That problem can be solved or mitigated but it's just something we need to keep in mind.

davidBar-On:
The pull request does not solve the problem originally reported in #914, as actually a problem in interpreting the output, not a bug.  It may be useful to change the output to make it more clear.  I.e., in the following case:


175

davidBar-On:
The pull request does not solve the problem originally reported in #914, as actually this is a problem in interpreting the output, not a bug. It may be useful to change the output to make it more clear and more consistence with the positive lost count. I.e., in the following case:

41:54.4: [  8]  30.00-31.00  sec   242 KBytes  1.98 Mbits/sec  4.966 ms  2/173 (1.2%) 
41:55.4: [  8]  31.00-32.00  sec   247 KBytes  2.03 Mbits/sec  5.648 ms  **-2/173** (-1.2%)

actually 175 packets were received in the 31-32 interval (173-(-2)).  Therefore, it may be more better if the output for this interval will be:

41:55.4: [  8]  31.00-32.00  sec   247 KBytes  2.03 Mbits/sec  5.648 ms  **-2/175** (-1.2%)

Meaning: 175 packets were received, including two recovered packets.  This is probably better as in the previous interval the "2/173" means 173 packets were received and two were lost.

Regarding the changes in the pull request, they may resolve the problem were it seems that the first packet was not received properly.  I don't have a way to verify that but I also didn't find a bug in the code that suggests that maybe there was a buffer overflow or such kind of problem.  If changing the header is a problem, then checksum may be added at the end.  Other alternative may be to ignore the first few packets, assuming that there may be a problem with them, by adding a parameter that defines how many of the first packets to ignore (similar --omit, but ignoring a count of packets instead of count of seconds).

bmah888:
@davidBar-On Thanks for your suggestion on the output format...thinking about that...hmmm...

davidBar-On:
Closing - sent a replacement PR #1260 which is easier implementation and probably works better.

PR #968, opened 2020-03-18
Fix for possible leaks or double frees

mruprich:

```
```

* Issues fixed (if any):

   I ran a coverity scan on the iperf3 code and there are a couple of possible leaks or double frees in the iperf_sctp.c file. I tried to resolve these in this PR. These might very well be false positives but I think at least some of these make sense. Take a look and let me know what you think.

Thanks.

### DIFF
--- a/src/iperf_sctp.c
+++ b/src/iperf_sctp.c
@@ -479,6 +479,7 @@ iperf_sctp_connect(struct iperf_test *test)
     if (!TAILQ_EMPTY(&test->xbind_addrs)) {
         if (iperf_sctp_bindx(test, s, IPERF_SCTP_CLIENT)) {
             freeaddrinfo(server_res);
+            close(s);
             return -1;
         }
     }
@@ -502,7 +503,6 @@ iperf_sctp_connect(struct iperf_test *test)
         i_errno = IESENDCOOKIE;
         return -1;
     }
-    freeaddrinfo(server_res);
 
     /*
      * We want to allow fragmentation.  But there's at least one
@@ -522,6 +522,7 @@ iperf_sctp_connect(struct iperf_test *test)
         return -1;
     }
 
+    freeaddrinfo(server_res);
     return s;
 #else
     i_errno = IENOSCTP;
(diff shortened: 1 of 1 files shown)


src/iperf_sctp.c
```diff
@@ -486,16 +492,17 @@ iperf_sctp_connect(struct iperf_test *test)
...
     /* Send cookie for verification */
     if (Nwrite(s, test->cookie, COOKIE_SIZE, Psctp) < 0) {
 	saved_errno = errno;
 	close(s);
+	freeaddrinfo(server_res);
 	errno = saved_errno;
         i_errno = IESENDCOOKIE;
         return -1;
     }
+    freeaddrinfo(server_res);
```
bmah888, line 505: This addition here will cause a double-free if the `setsockopt()` call a few lines down fails...its error-handling code also does `freeaddrinfo(server_res)`.  Perhaps this new line should move below that block (but before the `return s`).

src/iperf_sctp.c
```diff
@@ -473,8 +477,10 @@ iperf_sctp_connect(struct iperf_test *test)
 
     /* clients must call bind() followed by sctp_bindx() before connect() */
     if (!TAILQ_EMPTY(&test->xbind_addrs)) {
-        if (iperf_sctp_bindx(test, s, IPERF_SCTP_CLIENT))
+        if (iperf_sctp_bindx(test, s, IPERF_SCTP_CLIENT)) {
+            freeaddrinfo(server_res);
```
bmah888, line 481: Should there be a `close(s)` somewhere inside this error handling case?

src/iperf_sctp.c
mruprich: Right, totally missed that.

src/iperf_sctp.c
mruprich: Again, you are right.

mruprich:
I fixed the things suggested earlier. Hopefully now it is OK. Thanks @bmah888

bmah888:
Sorry, I somehow missed your changes. This looks great.  I intend to merge this when I get a few cycles to do one more round of tests on it.  Thanks for the pull request!

PR #984, opened 2020-04-20
Resolved name clashes between windows.h and openssl (cygwin under win…

ralcini:
windows.h declares some const defined on openssl when using cygwin.
It can be sufficient do before importing openssl headers

#undef X509_NAME
#undef X509_CERT_PAIR
#undef X509_EXTENSIONS
#undef OCSP_REQUEST
#undef OCSP_RESPONSE

but using the CFLAGS -DNOCRYPT it solve the problem

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -123,6 +123,11 @@ else
 	  fi ]
     )
     if $have_ssl; then
+        case $host in
+           *-*-cygwin)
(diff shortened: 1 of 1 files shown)


bmah888:
Merged this, thanks for the pull request!

PR #997, opened 2020-05-15
fix: Fix memory management issues related to cJSON_Print().

### LINKED ISSUE #883: Audit use of cJSON_Delete
In the cJSON library, `cJSON_Delete` is intended for use to deallocate memory used to hold JSON structures.  In iperf3 usage, we have several cases where that API is also used to deallocate memory that was allocated by `cJSON_Print`, which is wrong.  The correct way to fix this is with `cJSON_free()`, which was added in cJSON 1.5.0-ish.

We need to audit the uses of `cJSON_Delete` in iperf3 to see which ones are correct and which ones need to be changed to `cJSON_free`.  Note that there does not appear to be any obvious problems caused by this incorrect API usage at this time, nevertheless this problem should be corrected.

bmah888:
Fixes #883.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1755,7 +1755,9 @@ send_parameters(struct iperf_test *test)
 	cJSON_AddStringToObject(j, "client_version", IPERF_VERSION);
 
 	if (test->debug) {
-	    printf("send_parameters:\n%s\n", cJSON_Print(j));
+	    char *str = cJSON_Print(j);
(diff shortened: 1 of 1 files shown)


scottmsilver:
Hi there I think this commit may break json_output_string since it NULLs it out (and frees it) before the client can grab it. I think you probably want to strdup the output of cJSON_Print() and then free that.  Later on in iperf_free_test looks json_output_string will get freed.

PR #1002, opened 2020-05-18
fix(auth): Fix a potential buffer overflow in iperf3 client.

### LINKED ISSUE #996: Possible stack corruption in iperf_auth.c: encode_auth_setting()
# Context
* Version of iperf3:
3.7, master
Linux

# Bug Report
In iperf_auth.c encode_auth_setting() function stack variable _text_ is limited to 150 bytes.
It is used to prepare iperf authentication credentials before sending them on iperf server. 
Using it in the following manner:
`sprintf (text, "user: %s\npwd:  %s\nts:   %ld", username, password, utc_seconds);`
is not safe.

Problem: 
Calling sprintf is not safe in this case due to username and password are passed via API and don't have determined length limit. This may lead to authenticate string _text_ exceeding 150 bytes and corrupting stack.

* Expected Behavior
iperf API to work with authenticate credentials of non-limited size

* Actual Behavior
Having username and/or password long enough will lead to stack corruption

* Possible Solution
Consider using dynamic allocation for text buffer with pre-calculated size in a way like this:
```
const char *auth_text_format = "user: %s\npwd:  %s\nts:   %ld";
const int text_len = snprintf(NULL, 0, auth_text_format, username, password, utc_seconds);
char *text = (char *)malloc(text_len + 1);
sprintf(text, auth_text_format, username, password, utc_seconds);
...
free(text);
```

bmah888:
This condition was only possible when configuration authentication
via the libiperf API.

While here, also fix a few other sundry issues:

* Remove arbitrary length limits on username and password.

* Improved error handling.

* Updated error messages for readability.

* Fixed minor typo in some identifiers.

Fixes #996.


bmah888:
This eventually got merged in two pieces...06e3f08d and 5b7b99ed.

PR #975, opened 2020-04-02
Adding UDP traffic profile options for variable packet size and delay between packets

davidBar-On:
* Version of iperf3 (or development branch, such as `master` or
  `3.1-STABLE`) to which this pull request applies:
Master

* Issues fixed (if any):
None - new functionality

* Brief description of code changes (suitable for use as a commit message):
Adding options to support different UDP traffic profiles to test behavior under different network conditions (not just throughput): variable packet length (-l) - random or fixed delta (--step), delay (--sleep) between packets or between burst of packets (--bundle).

PR #1004, opened 2020-05-21
Adding UDP traffic profile options for variable packet size and delay between packets

davidBar-On:
* Version of iperf3 (or development branch, such as master or
3.1-STABLE) to which this pull request applies:
Master

* Issues fixed (if any):
None - new functionality (resubmit of pull request #975)

Adding options to support different UDP traffic profiles to test behavior under different network conditions (not just throughput): variable packet length (-l) - random or fixed delta (--step), delay (--sleep) between packets or between burst of packets (--bundle)

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,10 @@ New options:
     -k, --blockcount #[KMG]   number of blocks (packets) to transmit (instead 
                               of -t or -n)
     -L, --flowlabel           set IPv6 flow label (Linux only)
+    -l, --length l1/[l2]      UDP - added optional range of lengths (per step or random)
+    --step                    UPD - delta between packets length when length range is specifid
+    --sleep n1[/n2]           UDP - wait time between packets or packets bundle
+    --bundle n                UDP - number of packets to send before sleep waiting time
 
 Changed flags:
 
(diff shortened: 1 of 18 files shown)


davidBar-On:
@bmah888 will you consider adding this pull request to iperf3?  Although the proposed enhancements are not intended for testing maximum throughput, they allow to set various traffic profiles that can be used to better understand the expected actual throughput and packets loss.  That may be especially important for testing throughput on unreliable networks or networks with low throughput.  (In `test_commands.sh` I have added some examples of using the new features.)

bmah888:
Sorry for the delay, but thanks for the pull request!  I've been thinking about this feature change for awhile.

I'm going to respectfully decline this code change.  I understand that it provides more flexibility in some test scenarios, it also adds a fair amount of complexity to the UDP tests (3 new command line flags and expanding the syntax of an existing one).  I'm concerned that if we merge this, we'll be adding functionality that doesn't really benefit our main consumers but will make the code harder to maintain (just by adding more to it).

An implementation issue is that the use of nanosleep() for timing will likely not work correctly in the case of multiple parallel UDP tests. This is an unfortunate consequence of iperf3's single-threaded design, the nanosleep() call, when executed, will briefly block all running tests.

davidBar-On:
OK, I understand.

A comment regarding the use of nanspllep() in case this will be needed in the future. The solution here estimates the actual minimum sleep time supported and send several packets between sleeps to if needed. E.g. if minimum sleep time is 15ms and interval between packets should be 5ms, then 3 packets will be sent between the sleeps.  The number of UDP streams can be added to the calculation, so if there are 2 streams, 6 packets will be sent in this example, and if a packet should be sent every 100ms, the sleep time (per stream) will be 50ms.

PR #1027, opened 2020-07-16
Hardening the systemd service

c79:


bmah888 (commented):
Thanks for the pull request!

I understand the goal of this change, but I don't have a lot of experience / background with systemd services to evaluate the submitted patch.  It'd be great if someone out there in iperf3 land could take a look and give their feedback.

PR #1033, opened 2020-08-04
fix: Plug various minor memory leaks (#1023)

### LINKED ISSUE #1023: Leaks when using iperf_client API
# Context

* Version of iperf3:
3.7

Linux

# Bug Report

* Expected Behavior
No file descriptor or memory leaks when using client API

* Actual Behavior
File descriptor and memory leaks per each test run via API

* Steps to Reproduce
Descriptors and memory leaks when using iperf_client API:
1. File
  a. When passing a logfile to the API with iperf_set_test_logfile() during iperf_free_test() corresponding test->outfile file descriptor is not being closed leading to a FD leak.
  b. There is no free() call for test->logfile string whih causes memory leak after each test run.

2. Sockets
Under normal circumstances when iperf client test is finalized successfully iperf_client_end() is being called within the API when entered DISPLAY_RESULTS state.
In case of a failure and early exit (for example with IECTRLCLOSE error) iperf_client_end() is not being called leaving stream sockets file descriptors opened forever.

3. Memory
Few more memory leaks were detected:
  a. the one mentioned in 1.
  b. iperf_auth.c load_pubkey_from_base64, there should be free(key) to release memory allocated by Base64Decode
  c. in case of API use with json_output if there was an error during the test (return -1) corresponding cJSON structures allocated with iperf_json_start() are not freed in iperf_run_client().

* Possible Solution
1. In iperf_free_test()  fclose(test->outfile) and free(test->logfile) (watch out for default case when test->outfile=stdout)
2. Call iperf_client_end() in case of test failure to close opened sockets. 
3. Call iperf_json_finish() in case of test failure to release all preallocated json output results.

Example:
```   
void
iperf_free_test(struct iperf_test *test)
{
    ...
    if (test->logfile)
    {
        free(test->logfile);
        if (test->outfile)
            fclose(test->outfile);
    }
    ...
}
```

```
int
iperf_run_client(struct iperf_test * test)
{
    ...
    //return -1;
    goto cleanup_and_fail; // in every place with failure return
    ...
cleanup_and_fail:
    iperf_client_end(test);
    if (test->json_output)
        iperf_json_finish(test);
    iflush(test);
    return -1;
}
```

```
EVP_PKEY *load_pubkey_from_base64(const char *buffer) 
{
    ...
    Base64Decode(buffer, &key, &key_len);
    ...
    BIO_write(bio, key, key_len);
    free(key);
    ...
}
```

bmah888:
Based on submissions by @oleh-yudin-ring.  Fixes #1023.

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -1,5 +1,5 @@
 /*
- * iperf, Copyright (c) 2014-2018, The Regents of the University of
+ * iperf, Copyright (c) 2014-2020, The Regents of the University of
  * California, through Lawrence Berkeley National Laboratory (subject
  * to receipt of any required approvals from the U.S. Dept. of
  * Energy).  All rights reserved.
@@ -478,7 +478,7 @@ iperf_run_client(struct iperf_test * test)
 
     /* Start the client and connect to the server */
     if (iperf_connect(test) < 0)
-        return -1;
+        goto cleanup_and_fail;
 
     /* Begin calculating CPU utilization */
     cpu_util(NULL);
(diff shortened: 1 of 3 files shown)


src/iperf_client_api.c
```diff
@@ -582,4 +582,12 @@ iperf_run_client(struct iperf_test * test)
     iflush(test);
 
     return 0;
+
+  cleanup_and_fail:
+    iperf_printf(test, "cleanup_and_fail\n");
```
rantala, line 587: Hi, this added output also when using the JSON output format, breaking parsing of the resulting JSON output (not a big issue in practice as there was some error anyways). Could this printf just be dropped?

bmah888: Wow that's a pretty rookie mistake...I shouldn't have left that debugging code in there (and to be clear this is totally my fault...it wasn't a part of the original patches in #1023).  I'll get rid of it.  Good catch, thanks for the review.

Issue #873, opened 2019-05-17
UDP throughput issue


borkra:
Have anybody tried to implement UDP throughput test using sendmmsg/recvmmsg on Linux?

Because currently iperf UDP performance test, tests more how fast particular CPU performs context switch from userspace to kernel and back (at least on a single core CPUs), than network throughput. 

During TCP test Iperf sends 128K bytes buffer in a single call, so 1 userspace - kernel transition per 128K
During UDP test Iperf sends 1470 bytes bytes buffer in a single call, so nearly 100 - userspace - kernel transition per 128K

I believe using sendmmsg/recvmmsg would allow for UDP test to have performance comparable to TCP.

borkra:
More like enhancement request, then a question.

bmah888:
iperf3 isn't about trying to compare the performance between TCP and UDP in terms of socket calls.  UDP normally requires one system call per packet sent, whereas TCP allows for some fragmentation...that's just the way the socket API works.

If someone's interested in implementing this, they should consider that sendmmsg and recvmmsg are Linux-specific (at least according to the manual page entries on these two system calls), and iperf3 runs on a variety of other platforms as well.

borkra:
Few things:

- Yes it's not trying compare performance in terms of socket calls but on some platform it end up being just that especially on slower arms. Thus the problem.

- UDP support fragmentation too up to with packet size 64k, I agree it rearely used.

- Safe to say rear tcp user sends 128K of data in one call, so both benchmark is somewhat artificial. I would just prefer them to be artifical the same way.

gegles:
Any progress on this? @borkra , I too would be very interested in seeing this happen.

davidBar-On:
I have sent the pull request #1034 with a proposed implementation to add `sendmmsg/recvmmsg` support for UDP (not supported when reading from a file or writing to a file).  An option `--send-recvmmsg` was added to enable the feature.  When enabled, `burst` number of packets (defined by the `-b` option) are all sent by one `sendmmsg` call.

In my machine (same one for the sender and receiver) improvement is up to about 15% and this is achieved when burst size is about 50.

PR #1034, opened 2020-08-09
Add sendmmsg/recvmmsg support

davidBar-On:
  3.8.1 master

* Issues fixed (if any):
#873 

Add **`--send-recvmmsg` option** to use `sendmmsg/recvmmsg` for sending and receiving UDP messages.
On the sender size `sendmmsg` is sending the `bust` (from the `-b` option) number of packets.
These changes are **not** supported when sending from a file or writing to a file.

A `HAVE_SEND_RECVMMSG` flag was added in `config.h.in` to indicate whether functions are supported on the platform.


davidBar-On:
@bmah888 I have change the PR to use `-Z` option for using `sendmmsg` for UDP, instead of adding a new option (I also rewrote the PR description).  I also enhanced implementation, so if `-Z` is not used for UDP, practically there should **not** be any change to the current iperf3 behavior.  Therefore the risk of adding these changes is low.

The UDP throughput enhancement achieved for high throughput interfaces is quite substantial, especially when the UDP messages are not large (witch is usually the case).

database64128:
@bmah888 Any plan to have this PR reviewed and merged?

src/iperf_udp.c
```diff
@@ -212,52 +279,111 @@ iperf_udp_recv(struct iperf_stream *sp)
...
+	    }
 
-	memcpy(sp->buffer, &sec, sizeof(sec));
-	memcpy(sp->buffer+4, &usec, sizeof(usec));
-	memcpy(sp->buffer+8, &pcount, sizeof(pcount));
+	    /* Sending messages and making sure all packets are sent */
+	    i = 0;	/* count of messages sent */
+	    r = 0;	/* total bytes sent */
+	    while (i < sp->sendmmsg_buffered_packets_count) {
+	        j = sendmmsg(sp->socket, &sp->msg[i], sp->sendmmsg_buffered_packets_count - i, MSG_DONTWAIT);
```
database64128, line 336: Before each `sendmmsg(2)` call, you should poll for socket write readiness. In my tests, this can significantly improve performance.

src/iperf_udp.c
```diff
@@ -212,52 +279,111 @@ iperf_udp_recv(struct iperf_stream *sp)
...
+		if (j < 0) {
+		    r = j;
+		    break;
+		}
```
database64128, line 340: Did you forget to `i += j;`?

database64128 (changes requested):

src/iperf_udp.c
davidBar-On: This is done below in line 349.  The operation is delayed since the original `i` value is required in the `while` body.

src/iperf_udp.c
davidBar-On: Can you add more details:
1. How do you do the polling?
2. From your experience, what should be done if the socket is not ready for write?  The current design of the code is that the function does not return before all was sent (or there was an error).  Do you suggest that in case the socket is not ready for write the function will return successfully, but without sending anything or before all was sent?
3. Do you understand why the method you suggest improve performance?  I am asking since in any case, iperf3 will retry sending.

Thanks

database64128: > Do you understand why the method you suggest improve performance?

In my case, I was doing raw syscalls in my Go program. The UDP socket opened by the Go runtime is in non-blocking mode. With `sendmsg(2)`, if the sending operation was going to block, `sendmsg(2)` would return `-EAGAIN` or `-EWOULDBLOCK`, which is handled by the Go runtime to poll for socket write readiness with `epoll`. The calling goroutine can then be parked by the runtime to free the OS thread. (My limited understanding of Go internals might be inaccurate.)

Now with `sendmmsg(2)`, according to the manual, _a nonblocking call sends as many messages as possible (up to the limit specified by vlen) and returns immediately._ By treating a non-complete return value the same way as `-EAGAIN` and `-EWOULDBLOCK`, that is, instead of immediately calling `sendmmsg(2)` again, I instructed the Go runtime to poll for write readiness before the next `sendmmsg(2)` call. [This change](https://github.com/database64128/swgp-go/commit/6676ce8dcb3be19f7bc6cb414afcd18af87b64f8) yielded a 10% increase in throughput.

> The current design of the code is that the function does not return before all was sent (or there was an error).

I'm not familiar with `iperf3`'s code base. I just read some code, and it seems to me that `iperf3` uses sockets in blocking mode for UDP tests. In this case, maybe it's better to simply drop the `MSG_DONTWAIT` flag, `sendmmsg(2)` would then only return when all messages have been sent. This saves even more syscall overhead.

davidBar-On: @database64128, thanks a lot for the detailed explanation.

I will have to check how easy it is to implement this.  To minimize iper3 design changes, the approach I took for `sendmmsg` is to accumulate packets ierf3 is sending and send them in bursts using `sendmmsg`.  It may be that instead of the `for` loop, `sendmmsg` can be called once.  In this case all the packets that were not sent can either be moved to the beginning of the buffer or ignored. (The issue with ignoring is that the packets are numbered, so the new packets numbering should start from the last successful packet sent.)

simon-barber-cisco: Any update on this? Improved UDP send/recv performance on iperf would be very helpful!

davidBar-On:
I have rebased using version 3.20+.  However, the previous version was pre-3.16 and did not use threads for sending/receiving the data.  Now, since threads are used, the CPU is the limiting factor on my computer, and I get the same results for UDP with or without `-Z`.   **Tests should be run on a more powerful environment** to see whether and how much using `sendmmsg` improves performance.

PR #1046, opened 2020-08-26
set TCP_NODELAY flag for ctrl socket on server side

### LINKED ISSUE #1045: Client doesn't send traffic in bi-directional UDP test
# Context

* Version of iperf3:

Current (August 26 2020) git HEAD 52d0de3bfcd37

* Hardware:

x86, iperf session over virtual LTE network over TUN devices

Ubuntu 20.04

I am running tests in a scenario where iperf traffic goes over a TUN device through a virtual LTE network using [srsLTE](https://github.com/srsLTE/srsLTE). An app-note on how to set this app can be found [here](https://docs.srslte.com/en/latest/app_notes/source/zeromq/source/index.html#zeromq-appnote). I couldn't reproduce the issue on a traditional network.

# Bug Report

The issue I am seeing is when bi-directional UDP traffic is generated at a very large rate. This rate is well above the capacity of the channel so packets are expected to be dropped, but it should not be zero. If I lower the expected UDP rate, the test runs ok. I assumed that the issue may arise due to dropped TCP packets from the control link (they are equally dropped like the UDP datagrams). However, since the issue arises on the client side and the server side works fine, I am thinking it might be something else.

* Expected Behavior

Both client and server shall start to generate and send UDP traffic at the requested rate.

* Actual Behavior

The client instructs the server to generate traffic. The server does so and I see traffic arriving at the client. However, the client itself does not send traffic (stops after first call to `iperf_udp_send()`).

* Steps to Reproduce

I am unable to reproduce this on a normal network. However, using the above mentioned virtual LTE network the issue can be reproduced with:
```
$ sudo ip netns exec ue1 iperf3 -c 192.168.3.1 --bidir -u -b 500M --debug --verbose -t 1
iperf 3.9+
Linux andre-xps15 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64
Control connection MSS 1448
Setting UDP block size to 1448
send_parameters:
{
        "udp":  true,
        "omit": 0,
        "time": 1,
        "parallel":     1,
        "bidirectional":        true,
        "len":  1448,
        "bandwidth":    500000000,
        "pacing_timer": 1000,
        "client_version":       "3.9+"
}
Time: Wed, 26 Aug 2020 10:54:42 GMT
Connecting to host 192.168.3.1, port 5201
      Cookie: bsdwbqufga62k3prewx2oxduqinc36b4fsji
      Target Bitrate: 500000000
SNDBUF is 212992, expecting 0
RCVBUF is 212992, expecting 0
Setting application pacing to 62500000
[  5] local 192.168.3.2 port 37764 connected to 192.168.3.1 port 5201
SNDBUF is 212992, expecting 0
RCVBUF is 212992, expecting 0
Setting application pacing to 62500000
[  7] local 192.168.3.2 port 39512 connected to 192.168.3.1 port 5201
Starting Test: protocol: UDP, 1 streams, 1448 byte blocks, omitting 0 seconds, 1 second test, tos 0


sent 1448 bytes of 1448, total 1448
pcount 1 packet_count 0
interval_len 62.150672 bytes_transferred 1448
interval forces keep
interval_len 62.150674 bytes_transferred 1448
interval forces keep
[ ID][Role] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5][TX-C]   0.00-62.15  sec  1.41 KBytes   186 bits/sec            1
[  7][RX-C]   0.00-62.15  sec  1.41 KBytes   186 bits/sec  0.000 ms  0/1 (0%)
Test Complete. Summary Results:
[ ID][Role] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5][TX-C]   0.00-62.15  sec  1.41 KBytes   186 bits/sec  0.000 ms  0/1 (0%)  sender
[  5][TX-C]   0.00-62.15  sec  0.00 Bytes  0.00 bits/sec  0.000 ms  0/0 (0%)  receiver
CPU Utilization: local/sender 99.6% (17.6%u/82.0%s), remote/receiver 0.0% (0.0%u/0.0%s)
CPU Utilization: local/receiver 99.6% (17.6%u/82.0%s), remote/sender 0.0% (0.0%u/0.0%s)
[  7][RX-C]   0.00-62.15  sec  0.00 Bytes  0.00 bits/sec  0.000 ms  0/0 (0%)  sender
[  7][RX-C]   0.00-62.15  sec  1.41 KBytes   186 bits/sec  0.000 ms  0/1 (0%)  receiver
```

Iperf server is started with the below command and finished as expected:
```
$ iperf3 -s --debug --verbose
....
sent 1448 bytes of 1448, total 2562450304
sent 1448 bytes of 1448, total 2562451752
sent 1448 bytes of 1448, total 2562453200
sent 1448 bytes of 1448, total 2562454648
sent 1448 bytes of 1448, total 2562456096
sent 1448 bytes of 1448, total 2562457544
iperf 3.9+
Linux andre-xps15 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64
```

Note that in the example the client runs in the netns and the server not (it runs in the default netns). This is needed to run both on the same machine. However, the same also happens if I swap client and server netns, i.e., run the server in the `ue1` netns and client in the default one.

* Possible Solution

I am currently debugging the stalled send on the client-side but wanted to post the issue anyway to receive any possible pointer in case this happened before. I am aware that this might be difficult to reproduce for third-parties in the current form and I am not even sure it's an iperf3 issue.

andrepuschmann:

current git HEAD

* Issues fixed (if any):

this is an attempt to fix #1045 
it might be a too niche use-case for the wider audience but it does fix the issue I am having in my setup

in scenarios with high packet loss rate, for example due
to overflowing the capacity of the channel, the TCP traffic
on the control socket might be delayed too much. In a
particular case with a high-bandwidth bi-directionl UDP
test that exceeded the channel capacity by far, the UDP traffic
generated at the server was already causing full tx queues.
this caused the "delayed" TCP control packet to the client to be
also dropped. thus, the client never entered the TEST_RUNNING state.

this patch sets the TCP_NODELAY flag on the server's control
socket and makes sure that the control traffic goes through the
client before the UDP stream is even started.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -119,6 +119,9 @@ iperf_accept(struct iperf_test *test)
     if (test->ctrl_sck == -1) {
         /* Server free, accept new client */
         test->ctrl_sck = s;
+        // set TCP_NODELAY flag to "flush" control messages as quickly as possible
+        int flag = 1;
+        setsockopt(test->ctrl_sck, IPPROTO_TCP, TCP_NODELAY, (char *) &flag, sizeof(int));
         if (Nread(test->ctrl_sck, test->cookie, COOKIE_SIZE, Ptcp) < 0) {
             i_errno = IERECVCOOKIE;
             return -1;


bmah888 (changes requested):
Thanks for the PR!  The basic idea is good.  We also need a bit of error-handling, so maybe wrap the `getsockopt()` call in a check for the return value being < 0, and if so set `i_errno = IESETNODELAY` (that error code already exists for a different purpose) and return...similar to how is done with the call to `Nread()` a few lines below.

Also it might be good to do something similar for the client side, the right spot is probably about 15 lines into the function `iperf_connect` in `iperf_client_api.c`.

bmah888:
I have an re-implementation of this (which incorporates my comments) on the `nodelay` branch. Comments (from anyone) appreciated.

bmah888:
Replaced by #1063

PR #1056, opened 2020-09-28
auth: Ensure 64-bit time_t works on 32-bit systems

awilfox:
  `3.1-STABLE`) to which this pull request applies:  3.9 and master

* Issues fixed (if any):

On a 32-bit PowerPC Linux system using musl libc (with 64-bit time_t),
the t_auth test fails because `long` is not the same type as `time_t`.

This patch uses an int64_t temporary value, which can be truncated to
32-bit if necessary.

### DIFF
--- a/src/iperf_auth.c
+++ b/src/iperf_auth.c
@@ -35,6 +35,8 @@
 #define _WITH_GETLINE
 #include <stdio.h>
 #include <termios.h>
+#include <inttypes.h>
+#include <stdint.h>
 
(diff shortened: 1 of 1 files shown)


bmah888 (approved):
Looks pretty good, thanks for the pull request!  I don't have any systems with 32-bit longs and 64-bit time_t, but I was at least able to test with 64-bit and 32-bit systems that I have, to verify that there are no regressions.

PR #1059, opened 2020-09-30
fix: Fix regression in #997 where JSON output was free-ed too early.

### LINKED ISSUE #1058: Regression with json_output_string from #997
Hi there I think this commit may break json_output_string since it NULLs it out (and frees it) before the client can grab it. I think you probably want to strdup the output of cJSON_Print() and then free that.  Later on in iperf_free_test looks json_output_string will get freed.

_Originally posted by @scottmsilver in https://github.com/esnet/iperf/pull/997#issuecomment-683512806_

bmah888:
It didn't make anything crash, but API clients wouldn't be able to
access the JSON output.

Bug pointed out, and fix suggested by, @scottmsilver.  Fixes #1058.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4214,13 +4214,18 @@ iperf_json_finish(struct iperf_test *test)
     if (test->server_output_text) {
 	cJSON_AddStringToObject(test->json_top, "server_output_text", test->server_output_text);
     }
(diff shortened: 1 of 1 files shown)


bmah888:
@scottmsilver Want to look this over? I think this fixes the problem in the way you suggested. Thanks for the hint!

scottmsilver (approved):

PR #1063, opened 2020-10-23
enh: Set TCP_NODELAY on control connections.  Reimplementation of #1046.

bmah888:
The goal is to improve the responsiveness and/or reliability of
messages on the control connection, particularly in cases where the
test traffic fills up the path between client and server.

Originally suggested and implemented by @andrepuschmann, this version
adds some error-checking and a version for the client side as well as
the server.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -119,12 +119,12 @@ iperf_accept(struct iperf_test *test)
     if (test->ctrl_sck == -1) {
         /* Server free, accept new client */
         test->ctrl_sck = s;
-	// set TCP_NODELAY for lower latency on control messages
-	int flag = 1;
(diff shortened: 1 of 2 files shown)


src/iperf_server_api.c
```diff
@@ -119,6 +119,13 @@ iperf_accept(struct iperf_test *test)
     if (test->ctrl_sck == -1) {
         /* Server free, accept new client */
         test->ctrl_sck = s;
+	// set TCP_NODELAY for lower latency on control messages
```
andrepuschmann, line 122: indent is wrong

andrepuschmann (commented):
Minor thing, but the copy+paste seems to not match the file indent.

src/iperf_server_api.c
bmah888: Thanks for catching that, spaces vs. tabs problem. (I'm more of a tabs guy but it's more important to match the prevailing indent style, which I didn't do.) Will fix.

Issue #1065, opened 2020-11-04
Allow to configure the acceptable time drift between client/server for authentication


ralcini:
At the moment, the iperf3 authentication procedure is based on the server side validation of an encrypted token containing a time reference included by the client. If the time reference is not aligned to the server time, the token is considered invalid in order to avoid reply attacks.

The time drift allowed is embedded in the code here: https://github.com/esnet/iperf/blob/2a1309faf80f07bda1ba020ce8cfa09529561e57/src/iperf_auth.c#L70

In our scenario, we are noting that client devices are not always perfectly in sync with a NTP server and can be useful to make configurable (with an additional command line option) the allowed drift time with a default value (the current one).

PR #1070, opened 2020-11-09
Configurable value for time drift between client/server for authentication request issue1065

ralcini:
* feature: Allow to configure a custom value for time drift between client/server for authentication requested in #1065 

The use case is to support scenarios where it's not possible to enforce sync between client and server times.

### DIFF
--- a/docs/invoking.rst
+++ b/docs/invoking.rst
@@ -209,6 +209,10 @@ the executable.
                  The  file  is  a  comma separated list of usernames and password
                  hashes; more information on the structure of  the  file  can  be
                  found in the EXAMPLES section.
+
+          --time-drift-threshold seconds
+                 time drift threshold (in seconds) between the server and client
+                 during the authentication process.
    
    CLIENT SPECIFIC OPTIONS
           -c, --client host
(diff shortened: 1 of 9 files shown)


bmah888 (changes requested):
Thanks for the PR!  I did some quick functional testing, looks good.

The one change I'd like to see is to use the term "clock skew" instead of "clock drift". Clock drift is a difference in the rate between two clocks. Clock skew is a difference in the absolute value between two clocks, and that is the check that the authorization code does here. I think a global substitution of "drift" -> "skew" would take care of this. Would it be possible to do that? Quite happy to merge after that change. (Also it's great that you did all the documentation updates in this PR as well, thank you!)

ralcini:
Hi @bmah888 , thank you for your review. I've just committed the proposed change

bmah888:
Thanks @ralcini! Looks good so far, I'll do one more round of tests on this and then merge. Hopefully by the end of this week.

bmah888 (approved):
Looks great thanks again!

ralcini:
@bmah888 thank you for the quick merge approval!

Issue #1066, opened 2020-11-06
Add new option to allow multiple clients


hamzash9500:
Add a new server option to allow multiple client connections.

If the iPerf server is started with this option, the server will accept multiple parallel client connections. The clients can run tests in parallel.

PR #1074, opened 2020-11-17
Multiple Cleint support 

davidBar-On:
* Version of iperf3
latest master

* Issues fixed (if any):
#1066 

#### Brief description of code changes (suitable for use as a commit message):
Suggested change to support multiple clients.  The approach is:
- When the main server receive a new connection request and it is busy, it will allocate new port number, fork/exec new server and redirect the client to the new port .
- Ports are allocated beginning at the port number of the main server plus 1 (e.g 5201+1=5202) up to plus 15 (e.g. 5216).  Redirection is done by sending the offset from the main port (e.g 3 for 5204) in the control channel char message, were offset 2 is sent as `CONTROL_PORT_MIN+1`.  The 16 servers limit is `(CONTROL_PORT_MAX - CONTROL_PORT_MIN)`.
- Number of clients supported is specified by a new server option `--max-servers` (the number includes the main server).  The main server keep the list of servers` process IDs and when a request is received knows to check if they are still running and if a the client is allowed to run.
- The server executed for the client is waiting up to 30 secs (hard coded) for the client connection and terminates after one test (or after one connection request if there was a failure).  The 30 seconds is specified to the new server by `--connect-timeout` option (option is new for the server, but exist for the client for different purpose).


sraillard:
@bmah888 Do you plan to merge this PR?

ralvaradof:
@bmah888 Please merge this PR, this is a must in any multi user scenario or when you have a lot of testing clients with automatic tests.

sraillard:
There is maybe an issue in merging this patch as it's removing a previously added option: `--time-skew-threshold`

davidBar-On:
> There is maybe an issue in merging this patch as it's removing a previously added option: `--time-skew-threshold`

The option is not removed.  It is just that this PR was submitted before that option was added.  It is not a problem to merge the changes, but there is no point in doing that before there is a decision to merge this PR into the master branch, as over time there may be other changes to the master that will have to be merged.

bmah888:
I'll say I'm interested in merging this PR, although iperf3 was originally not intended to support multiple connections at once. The reason is that in the intended environment for iperf3, endhost performance can be (and in many cases is) the limiting factor, not the network. If an iperf3 server were to support connections from multiple clients simultaneously, this runs the risk of having multiple, unrelated connections interfere with each other. perfSONAR (for which iperf3 was originally implemented) actually uses a coordination mechanism between test nodes to attempt to ensure non-interference.

I haven't really given this PR the review it needs and deserves, but I do like that multiple client support needs to be explicitly enabled, and retains the current behavior by default.

damhau:
For anybody interested I've build a workaround with pyhton before this PR was started. https://github.com/damhau/iperf3_python_exporter

sraillard:
One important point, if I understand well this patch, the client should be also updated to correctly react to the change of port sent by the server. That means this will not allow existing clients to have multiple concurrent sessions with the updated server.

davidBar-On:
> That means this will not allow existing clients to have multiple concurrent sessions with the updated server.

This is correct.

davidBar-On:
> > That means this will not allow existing clients to have multiple concurrent sessions with the updated server.
> 
> This is correct.

Sorry, replied to quickly.  With this PR existing clients can work with the updated server.  However, if the server is set by `--max-servers` to support more then 1 client in parallel, then if a new client request is received while the server is busy it will activate a second server.  That will probably cause the client to fail because it will receive a response it doesn't understand (the request to reconnect to the new server - I didn't try what happens in this case).

sraillard:
I wonder if the best solution (but not the easiest) would be to check the cookie sent when the client is establishing the TCP connections (control or data) instead of redirecting the client to a fork of the server on a another port. This would also allow older clients to work with the newer server.

davidBar-On:
>...  check the cookie sent when the client is establishing the TCP connections (control or data) instead of redirecting the client to a fork of the server on a another port.

Not sure I fully understand the suggestion, but current iperf3 can support only one test from one client in parallel.  Note that each test may have different parameters, so with current design a server cannot support several parallel tests, and a major redesign is needed to change that.  It seems that the python script solution suggested by @damhau is also a good solution (note it also "forks" new servers).

In any case, I am not sure that this is a major issue.  If the multi-servers option is not used then old clients can work with the new server.  The issue is only when the `--max-servers` is set to support more than one client.  I assume that it is reasonable to expect that if the server is set to support multi-client then the clients in this environment will also be upgraded.

sraillard:
Yes, I understand that checking the cookie (the 37 first bytes sent on each TCP session initiated by the client) will only work with heavy refactoring as in that case the server must be able to support multiple different test contexts at the same time, my opinion was more than this would be the cleanest option to allow for multiple clients.

In my case, I have found a solution using HA-Proxy as the load balancer with multiple iperf3 servers running, using round robin and stick table based on IP source of client. This seems to work fine for now.

zokl:
@davidBar-On Would this patch help if the client is reconnected with a different address? Typically this is a client connection on a network with outages, where it gets a new IP address after reconnection. Now what happens is that the measurements are dropped, the server is not accepting new requests. This behavior is ubiquitous in mobile networks.

davidBar-On:
@zokl the approach of this patch can help to allow the client to reconnect while the old connection is still active on the server.  However, this will be a new connection and the measurements from the previous connection will not be accumulated.

Which version of iperf3 are you using? Version 3.10 introduced a new option `--rcv-timeout` to the receiving side (depends on whether `-R` is used) that specifies the time to wait for data before disconnecting.  The default is 2 minutes, so the client should be able to reconnect after 2 minutes.

How long does it take for the client to try to reconnect?  Alternative approach to using this PR's multi client approach is to set the server's `--rcv-timeout` to time that is short enough, so when the client tries to reconnect the server will already be available. (I assume it takes some time since the time the client is disconnected from the mobile network until it reconnects and gets a new IP address.)

zokl:
> @zokl the approach of this patch can help to allow the client to reconnect while the old connection is still active on the server. However, this will be a new connection and the measurements from the previous connection will not be accumulated.
> 
> Which version of iperf3 are you using? Version 3.10 introduced a new option `--rcv-timeout` to the receiving side (depends on whether `-R` is used) that specifies the time to wait for data before disconnecting. The default is 2 minutes, so the client should be able to reconnect after 2 minutes.
> 
> How long does it take for the client to try to reconnect? Alternative approach to using this PR's multi client approach is to set the server's `--rcv-timeout` to time that is short enough, so when the client tries to reconnect the server will already be available. (I assume it takes some time since the time the client is disconnected from the mobile network until it reconnects and gets a new IP address.)

Thank you, I will try it.

typzzh:
what is the status of this PR now ?

DorChen:
> 

@sraillard when use HA-Proxy solution, how does udp test works? HA-Proxy not support UDP proxy.

sraillard:
@DorChen : we are using only TCP, so HA Proxy is fine. The only limitation we have found so far is that all IP sources addresses of the multiple TCP connections established (for one test) should be the same.

davidBar-On:
@DorChen, it seems that other proxy solutions should be used to support UDP.  E.g. see this [StackOverflow discussion](https://stackoverflow.com/questions/31255780/udp-traffic-with-iperf-for-haproxy).

@sraillard, yust for information.  Are you using stickiness to make sure that the control stream and the data stream(s) for specific test from the same client will go to the same server?  E.g. I saw the following definition that uses the cookie that is sent at the beginning of the stream connection for the stickiness.  Are you using something like that?
```
backend ....
    ....
    stick-table type binary len 36 size 10k expire 1m
    tcp-request content track-sc0 req.payload(0,36)
    stick on req.payload(0,36)
    ...
```

sraillard:
Yes, we are using  **stick-table**, here is part of the configuration:
```
backend back-iperf3
        balance roundrobin
        stick-table type ip size 1m expire 5s
        stick on src
        server iperf01 127.0.0.1:10001
        server iperf02 127.0.0.1:10002
        ...
```
We are not using the same configuration as you wrote.

DorChen:
> @DorChen, it seems that other proxy solutions should be used to support UDP. E.g. see this [StackOverflow discussion](https://stackoverflow.com/questions/31255780/udp-traffic-with-iperf-for-haproxy).
> 
> @sraillard, yust for information. Are you using stickiness to make sure that the control stream and the data stream(s) for specific test from the same client will go to the same server? E.g. I saw the following definition that uses the cookie that is sent at the beginning of the stream connection for the stickiness. Are you using something like that?
> 
> ```
> backend ....
>     ....
>     stick-table type binary len 36 size 10k expire 1m
>     tcp-request content track-sc0 req.payload(0,36)
>     stick on req.payload(0,36)
>     ...
> ```

@davidBar-On  Thanks, now I use ipvs with rr and persistent option, support tcp and udp, it works.

ipvsadm -A -t xxx.xxx.xxx.xxx:8888 -s rr -p
ipvsadm -a -t xxx.xxx.xxx.xxx:8888 -r 172.17.0.3:8888 -m
ipvsadm -a -t xxx.xxx.xxx.xxx:8888 -r 172.17.0.4:8888 -m
ipvsadm -A -u xxx.xxx.xxx.xxx:8888 -s rr -p
ipvsadm -a -u xxx.xxx.xxx.xxx:8888 -r 172.17.0.3:8888 -m
ipvsadm -a -uxxx.xxx.xxx.xxx:8888 -r 172.17.0.4:8888 -m

sraillard:
> Are you using stickiness to make sure that the control stream and the data stream(s) for specific test from the same client will go to the same server?

Yes, exactly! As the client will open one control stream and one or many data stream, we have to be sure that each connection from the same client (so in my case, for the same source IP address) is going to the same server. Maybe using the cookie should be better as you can do multiple tests from one IP source address.

Regarding ipvsadm, this is interessing, but did you check exactly how the round robin algorithm is working as all connection from one client must go to only one server?

Issue #1076, opened 2020-11-23
iperf server should start measuring CPU utilization after the client connects not earlier


jtluka:
# Context

* Version of iperf3:

I used a build from https://github.com/olichtne/iperf/tree/fix_burst_mode, but I believe this is bug is present in latest iperf code as well. This branch is based on commit c47e43d5a3e222318e5c7fd17178c57e5b10c9a3.

* Hardware:
Any

* Operating system (and distribution, if any):
Linux

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

# Bug Report

The iperf server reports CPU utilization since the launching of the iperf server process and not since the actuall test start.

* Expected Behavior

The iperf server should report CPU utilization after the iperf client connects and starts the test.

* Actual Behavior

The iperf server CPU utilization depends on how long the server runs before the client connects and starts the test.

* Steps to Reproduce

Simply compare the CPU utilization in json output of server for following test scenarios:
1. start iperf server with json output format, wait 2 seconds, run iperf client that connects to the server
```
                "cpu_utilization_percent":      {
                        "host_total":   28.56335027483129,
                        "host_user":    6.18877428988225,
                        "host_system":  22.374568343317126,
                        "remote_total": 0,
                        "remote_user":  0,
                        "remote_system":        0
                },
```

2. start iperf server with json output format, wait 10 seconds, run iperf client that connects to the server
```
                "cpu_utilization_percent":      {
                        "host_total":   18.155508141151934,
                        "host_user":    4.0793554618947807,
                        "host_system":  14.076152679257154,
                        "remote_total": 0,
                        "remote_user":  0,
                        "remote_system":        0
                },
```

* Possible Solution

The initialization call to `cpu_util()` should be done after the connection from client is accepted.

olichtne:
Optionally, it would be interesting if cpu utilization was sampled at the same rate as the streams.

PR #1077, opened 2020-11-23
iperf_server_api: start calculating CPU utilization right before TEST…

jtluka:
…_START

Fixes issue 1076.

Signed-off-by: Jan Tluka <jtluka@redhat.com>

```
```

master

* Issues fixed (if any):

https://github.com/esnet/iperf/issues/1076

The cpu utilization counter initialization has been moved before sending TEST_START message instead of initialization right after iperf server process start. This fixes an issue if iperf server process is left running without client connected and the CPU utilization numbers are inaccurate.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -444,9 +444,6 @@ iperf_run_server(struct iperf_test *test)
         return -2;
     }
 
-    // Begin calculating CPU utilization
-    cpu_util(NULL);
-
     test->state = IPERF_START;
     send_streams_accepted = 0;
     rec_streams_accepted = 0;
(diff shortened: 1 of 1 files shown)


jtluka:
@bmah888 hi, could you provide some feedback on the patch please? we'd like to use this patch in our worfkflows until it gets merged into the iperf upstream code but it would be great if I get some feedback on the patch first.

Thanks!

bmah888:
> @bmah888 hi, could you provide some feedback on the patch please? we'd like to use this patch in our worfkflows until it gets merged into the iperf upstream code but it would be great if I get some feedback on the patch first.

@jtluka: Sorry, I'm completely slammed with other projects right now (iperf3 is actually work for me). A very quick glimpse over your patch gives me a warm fuzzy feeling (<-- that's good) but I haven't had time to actually sit down and think about this problem or do any testing with it yet. Thanks for the PR and sorry for the delay, I'm hoping to look at it "soon". Keep poking me on it.

bmah888 (approved):
Thanks for the PR, this looks good! I have done some testing along the lines of #1076 before and after this change, and I agree with your analysis and solution.

PR #1082, opened 2020-11-27
fix tcp option -k bug when socket write return EAGAIN

hyswtj:

```
```

* Issues fixed (if any):

If the socket write returns EAGAIN, the whole block is not completed, then the 
control variable blocks_sent should not be increased.
Due to network congestion, option -k cannot accurately control the number of packets to transmit,

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -183,6 +183,7 @@ struct iperf_stream
     int       green_light;
     int       buffer_fd;	/* data to send, file descriptor */
     char      *buffer;		/* data to send, mmapped */
+    int       pending_size;     /* pending data to send */
     int       diskfile_fd;	/* file to send, file descriptor */
     int	      diskfile_left;	/* remaining file data on disk */
 
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1629,7 +1629,8 @@ iperf_send(struct iperf_test *test, fd_set *write_setP)
 		}
 		streams_active = 1;
 		test->bytes_sent += r;
-		++test->blocks_sent;
+		if (!sp->pending_size)
+		    ++test->blocks_sent;
                 if (no_throttle_check)
 		    iperf_check_throttle(sp, &now);
 		if (multisend > 1 && test->settings->bytes != 0 && test->bytes_sent >= test->settings->bytes)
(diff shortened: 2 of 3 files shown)


bmah888 (commented):
Thanks for the PR! I'm thinking about this.

As I understand it, if we get EAGAIN when we're doing a TCP write, we might be in a situation where some of the data for an intended send (a block of the `--length` parameter) got sent but not all.  iperf3 in its current form would count that block as being sent for the purposes of option `-k`.

The patch you've sent would make iperf3 go back (a subsequent pass through the sending loop) and actually send the bytes necessary to complete a block before allowing it count for the purpose of `-k`. That gets the bytes sent, but at the cost of additional write operations.

Which behavior is "correct" depends I think on whether we're trying to count blocks of data that got sent or whether we're trying to count attempts to send data (regardless of whether they're complete or successful). I'm trying to figure out which makes more sense.

Ultimately this ambiguity comes from the semantics of TCP, which is a bytestream protocol that doesn't preserve message boundaries (thus allowing partial writes to happen).  With UDP, which is a message-oriented transport protocol, this problem doesn't exist...a write either happens completely or not at all.

I need to ponder this a little more.

(On an unrelated note: There's now a merge conflict between your PR and some other changes I have merged recently. If we try to merge your PR, one of us will need to deal with that.)

hyswtj:
Which behavior is "correct" depends I think on whether we're trying to count blocks of data that got sent or whether we're trying to count attempts to send data (regardless of whether they're complete or successful). 
----------In my opinion, option -k is count blocks of data that got sent, and count attempts to send data is meaningless, maybe one attempt result in zero bytes sent.Besides, option -k and option -l can count the total bytes sent，equivalent to option -n.Is it the program trying to do?

hyswtj:
-k, --blockcount #[KMG]   number of blocks (packets) to transmit (instead of -t or -n)

To transmit， not attempt to transmit, I think so.

hyswtj:
I have rebased my branch to the newest baseline.

bmah888 (approved):
Looks good. I'm mostly (but admittedly, not entirely) convinced that this change gives the correct behavior. It seems to be a bit more intuitive than the original behavior anyway. I'll go ahead and merge this and see what others think. Thanks for the PR!

Issue #1081, opened 2020-11-26
--get-server-output does not return result to the client for some cases 


prachichoudhary22:
* Version of iperf3: 3.1.3

* Hardware: Ubuntu 16.04

# Bug Report
When the client is run with '--get-server-output', '--json' option and '--json' option on server, the client gets an error "Error - unable to receive results" for server output in some cases when bits per second on server side is 0.

* Expected Behavior
The client should have received the server output json after the completion of the test.

* Actual Behavior
The client gets an error "Error - unable to receive results" for server output.

* Steps to Reproduce
On the server, run "iperf3 -s --json"
On the client, run "iperf3 -c server_ip --json -u -w 4m -i 1 --bandwidth 1000000000 -t 60 --length 30000 --get-server-output"

davidBar-On:
This is probably a **duplicate of #1080** (with the same root cause that the requested window size is too high).

bmah888:
I agree this is probably a duplicate of a previously reported issue.

prachichoudhary22:
This error is mostly seen when the bits per second at the receiver side is 0. In that case, when "--json" flag is used, the client fails with error "Unable to receive outputs".

#1080 and #1081 both are related to this particular flag "--json", the difference being for #1080 , the output is not fetched when receiver shows 0 bits per second. In #1081, the issue is "--udp-counters-64bit" in combination with "--json" and "--get-server-output".

FYI: The behavior is observed even when -w 4m was not configured.

Issue #1080, opened 2020-11-26
--get-server-output option does not work when using option --udp-counters-64bit


prachichoudhary22:
* Version of iperf3: 3.1.3

* Hardware: Ubuntu 16.04

# Bug Report
When the client is run with '--get-server-output',  '--udp-counters-64bit', '--json' option and '--json' option on server, the client gets an error "Error - unable to receive results" for server output.

* Expected Behavior
The client should have received the server output json after the completion of the test.

* Actual Behavior
The client gets an error "Error - unable to receive results" for server output.

* Steps to Reproduce
On the server, run "iperf3 -s --json"
On the client, run "iperf3 -c server_ip --json -u -w 4m -i 1 --bandwidth 1000000000 -t 60 --length 30000 --get-server-output --udp-counters-64bit"

davidBar-On:
Running the command using **iperf3 version 3.9+** gives the error: `iperf3: error - socket buffer size not set correctly`.  Using the `-d` debug option adds `SNDBUF is 327680, expecting 4194304`.  I.e. the test fails because the requested window buffer is 4MB, but in my computer buffer size is limited to 320KB.

The original problem is probably the same, and it may be that iperf3 version 3.1.3 did not properly check or notify about the issue.

(There is one related problem in 3.9+ that after this failure new clients cannot establish UDP stream with the server.  I submitted PR #1084 to fix this.)

bmah888:
That socket buffer size is fairly large. Is the client system configured to support a buffer setting that large? What if you try using a smaller value (or don't specify the parameter at all)?

prachichoudhary22:
Even if I do not specify that parameter, I see that when I use option "--json", we do not receive the output. However, I do get the output if I don't use "--json" flag.
Maybe because of the logs which are there because of the flag "--udp-counters-64bit"?

davidBar-On:
I tried the test using iperf3 **3.9+** in Ubuntu under Windows and it worked o.k.

PR #1084, opened 2020-11-28
Closing server prot-listener socket after stream setup failure

davidBar-On:
3.9+

* Issues fixed (if any):
#1080, #1081

Making sure the server closes `prot-listener` socket at the end of a session with a client.  Without this change the socket may remain open, for example when there is a failure during the UDP stream establishment because the `--window` value is too high.
Closing the socket was added to `cleanup_server()` as it seem to handle all cases where the socket may remain open.


bmah888 (approved):
Good fix, thanks for the PR!

PR #1090, opened 2020-12-02
Bitrate throttling when burst is specified

davidBar-On:
3.9+
* Issues fixed (if any):
None

When the `-b` option specifies a `burst` value, throttling the bitrate does not work.  This is because [iperf_check_throttle()](https://github.com/esnet/iperf/blob/d3049a60fec5332733a3a8b61a1d46f00878b30c/src/iperf_api.c#L1542) does not perform the check when burst value was defined.

This PR is a proposed solution - remove the dependency of `iperf_check_throttle()` on the burst value and make sure that calling the function depend on the burst value.  (Except for the call by [send_timer_proc()](https://github.com/esnet/iperf/blob/d3049a60fec5332733a3a8b61a1d46f00878b30c/src/iperf_api.c#L1706) which does not seem to be related to the change.)


bmah888:
I confirm the bug where specifying a burst size makes throttling the bitrate not work.

bmah888 (approved):
Thanks for the PR...this was kind of subtle!

Issue #221, opened 2014-11-20
A test request from an older client can permanently freeze the server


annualp:
I have found that an older client sending a test request will make the iperf3 server (version 3.0.9) completely unresponsive.  The test request from the client does not work and the server will no longer answer any more test requests.  When I do an strace and attach to the server when this problem is happening I see that it is stuck in a read attempt.

strace -p 792
Process 792 attached - interrupt to quit
read(4,

I have tested this with the server running on two different machines.

first machine:
Linux dev-ftp 3.2.0-40-generic-pae #64-Ubuntu SMP Mon Mar 25 21:44:41 UTC 2013 i686 i686 i386 GNU/Linux

second machine:
Linux G06079 2.6.32.13-0 #1 SMP Fri May 14 10:01:26 CDT 2010 i686 i686 i386 GNU/Linux

The older client I am using is a Windows port.  I am aware that Windows is not a supported platform, but if you Google "iperf3 windows" the result on sourceforge.net (http://sourceforge.net/projects/iperf3cygwingui/) is one of the top results.  It is quite likely someone would try and use this.  I downloaded iperf3cygwingui-trunk318-03-src.zip and this is the version information it reports when I run iperf3 from the Windows command line:

iperf version 3.0b4-trunk318-03-20120720 (2 Aug 2010)

Note: this problem seems similar to the (closed) problem #202 (Connecting to server process during an existing test freezes the server).  However, with that problem the server's functionality was restored after the connection attempt failed.  With this problem the server remains unresponsive.

rpodgorny:
any progress on this?

bmah888:
Not really.  On a marginally related note, we do know that there was "something" that happened between the iperf3 betas and the GA release that broke compatibility.

bmah888:
This is still an issue...trivially easy to do by firing up an iperf3 server and then connecting to the server's port (i.e. with telnet or nc) and then not doing anything.  One thing that's a little complicated is that the client can stop talking to the server at any point and cause the server to stall out.

loxK:
I have this issue too. I have scheduled bandwidth monitoring between two nodes. When the connection fails during a bandwidth measurement the iperf3 server node just hangs forever.

davidBar-On:
`iperf_udp_connect()` sets timeout for reading from the socket:
```c
    #ifdef SO_RCVTIMEO
        /* 30 sec timeout for a case when there is a network problem. */
        tv.tv_sec = 30;
        tv.tv_usec = 0;
[...]

PR #1101, opened 2021-01-10
Server select timeout to prevent server to get stack because of client or network errors

davidBar-On:
3.9+

* Issues fixed (if any):
#221

A proposed fix to prevent the the Server to be stack forever in case the client is stack of there is a network problem. 

The main change is adding a timeout to the server's `select()`, so the fix should work as long as the `select()` timeout mechanism works properly.  Two different timeouts are used:
1. When the server is idle waiting for connect request from a client, the timeout is specified by a new `--idle-timeout` parameter.  The default is no timeout, to keep backward compatibility.
2. After receiving connect request from a client until the test ends the timeout is 2 minutes.  I didn't add a parameter to set this timeout assuming that there is no scenario where the server will not receive anything from the client for more than 2 minutes.

If the `select` times out because of either of these timeouts, it will terminate the test and wait for a new test request.

A change I kept from the debugging phase is to add test number to the server message when waiting to new test request.  E.g. "`Server listening on 5201 (test #2)`"

 I don't have the full testing capabilities for such problems, but the fix works properly for all the scenarios I was able to test.


src/iperf_locale.c
```diff
@@ -133,6 +133,8 @@ const char usage_longstr[] = "Usage: iperf3 [-s|-c host] [options]\n"
                            "                            credentials\n"
                            "  --time-skew-threshold    time skew threshold (in seconds) between the server\n"
                            "                            and client during the authentication process\n"
+                           "  --idle-timeout #         restart server if is idle for #[sec] to overcome\n"
```
bmah888, line 136: This addition should go outside the HAVE_SSL conditional.

bmah888 (commented):
Thanks for the PR! I generally like this code change, but I'm having some problems testing it (in particular, triggering the conditions that cause the . Could you tell me how you tested this to be sure it does what you want?

bmah888:
Thanks, I appreciate your low-risk approach. I'm just trying to figure out how to get the changes to actually make a difference.  Here are a couple of things I tried, against a server invoked as `iperf3 --server --idle-timeout 5`, running on the same host (CentOS 7 VM):

% iperf3 --client localhost
(this was good...gave me a normal output...the only way I could tell the difference on the server was the adjusted output)

% iperf3 --client localhost
(then suspend the client process after a few seconds...I expected the server to close the connection after the timeout elapsed)

% nc localhost 5021
(didn't do anything...somehow I expected the server to close the connection after the 5 seconds I specified on the command-line)

So I'm trying to figure out basically how I can trigger the cases where the server will (with this change) behave better.

davidBar-On:
Oops! My previous reply was incorrect.  I confused myself between the idle and active cases ....  Sorry for that.

The `--idle-timeout` is for the **server idle mode**, i.e. when it is waiting for new connection.  E.g., with `iperf3 --server --idle-timeout 5`, **when the server is idle** the server will "restart" every 5 seconds ("restart" means closing connections the same as a test ended - not restart the process).  Testing this is done by just not running a client and running the server with  `-d` so it will print the "_Server listening on ..._" message every time it "restarts".  The `-d` is needed since I thought that normally the serer "restart" in idle mode should be silent.

The temporary `sleep` was added only for the **server active mode** - when a client is connected.  In this case the timeout is 2 minutes - defined by `NO_MSG_RCVD_TIMEOUT`, so the client sleep should be more then 2 minutes.  Actually for testing I changed `NO_MSG_RCVD_TIMEOUT` to few seconds for faster testing and to make sure that other timeouts (e.g. TCP timeout) will not be triggered before this timeout.

Therefore, in the following cases:
>% iperf3 --client localhost
(then suspend the client process after a few seconds...I expected the server to close the connection after the timeout elapsed)
% nc localhost 5021
(didn't do anything...somehow I expected the server to close the connection after the 5 seconds I specified on the command-line)

the connection should timeout after 2 minute.

bmah888:
Thanks! I owe you some testing / review on the latest version of your PR.

bmah888:
I understand now why I wasn't able to test in a meaningful way before. The problem was that there is already a server-side check that is used for timed tests (the default or with `--time`) that effectively limits the test on the server to the configured test time (default 10 seconds) plus 40 seconds. I was testing with the default 10-second tests, so that timer fired at 50 seconds into the test, much shorter than your 2-minute idle timer.

So if I do `iperf3 --client localhost --bytes 100G` and suspend the test in the middle, then wait around for 2 minutes, I get the 2-minute idle timeout message as expected.

Still fiddling around with this change but I now understand what it's doing, thanks.

bmah888 (approved):
Apologies for taking so long to review this (and other) PRs you have submitted. Thanks for all your help!

This looks pretty good. I might do a few language nitpicks here and there after merging but that will just take me a few minutes I think. Also there is a slight problem when running the server with `--json` (it doesn't print any JSON output if the active test timeout happens)...but that's mostly due to a bug elsewhere in iperf3 that I will fix (and then tweak your code to track that fix).

PR #1110, opened 2021-01-28
Enable writing to pidfile in client mode

wjowsa:
Writing to a pidfile in client mode might be useful in some [cases](https://github.com/openwrt/luci/pull/4651).
It seems to be a low-hanging fruit so perhaps it is worth to enable such a functionality.


bmah888 (commented):
Thanks for the PR! The code functionality looks good.

We need to do a couple of minor doc fixes to relocate the documentation for `-I` / `--pidfile` from the "server-only" options to the "server or client" options. The two places are in `src/iperf_locale.c` (the definition of the string `usage_longstr()`) and in `src/iperf3.1` (the manual page for the iperf3 command).

If you can do that on your branch, that'd be wonderful. I'd be happy to merge after that.

(If I get really impatient and want to finish off this PR, I might merge anyways and fix up the documentation myself afterwards. But it's probably better if we do this all at once.)

bmah888 (approved):
Looks great, thanks for the doc changes! Hoping to merge this later today, tomorrow at the latest.

wjowsa:
Thanks for approving this change so fast!

PR #1113, opened 2021-02-02
Fix iperf_send() termination test in bytes/blocks mode

hanvari:
In iperf_send() function, the check for termination test
in bytes/blocks mode is at the end of the iteration loop
for multisend (outer loop) and streams (inner loop).
If for any iteration of multisend (outer) loop bytes/blocks
sent reaches the desired limit, it still continues to send
data until the loop is exhausted. (The `break;` command does not
help, since it is already inside the streams (inner) loop).

This is a simple fix which brings the condition check to the
beginning of the inner loop, so it will skip the iteration if
the bytes/blocks count is already reached the target; hence
avoiding to send more data to the network.

```
```

* Issues fixed (if any): N/A

```
Fix iperf_send() termination test in bytes/blocks mode

In iperf_send() function, the check for termination test
in bytes/blocks mode is at the end of the iteration loop
for multisend (outer loop) and streams (inner loop).
If for any iteration of multisend (outer) loop bytes/blocks
sent reaches the desired limit, it still continues to send
data until the loop is exhausted. (The `break;` command does not
help, since it is already inside the streams (inner) loop).

This is a simple fix which brings the condition check to the
beginning of the inner loop, so it will skip the iteration if
the bytes/blocks count is already reached the target; hence
avoiding to send more data to the network.
```

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1662,6 +1662,10 @@ iperf_send(struct iperf_test *test, fd_set *write_setP)
 	SLIST_FOREACH(sp, &test->streams, streams) {
 	    if ((sp->green_light && sp->sender &&
 		 (write_setP == NULL || FD_ISSET(sp->socket, write_setP)))) {
+        if (multisend > 1 && test->settings->bytes != 0 && test->bytes_sent >= test->settings->bytes)
+            break;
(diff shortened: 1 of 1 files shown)


bmah888 (approved):
Looks good, thanks for sending this PR. I agree with your logic analysis, and in testing, the change fixes some oddities I have seen in particular with multiple parallel connections. Let's get this merged.

PR #1115, opened 2021-02-02
diskfile_send() sent data capped at file-size

hanvari:
**Attention** Depends on pull request `Fix/Optimize test termination condition check #1114`

**Issue:** `diskfile_send()` unconditional call to `sp->snd2`
would result in sending full buffer size everytime,
regardless of the file size.

**Fix:** The function updated to check for end-of-file (reading 0 bytes)
and,
  1. set `sp->pending_size` to appropriate data length available to be sent 
  2. check for end condition and avoid sending data more than file size.

**Note:** The fix is only for the maximum cap on the data size sent on the network.
If other parameters (-t, -n, etc.) yield smaller size or shorter time then needed,
the file will still be partially sent to the network.

```
```

`master` + Depends on pull request `Fix/Optimize test termination condition check #1114`

* Issues fixed (if any): N/A

```
**Issue:** `diskfile_send()` unconditional call to `sp->snd2`
would result in sending full buffer size everytime,
regardless of the file size.

**Fix:** The function updated to check for end-of-file (reading 0 bytes)
and,
  1. set `sp->pending_size` to appropriate data length available to be sent 
  2. check for end condition and avoid sending data more than file size.

**Note:** The fix is only for the maximum cap on the data size sent on the network.
If other parameters (-t, -n, etc.) yield smaller size or shorter time then needed,
the file will still be partially sent to the network.
```

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4088,24 +4088,46 @@ static int
 diskfile_send(struct iperf_stream *sp)
 {
     int r;
+    int buffer_left=0; // represents total data in buffer to be sent out 
     static int rtot;
 
     /* if needed, read enough data from the disk to fill up the buffer */
     if (sp->diskfile_left < sp->test->settings->blksize && !sp->test->done) {
-	r = read(sp->diskfile_fd, sp->buffer, sp->test->settings->blksize -
-		 sp->diskfile_left);
-	rtot += r;
-	if (sp->test->debug) {
-	    printf("read %d bytes from file, %d total\n", r, rtot);
-	    if (r != sp->test->settings->blksize - sp->diskfile_left)
-		printf("possible eof\n");
-	}
-	/* If there's no data left in the file or in the buffer, we're done */
-	if (r == 0 && sp->diskfile_left == 0) {
-	    sp->test->done = 1;
-	    if (sp->test->debug)
-		printf("done\n");
-	}
+    	r = read(sp->diskfile_fd, sp->buffer, sp->test->settings->blksize -
+    		 sp->diskfile_left);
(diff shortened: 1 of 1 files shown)


src/iperf_api.c
```diff
@@ -4088,24 +4088,46 @@ static int
 diskfile_send(struct iperf_stream *sp)
 {
     int r;
+    int buffer_left=0; // represents total data in buffer to be sent out
```
bmah888, line 4091: Thanks for the PR! I am wondering how this code would handle the situation where upon entry, the buffer has `blksize` bytes? In this case, `buffer_left` will be initialized to zero and, because we don't need to read any data, it'll stay zero until we get to the check in line 4123, at which point we declare we're done.

Would it make more sense to initialize `buffer_left` to `sp->diskfile_left` instead?

bmah888 (commented):
I put a comment on the initializer in line 4091. Thanks!

src/iperf_api.c
hanvari: You are right. To keep this function self-contained that corner case should be handled here.
It will imply updating line 4091 to initialize `buffer_left` to `sp->diskfile_left` as you suggested. Consequently the line 4098 will need to be updated to remove adding `sp->diskfile_left` again to the `buffer_left`. 

Lastly, for better clarity and code readability, it probably makes sense to replace conditions in lines 4105 and 4115 to use the alternative conditions already mentioned in the comments.

bmah888 (approved):
I'm going to commit this as-is, and then follow-up with another commit with the changes we discussed in the comments.

PR #1116, opened 2021-02-02
Auto adjustment of test-end condition for file-transfer

hanvari:
In file transfer mode (-F), if no test-end condition is set,
(bytes, blocks, duration), it will automatically adjust it to
file size (in bytes).
If call to stat() fails, it will skip setting the bytes count, hence defaulting to duration mode (10 seconds).

* Issues fixed (if any): N/A

**Auto adjustment of test-end condition for file-transfer**

In file transfer mode (-F), if no test-end condition is set,
(bytes, blocks, duration), it will automatically adjust it to
file size (in bytes).
If call to stat() fails, it will skip setting the bytes count, hence defaulting to duration mode (10 seconds).

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1512,11 +1512,13 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
         test->role == 'c'
         ){
         struct stat st;
-        stat(test->diskfile_name, &st);
-        iperf_size_t file_bytes = st.st_size;
-        test->settings->bytes = file_bytes;        
-        if (test->debug)
-            printf("End condition set to file-size: %d bytes\n", test->settings->bytes);
+        if( stat(test->diskfile_name, &st) == 0 ){
+            iperf_size_t file_bytes = st.st_size;
+            test->settings->bytes = file_bytes;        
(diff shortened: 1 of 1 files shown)


hanvari:
@bmah888 Hi Bruce, this is a minor adjustment for file transfer mode, and only depends on a POSIX `stat()` library function.
Is there a reason it is not merged yet?

bmah888:
> @bmah888 Hi Bruce, this is a minor adjustment for file transfer mode, and only depends on a POSIX `stat()` library function.
> Is there a reason it is not merged yet?

I can't think of a reason in particular...most recently I've been on vacation but that doesn't excuse me sitting on this PR for the time before that, apologies! I'll try to look at this as I'm getting caught up.

bmah888 (commented):
What happens if there's an error in the stat(2) system call?

hanvari:
> What happens if there's an error in the stat(2) system call?

PR updated to appropriately handle stat() error.
If error happens, it will skip setting the bytes count, hence defaulting to duration mode (10 seconds).

bmah888 (approved):
Thanks, looks pretty good!

Issue #1107, opened 2021-01-25
Reduce default UDP Package size to physically MTU - use WSA Registered IO on windows to receive UDP packages


Febbe:
I found out, what's the reason, that the UDP RSS Queues aren't working: The Intel(R) Ethernet Converged Network Adapter X550-T2 does not send Jumbo Packages even if configured (other NIC does), so all UDP packages larger than the default 1500MTU are split into IP packages. IP packages do not have ports and the UDP-RSS-Hashing algorithm can't work. Even if it would know that the IP package is part of an UDP package.

The only way is to reduce the MTU size on the sending side of iperf to the physically MTU.
Also windows does not receive more than one UDP package from the kernel, when using the default WINSOCK API. Which will result in very high CPU load. To solve this iperf should use the WSA Registered I/O API.

_Originally posted by @Febbe in https://github.com/esnet/iperf/issues/1049#issuecomment-766829448_

bmah888:
I'd glad you were able to find out what's causing this problem! Unfortunately ESnet doesn't support iperf3 on Windows, so if there's a solution it's going to have to come from the community (as well as not break iperf3 for supported platforms).

Febbe:
Does "ESnet doesn't support iperf3 on Windows" mean, that you as ESnet won't fix that, but you'll accept PR's from the community?

As I know RSS-Queues are also utilized on Linux platforms, so even there it would make sense to, at least, add a flag to forbid package fragmentation for udp, instead of using the default -l 8K flag.

bmah888:
> Does "ESnet doesn't support iperf3 on Windows" mean, that you as ESnet won't fix that, but you'll accept PR's from the community?

We'll take PRs from the community if they don't break functionality (use conditional compilation if possible, there are several examples in-tree) or cause major changes to the design of the iperf3 design. We don't have the ability to actually test PRs specific to Windows in our environment, so the best chance for getting a PR accepted is if it's small and easily-understood by non-Windows developers.

davidBar-On:
Submitted PR #1117 to add option for setting the don't fragment flag.

Issue #638, opened 2017-09-13
Setting TOS/DSCP when using Reverse Mode (--tos with -R)


brandon-shannon:
# Context

* Version of iperf3: 3.0.11

* Hardware: Macbook Pro

# Bug Report

* ACTUAL Behavior
When doing a Reverse test (-R), the --tos value isn't used to tag traffic from the server to the client, making the TOS setting irrelevant.  Only traffic from the client to the server is tagged, but on a reverse test that's not the important bit.

* EXPECTED Behavior
Packets sent from server to client would be tagged correctly.

* Steps to Reproduce
1. iperf3 -c 104.131.128.139 -p 2222 --tos 72 -R
2. Look at the packets in WireShark

brandon-shannon:
For what it's worth, I can see that the client tells the server the ToS/DSCP value to use.  It looks like the server just doesn't do it.  Here's a capture using UDP via WireShark.

`{"udp":true,"omit":0,"time":10,"parallel":1,"reverse":true,"len":1448,"bandwidth":1048576,"pacing_timer":1000,"TOS":184,"client_version":"3.2"}`

gwbrown:
This is an odd bug.  If iperf is started in its default dual IPv6/IPv4 mode, the server will fail to set the appropriate DSCP value on outgoing packets in IPv4 streams.  **A workaround is to set the `-4` on the server if you are running IPv4 tests.**

In detail, the socket on the server is opened in IPv6 mode in order to be able to accept both IPv6 and IPv4 connections per [this comment in iperf_tcp.c](https://github.com/esnet/iperf/blob/4a45b32b4770f22e37dde22cce16dbc76a0ea9b6/src/iperf_tcp.c#L169).  When the server goes to set the DSCP option on the socket for a specific test [here in iperf_api.c](https://github.com/esnet/iperf/blob/4a45b32b4770f22e37dde22cce16dbc76a0ea9b6/src/iperf_api.c#L3280), it detects that the socket is an IPv6 socket and sets the appropriate IPv6 option **even when a test is using IPv4**.  I've done some testing, and if iperf_api.c is modified to attempt to set **both** the IPv6 and IPv4 DSCP options when the socket is detected as IPv6, the behavior will be correct (although if the connection is really an IPv6 connection, the call to set the appropriate IPv4 option will fail).

I'm going to do some more testing and intend to submit a pull request with this fix.

bmah888:
[...]

PR #1119, opened 2021-02-03
IP dont fragment support

davidBar-On:
3.9+

* Issues fixed (if any):
#1107

Adding support for setting IP dont-fragment flag, by adding `--dont-fragment` option.  (There is an option of adding value to this option to allow setting do-fragment, but decided not to do that since it may not be required and it complicates the compatibility testing for the different environments.)

The option works well for **Windows** using CigWin.  I don't have test environment for **Unix** (BSD, etc.).

For **Linux**, for some reason my system always set the Don't Fragment flag, so I was not able to test whether setting the option works. However, I tested that that the opposite works - asking to not set the flag (`IP_PMTUDISC_DONT`).  Therefore I believe that asking to set the flag (`IP_PMTUDISC_DO`), which is what is in the code, also works o.k.


bmah888:
Thanks for the PR! I've been looking at this, as you can probably tell from me flailing around with the conflict resolution tool in the Github Web UI.

I have some medium-sized changes to suggest on top of this PR, mostly having to do with the fact that software portability is hard and dealing with the IP DF socket options isn't particularly portable. So I ended up refactoring conditional compilation and doing some autoconf goop that conditionalizes argument parsing and help text, among other things.

Still being a relative Github newbie, what's the right way for us to coordinate this? I can do comments on your PR but that's going to be a little clumsy due to the scale of the changes. It feels kind of wrong to just close your PR and start over (although I admit I've done that in the past). Any suggestions / hints?

bmah888:
Also just noting here that the way to turn off the Linux default DF behavior is:
```
sysctl net.ipv4.ip_no_pmtu_disc=1
```
This then can be overridden on a per-socket basis via the socket option.

davidBar-On:
> I ended up refactoring conditional compilation and doing some autoconf goop that conditionalizes argument parsing and help text, among other things.

From searching the Internet I was hopping that Linux is the only exception .... Make sure to keep the non-Linux case at least for Windows.

>Still being a relative Github newbie, what's the right way for us to coordinate this? I can do comments on your PR but that's going to be a little clumsy due to the scale of the changes. It feels kind of wrong to just close your PR and start over (although I admit I've done that in the past). Any suggestions / hints?

I am also Github newbie.  From what I read [here](https://stackoverflow.com/questions/39061191/is-it-possible-to-change-other-peoples-code-in-pull-requests-github) you should be able to pull the branch from by fork repository: "**https://github.com/davidBar-On/iperf/tree/issue-1107-ip-dont-fragment-support**" (I set the permission). Then you can make the changes and push them back to my repository.  That should be the same as if I did the changes.

I never tried that, so don't know if it works.  If not, If not, I have no problem that you will close the PR and open new one.

bmah888:
> > I ended up refactoring conditional compilation and doing some autoconf goop that conditionalizes argument parsing and help text, among other things.
> 
> From searching the Internet I was hopping that Linux is the only exception .... Make sure to keep the non-Linux case at least for Windows.

Right...the non-Linux case is good for *BSD, et al. so it will definitely stay.  The other case is macOS, which doesn't support either mechanism as far as I can tell (and that is one of our supported platforms).

> > Still being a relative Github newbie, what's the right way for us to coordinate this? I can do comments on your PR but that's going to be a little clumsy due to the scale of the changes. It feels kind of wrong to just close your PR and start over (although I admit I've done that in the past). Any suggestions / hints?
> 
> I am also Github newbie. From what I read [here](https://stackoverflow.com/questions/39061191/is-it-possible-to-change-other-peoples-code-in-pull-requests-github) you should be able to pull the branch from by fork repository: "**https://github.com/davidBar-On/iperf/tree/issue-1107-ip-dont-fragment-support**" (I set the permission). Then you can make the changes and push them back to my repository. That should be the same as if I did the changes.
> 
> I never tried that, so don't know if it works. If not, If not, I have no problem that you will close the PR and open new one.

I'll try that, thanks. This will be good to make sure that my code changes don't break on Windows/Cygwin (which is a platform I cannot test).

bmah888:
I've pushed something to your branch (thank you for permissions) that I think will work, need to do some more testing across the various supported platforms (Linux, macOS, and FreeBSD).

bmah888:
Finally got everything working on all the platforms I support. Hopefully Windows/Cygwin still works?

davidBar-On:
Windows/Cygwin is defining `IP_DONTFRAGMENT` and not `IP_DONTFRAG`.  Therefore the following changes are needed.  I tested these changes and they work o.k. (I am not pushing the changes as something got wrong with my local git branch after pulling your changes, so I don't want to cause issues to the PR).

1. **configure**: `#ifdef IP_DONTFRAGMENT` should be added, the same as for `IP_DONTFRAG` and `IP_MTU_DISCOVER`.

2. **sec/iperf_api.c**: the `#if defined(IP_DONTFRAG)` should be be duplicated, with using `IP_DONTFRAGMENT` for both the `#if defined` and the `setsockopt` option.

bmah888:
I think I fixed this for Windows now, please let me know if this works. Note that I added some more checking to be sure we only try this with UDP/IPv4 tests.

While testing, I discovered that this doesn't work right with `--reverse`. I think this might be related to #638, at least the situation is somewhat similar (trying to set IPv4-specific socket options on an IPv6 socket that has an IPv4 mapped address). I don't necessarily need to hold up this PR for that problem, since it seems to exist with other features and a general solution to the problem might make `--dont-fragment` work as well.

davidBar-On:
@bmah888, I have tested the changes and they work fine under Windows Cygwin.

PR #518, opened 2017-02-25
Fix header includes and build failures on musl

f1rebird:
- include stdint.h in files where its types are used
- getsockopt expects socklen_t instead of int as its fifth argument
- remove includes of headers already included in iperf.h
- sort header includes alphabetically
- fixes #331 and closes #344 

Signed-off-by: Moritz Kick <f1rebird@users.noreply.github.com>

PR #1122, opened 2021-02-12
Move iperf.h include before tcp.h in net.c for _GNU_SOURCE definition

davidBar-On:
3.9+

* Issues fixed (if any):

Recently `iperf.h` include was included in `net.c`.  That caused make in Alpine to fail with error "_field 'tcpInfo' has incomplete type_".  `iperf.h` include was moved before `netinet/tcp.h` to apply the related fix by #518 - defining `_GNU_SOURCE`.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -26,6 +26,9 @@
  */
 #include "iperf_config.h"
 
+// iperf.h include should preceed netinet/tcp.h for _GNU_SOURCE definition
+#include "iperf.h"
+
 #include <stdio.h>
 #include <unistd.h>
 #include <errno.h>
@@ -61,7 +64,6 @@
 #include <poll.h>
 #endif /* HAVE_POLL_H */
 
-#include "iperf.h"
 #include "iperf_util.h"
 #include "net.h"
 #include "timer.h"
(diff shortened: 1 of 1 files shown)


bmah888 (commented):
Thanks for the PR! There's a bit of a problem here in that `iperf.h` declares a `FILE`, which is defined in `stdio.h`. (This causes breakage on macOS, not sure about other platforms.)

So `iperf.h` actually includes `netinet/tcp.h` (and it does the right thing with `_GNU_SOURCE` as you noted). I wonder if it might work better to remove the explicit inclusion of `netinet/tcp.h` and instead just include `iperf.h` below the system-related includes...maybe just before `iperf_util.h`. That's a little more consistent with what we did in other parts of iperf3. It seems to work on my MacBook Pro, haven't tried other platforms yet. Would that work on Alpine?

davidBar-On:
>... remove the explicit inclusion of `netinet/tcp.h` and instead just include `iperf.h` below the system-related includes...maybe just before `iperf_util.h`

This is working fine with Alpine, so I made the change to the PR (although it may be easier to just do it manually).

bmah888:
Ah, silly me, I didn't realize the net effect of my comment applied to your original change was just to delete the inclusion of `netinet/tcp.h`. I'll get this change onto the mainline one way or another, thanks!

bmah888 (approved):

PR #1125, opened 2021-02-18
Rcv timeout option

davidBar-On:
Latest 3.9+

* Issues fixed (if any):

Enhancement to PR ##1101 - add `--rcv-timeout` option to allow setting the timeout for receiving packet from the sender in a running test, instead of the constant 120 seconds set in #1101.  This is to allow short timeout, so the server will not be "busy" for long time doing nothing.  Also, the resolution is set to ms (with minimum of 100ms), as at least in fast networks that may be required.

Since both the server and the client can be a sender, the option is allowed for both.  The server setting is for all tests - the client is not sending its `--rcv-timeout` setting to the server.

While changing the help message, few printf constants from other help lines were changed to parameters.


bmah888 (commented):
Thanks for the PR! I glanced over the patch and it looks good so far...have not tried testing it yet.

I bet adding this flag is probably more work than either you or I thought it would be. One of the annoying things about the iperf3 code-base is there are too many options and too many ways for features to interact. Nice for users, but painful for developers sometimes.

bmah888 (approved):
Looks pretty good, and it seems to work as you described. I'm going to merge this.

I was going to suggest some wording fix-ups and a manpage update, but maybe I'll just fix those myself after the merge.

Thanks!

Issue #1061, opened 2020-10-06
Ubuntu on Windows WSL does not have tcp congestion.


robert-andrzejuk:
# Context
Using Microsoft Store - Ubuntu, is installed (20.04LTS)
Installing iperf3 with "sudo apt-get install iperf3"

* Version of iperf3:
version 3.7-3

* Operating system (and distribution, if any):
Ubuntu on Windows (20.04 LTS)

# Bug Report
When running: `iperf3 -c XXXX` an error occurs:
iperf3: error - unable to set TCP_CONGESTION: Supplied congestion control algorithm not supported on this host

* Expected Behavior
iperf3 runs without error

* Actual Behavior

* Steps to Reproduce

* Possible Solution
A build on Ubuntu on Windows should have the `HAVE_TCP_CONGESTION` undef'ed.

bmah888:
I'm afraid I don't completely understand the bug report...unless `XXXX` is a valid congestion control algorithm on Ubuntu on Windows, it's giving exactly the error message it should.  I'm not familiar with that runtime environment, but you might see if there's an equivalent to `sysctl net.ipv4.tcp_available_congestion_control` to list out what congestion control algorithms you have to choose from on that system.

If you meant `XXXX` as in everything you tried didn't work, note that the only way that `HAVE_TCP_CONGESTION` gets defined is if there's a `TCP_CONGESTION` socket option defined in the system headers (it's auto-detected at build time).  If Ubuntu on Windows defines that without actually implementing the socket option, then that's a bug of that environment, not iperf3 (or whatever other application).

davidBar-On:
I am also using Windows WSL and I see that `/usr/include/netinet/tcp.h` includes:
```c
/*
 * User-settable options (used with setsockopt).
 */
......
#define TCP_CONGESTION       13 /* Congestion control algorithm.  */
.......
```

However, there is no `/proc/sys/net/ipv4/tcp_available_congestion_control` file.  Therefore, the test in `configure` for whether TCP Congestion is supported may need to be changed to support WSL.  E.g. change:
```sh
if (eval "$ac_cpp conftest.$ac_ext") 2>&5 |
  $EGREP "yes" >/dev/null 2>&1; then :
  iperf3_cv_header_tcp_congestion=yes
else
  iperf3_cv_header_tcp_congestion=no
fi
```

to something like:
```sh
[...]

PR #1126, opened 2021-03-10
Fix issue 1061 - not fail in WSL1 when cannot get defalt congestion a…

davidBar-On:
Latest 3.9+

* Issues fixed (if any):
#1061

Whn congestion algorithm is not set (`-C` option), do not fail when cannot get the algorithm used.  This issue happens at least with WSL1 that is a Linux but still doesn't support congestion algs.  Instead of failing, in this case the `congestion_used` is set to "**unknown**".


bmah888 (approved):
Thanks for the PR! I have no feedback at this time, going to merge it shortly.

PR #1128, opened 2021-03-13
Fix to active TCP test termination by new connection

davidBar-On:
latest 3.9+

* Issues fixed (if any):
None

Fix two issues that caused an active TCP test to terminate if a new connection request was received while in streams creation phase.  One issue was in `iperf_tcp_accept()` - after identifying that the cookies of the new connection if from a new client, error was returned which caused the active test to terminate.  The other issue was in `iperf_run_server()` where congestion alg was set for the new client, although the stream to it was already closed by `iperf_tcp_accept()`.  That also cause the active test to terminate.

Another minor issue that was fixed is that after a client received a failure state (negative state) from the server, `iperf_client_end()` still tried to send back `IPERF_DONE` to the server.  That caused the client to issue failure message of "_unable to send control message: Connection reset by peer_" instead of "_the server is busy running a test_".

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -440,8 +440,11 @@ iperf_client_end(struct iperf_test *test)
     /* show final summary */
     test->reporter_callback(test);
 
-    if (iperf_set_send_state(test, IPERF_DONE) != 0)
-        return -1;
+    // Send response only if no error in server
+    if (test->state > 0) {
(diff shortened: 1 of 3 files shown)


bmah888:
Thanks for the pull request! This fixes some confusing errors that don't show up in our normal use cases, so I'm glad you found them. I'm going to implement these changes slightly differently to deal with the merge conflicts and reduce whitespace diffs.

Issue #1129, opened 2021-03-14
Wrong errno used for reporting connection failures


Safari77:
# Context
* Version of iperf3:
git de006004d4f431905cae311ddd2ca9a23b78669a

* Hardware:
x86_64
* Operating system (and distribution, if any):
Fedora 33

# Bug Report

* Expected Behavior
Correct error reported for connection failures
* Actual Behavior
Uses `errno` from non-related system calls

* Steps to Reproduce

1) Add iptables rule, for example.
2) Run it
```
$ strace iperf3 -c 1.2.3.4 2>&1 | tail -n 9
connect(4, {sa_family=AF_INET, sin_port=htons(5201), sin_addr=inet_addr("1.2.3.4")}, 16) = -1 ECONNREFUSED (Connection refused)
close(4)                                = 0
mmap(0x7df1fea80000, 65536, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7df1fea80000
mmap(0x7dfffea80000, 65536, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x7dfffea80000
write(-1, "\20", 1)                     = -1 EBADF (Bad file descriptor)
write(2, "iperf3: error - unable to send c"..., 68iperf3: error - unable to send control message: Bad file descriptor
) = 68
exit_group(1)                           = ?
+++ exited with 1 +++
```

davidBar-On:
@bmah888, I believe that adding **"`if (test->ctrl_sck != -1)`"** condition before the following code block in `iperf_set_send_state()` will solve the problem: https://github.com/esnet/iperf/blob/53a68308bad09a6a7c2758acfe6051123a865c68/src/iperf_api.c#L1631-L1635

PR #1131, opened 2021-03-21
Not sending state to to undefined socket and correct cookie buffer sizes

davidBar-On:
3.9+ latest

* Issues fixed (if any):
#1129

Combined two minor fixes:
1. Fix to issue #1129 - not sending state if control channel is not defined.
2. Add 1 byte to `make_cookie()` output buffer, as it creates a null terminated string (although the cookie is only used as a fixed `COOKIE_SIZE` length string).


davidBar-On:
PR includes errors so is canceled.  Replaced by #1132.

PR #1132, opened 2021-03-25
Fix issue 1129 for not sending stat to to undefined socket

davidBar-On:
3.9+ latest

* Issues fixed (if any):
#1129

Fix to issue #1129 - not sending state if control channel is not defined.
(Replacing PR #1131 which will be canceled.)


bmah888 (approved):
This looks good, it fixes one problem of trying to do operations on a socket that never got opened.

I actually found another one while in the process of testing this fix. But I wouldn't have found it if I wasn't reviewing this PR, so thank you, both for this fix and for motivating me to find another one.

PR #1137, opened 2021-04-10
fix: Fix a couple of buffer overrun hazards.

### LINKED ISSUE #1134: Buffer overflow hazard and incorrect return value in iperf_printf().
# Context
While discussing PR #1133, I realized there is a buffer overflow hazard and an incorrect return value on iperf_printf()

* Version of iperf3:
master as of 53a68308bad09a6a7c2758acfe6051123a865c68

* Hardware:
All

All

Found in code inspection. I have not verified if callers allow `iperf_printf()` to overflow buffers, but it appears possible.

# Bug Report

* Expected Behavior
1) Buffers should not overflow no matter how `iperf_printf()` is called.
2) The return value should (presumably) reflect the number of characters written to output.

* Actual Behavior
`iperf_printf()` uses an internal buffer `linebuffer[1024]` without bounds checking, and with incorrect bounds checking in vsnprintf().
`iperf_printf()` returns incorrect number of characters written.  The returned count does not include the timestamp or title written to output.

* Steps to Reproduce
None

* Possible Solution
Use of `snprintf()`
Correct use of `vsnprintf()`

_Please submit patches or code changes as a pull request._
I will attempt this fix unless someone jumps on it in the next few days.

bmah888:
Pointed out by @berkakinci.

Fixes #1134.

```
```

* Issues fixed (if any): #1134

See above.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4571,12 +4571,15 @@ iperf_printf(struct iperf_test *test, const char* format, ...)
                 return r0;
             r += r0;
 	}
-	va_start(argp, format);
-	r0 = vsnprintf(linebuffer + r, sizeof(linebuffer) - r, format, argp);
-	va_end(argp);
-        if (r0 < 0)
-            return r0;
-        r += r0;
+        /* Should always be true as long as sizeof(ct) < sizeof(linebuffer) */
+        if (r < sizeof(linebuffer)) {
+            va_start(argp, format);
+            r0 = vsnprintf(linebuffer + r, sizeof(linebuffer) - r, format, argp);
+            va_end(argp);
+            if (r0 < 0)
+                return r0;
+            r += r0;
+        }
 	fprintf(test->outfile, "%s", linebuffer);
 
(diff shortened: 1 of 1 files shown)


berkakinci (commented):
This part is fixed: now `linebuffer[1024]` won't overflow.
On setting `r` and `return r`:
Under `test->role == 'c'` There is another couple of missing counts at line 4548 and 4551.
The rest of this is getting in the weeds and would require a bunch of little changes: all of these functions can return `-1` on failure, so count wouldn't be correct.  Also, the `*nprint*` functions return the number that __would_ have been written_ if the buffer was large enough, so count wouldn't be correct if buffers weren't large enough.
If you're not averse to a bigger change: I think both the `role == 'c'` and `role == 's'` stanzas can share a lot more code.  Set up `linebuffer` in both cases.  Then only take the return value from a final (and single) `fprintf()` call.

bmah888:
The attempt to merge this pull request went sideways. Completely my fault. I ended up backing it out and then cherry-picking the individual commits one by one back onto the mainline.

Issue #1135, opened 2021-04-08
Why did the server take the initiative to disconnect from the client


sheva7777:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the iperf3 mailing list at iperf-dev@googlegroups.com or
question sites such as Stack Overflow
(http://www.stackoverflow.com/).  A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3:
iperf 3.9+ (cJSON 1.7.13)
* Hardware:
x86_64
* Operating system (and distribution, if any):
Centos 7.5
_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report
At first,the client1 and the server of iperf3 worked normally.But after receiving a RST packet from client2 on the port monitored by the server,the server sent a FIN packet to client1 immediately.Then The connection of client1 with server was interrupted

* Expected Behavior
A RST packet from client2 will not influence the connection of client1 with server
* Actual Behavior
A RST packet from client2 influence the connection of client1
client1:10.10.0.3
server:10.10.0.7
client2:10.10.0.4

```
    10.10.0.4.42830 > 10.10.0.7.3333: Flags [R.], cksum 0x1950 (correct), seq 1, ack 1, win 229, options [nop,nop,TS val 0 ecr 33553235], length 0
18:57:15.226954 IP (tos 0x0, ttl 64, id 37356, offset 0, flags [DF], proto TCP (6), length 52)
    10.10.0.7.3333 > 10.10.0.3.60746: Flags [F.], cksum 0x1444 (incorrect -> 0x1c3d), seq 1, ack 86918, win 1432, options [nop,nop,TS val 33553236 ecr 33566042], length 0
18:57:15.226998 IP (tos 0x0, ttl 64, id 54678, offset 0, flags [DF], proto TCP (6), length 52)
[...]

PR #1138, opened 2021-04-10
Server does not fail when rejecting connection during running test

davidBar-On:
3.9+ (self build)

* Issues fixed (if any):
#1135

Do not return error from `iperf_accept()` if it fails to send `ACCESS_DENIED` while running a test.  This is because the new connection request is not related to the running test, and is probably coming from another client or other networking activity (like ports scanning), and returning an error will terminate the active test.

A similar change may be required for the `acccept()` at the beginning of `iperf_accept()` when `test->ctrl_sck != -1`.  I dind't do it because I am not sure if this can happen because a client was able to close the socket before the `accept()` or it means there is a real problem.  Can be added if needed.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -150,10 +150,15 @@ iperf_accept(struct iperf_test *test)
 	/*
 	 * Don't try to read from the socket.  It could block an ongoing test. 
 	 * Just send ACCESS_DENIED.
+         * Also, if sending failed, don't return an error, as the request is not related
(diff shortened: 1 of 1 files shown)


bmah888:
@davidBar-On Thanks for doing the diagnosis on this problem, nice catch!

bmah888 (approved):
This looks good...the only change I would make is s/requist/request/ in the debug messages. I intend to merge this in the next day or two, if it's not fixed by then I can do it myself after the merge. Thanks!

Issue #862, opened 2019-04-29
iperf3 error :unable to read from stream socket: Resource temporarily unavailable


orenbigel:
Precondition: created alias IP on eth0 for both server and client, both are using the same version of iperf3.

Server:
iperf3 -s -V -B 162.243.144.198 

Client:
 iperf3 -c 162.243.144.198 -B 192.168.163.1  -u -b 1M -i 1 -t 120 -J

Result:
`{
  "start": {
    "connected": [],
    "version": "iperf 3.1.7",
    "system_info": "Linux ams-gw-3 3.10.0-957.10.1.el7.x86_64 #1 SMP Mon Mar 18 15:06:45 UTC 2019 x86_64",
    "timestamp": {
      "time": "Mon, 29 Apr 2019 01:03:41 GMT",
      "timesecs": 1556499821
    },
    "connecting_to": {
      "host": "162.243.144.198",
      "port": 5201
    },
    "cookie": "ams-gw-3.1556499821.535002.263807b47"
  },
  "intervals": [],
  "end": {},
  "error": "error - unable to read from stream socket: Resource temporarily unavailable"
}`

This is intermittent, some runs succeed in generating connection

joemce:
This error can occur when setting up UDP streams. During the CREATE_STREAMS state, the client sends a UDP packet to the server and the server sends back a UDP response. If either of these UDP packets are dropped, the client reports this error since no response is received from the server. This is considered a hard error. At the moment, no retransmission mechanism is used for UDP streams. 

A nice improvement to iperf3 would be a retry mechanism when setting up UDP streams. UDP packet loss on wireless networks is so common place. If you are creating multiple streams, any setup error for one stream will stops the whole test.

PR #1144, opened 2021-04-21
iperf_udp: implement UDP connect retry mechanism

olichtne:

* Issues fixed (if any): from our investigation we encountered issue #862 and this is an attempt to resolve that issue.

This implements a simple retry mechanism for the `iperf_udp_connect` function using a static iteration loop with a select call and attempting to resend the connect datagram.

This is mostly as a suggestion for a fix, there's probably a better way to implement some parts of this, for example the number of iterations should probably be either a constant or could be a new cli argument that could change the value. The retry mechanism could also be optionally disabled. Another idea is that the timeout value for the `select` function could also be  configurable, a 1 second wait here may not always be a good idea depending on the specific network that is being measured.1

I'm also not 100% sure if this doesn't potentially have side effects for the test initialization state transition on the slave side in case the connect datagram is not in fact lost and instead the reply is delayed or lost.

From my internal testing on servers this resolves our issues when running multiple iperf instances to simulate stress on the network stack by running multiple network applications.

### DIFF
--- a/src/iperf_udp.c
+++ b/src/iperf_udp.c
@@ -556,26 +556,49 @@ iperf_udp_connect(struct iperf_test *test)
     setsockopt(s, SOL_SOCKET, SO_RCVTIMEO, (struct timeval *)&tv, sizeof(struct timeval));
 #endif
 
-    /*
-     * Write a datagram to the UDP stream to let the server know we're here.
-     * The server learns our address by obtaining its peer's address.
-     */
-    buf = 123456789;		/* this can be pretty much anything */
-    if (write(s, &buf, sizeof(buf)) < 0) {
-        // XXX: Should this be changed to IESTREAMCONNECT? 
-        i_errno = IESTREAMWRITE;
-        return -1;
-    }
+    for (int i=0; i<5; i++){
+        /* retry the connection setup for up to 5 times */
 
-    /*
-     * Wait until the server replies back to us.
-     */
-    if ((sz = recv(s, &buf, sizeof(buf), 0)) < 0) {
-        i_errno = IESTREAMREAD;
-        return -1;
-    }
+        /*
+         * Write a datagram to the UDP stream to let the server know we're here.
+         * The server learns our address by obtaining its peer's address.
+         */
+        buf = 123456789;		/* this can be pretty much anything */
+        if (write(s, &buf, sizeof(buf)) < 0) {
+            // XXX: Should this be changed to IESTREAMCONNECT? 
+            i_errno = IESTREAMWRITE;
+            return -1;
+        }
 
-    return s;
+        int result;
+        fd_set read_set;
+        struct timeval timeout;
+
+        timeout.tv_sec = 1;
+
+        memcpy(&read_set, &test->read_set, sizeof(fd_set));
+        result = select(test->max_fd + 1, &read_set, NULL, NULL, &timeout);
+        if (result < 0 && errno != EINTR) {
+            i_errno = IESELECT;
+            return -1;
+        } else if (result > 0) {
+            /*
+             * Wait until the server replies back to us.
+             */
+            if ((sz = recv(s, &buf, sizeof(buf), 0)) < 0) {
+                i_errno = IESTREAMREAD;
+                return -1;
+            } else {
+                return s;
+            }
+        } else {
+            if (test->debug)
+                fprintf(stderr, "Retrying udp connection in 1s.");
+            sleep(1);
+        }
+    }
+    i_errno = IESTREAMREAD;
+    return -1;
 }
 
 


jtluka:
@olichtne , could you please update this pull request with a fix of uninitialized `timeout.tv_usec`:

```
@@ -575,6 +575,7 @@ iperf_udp_connect(struct iperf_test *test)
         struct timeval timeout;
 
         timeout.tv_sec = 1;
+        timeout.tv_usec = 0;
 
         memcpy(&read_set, &test->read_set, sizeof(fd_set));
         result = select(test->max_fd + 1, &read_set, NULL, NULL, &timeout);
```

Without this patch I get following error on aarch64:
```
iperf3: error - select failed: Invalid argument
```

src/iperf_udp.c
```diff
@@ -556,26 +556,49 @@ iperf_udp_connect(struct iperf_test *test)
...
-        return -1;
-    }
+    for (int i=0; i<5; i++){
+        /* retry the connection setup for up to 5 times */
+
+        /*
+         * Write a datagram to the UDP stream to let the server know we're here.
+         * The server learns our address by obtaining its peer's address.
+         */
+        buf = 123456789;		/* this can be pretty much anything */
```
talatb, line 566: Once we are running multiple processes of iperf, they will get the same buf for each process.
i would like to suggest adding +i to the buffer will solve the issue.

talatb (commented):
We had the same issue with multiple processes, we tried this PR and it works with small modification that addressed in the comments.

olichtne:
that sounds a little suspicious, i'm not sure how `+i` would help, can you describe your situation in a bit more detail? or how the `+i` is supposed to help?

secahtah:
On second thought, there is more to this. While testing using this today I discovered that the iperf3 UDP client just silently finishes the test even though there's 0 packets sent. Maybe more is required.

olichtne:
yeah, for our internal use, we keep using the fix from this PR that we have in a separate branch.

Since this didn't get more of a response from the maintainers of this project I didn't really look much deeper into this yet as I don't want to mess up any internals accidentally...

I do agree though on the cli argument being more comfortable for a real solution...


Not sure what you mean by your case of UDP client silently finishing with 0 packets sent, I think that may need more information/reproducer data to understand, and may potentially be a new issue completely.

davidBar-On:
> I'm also not 100% sure if this doesn't potentially have side effects for the test initialization state transition on the slave side in case the connect datagram is not in fact lost and instead the reply is delayed or lost.

I like the approach suggested by this PR, but I think that the above is the main issue with it and that the Stream creation protocol should be changed to make this suggestion feasible for general use.

If I understand `iperf_udp_accept()` code correctly , when the server receives the 4 bytes "datagram to the UDP stream to let the server know we're here", it uses the current control socket for the UDP Stream and creates a new control socket.  If the 4 bytes datagram is received by the server but the response is lost, then two of the possible scenarios are:
1. If the test is using only one stream (the default `-P 1`), then the retry of sending the 4 bytes datagram will be received by the server as a test datagram, and the server will not send a reply to it, so the test will not work.
2. If more then one stream is used, e.g. `-P 2`, the the retry datagram will be received by the server as initiation of the second stream, but the response to the client will be received by the client as an ack the the first stream.  Trying to initiate the second stream will fail since from the server point of view the two streams were already created.

The suggested approach probably works well when only the upload interface (client to server) is loaded, so the chance of loosing the the server's reply is low.

Note that a less elegant solution to the issue may be by running the client in a shell-script loop.

olichtne:
right, reading the explanation basically confirms that i don't entirely understand how the initiation protocol works which is why i didn't push for this harder.

But now after reading this explanation I am kind of questioning if this initiation protocol should even be in place this way for UDP.

To me thinking about it from an outside view of what I would expect for a UDP test is that a perf measurement tool would:
1. client would create a control connection and communicate the creation of individual opening of "streams" via different udp ports to the server
2. the server responds to this, confirming or failing this, but still via the control connection
3. on success the client simply starts sending the measurement PDUs via the individual udp sockets to the server

The point of this being that the UDP connections are not actual connections, and they're not established as part of the UDP protocol. You directly use them to measure their performance from the start.

In a situation that the udp packets don't arrive at the server you simply report:
* generator throughput: X mbps
* receiver throughput: 0 mbps

And close the connections on both sides via the control connection.

In my eyes, this situation is completely valid, it also would mean that since there is no "connection" creation it doesn't have to be retried.

davidBar-On:
I have submitted PR #1402 which should be an enhancement of this PR.  Although my system does not have the issues solved by the PR, I hope it is sufficiently tested.  Of course, it will help if the new PR will be tested on systems that does have this issues.

Issue #1143, opened 2021-04-21
Non JSON error message at the end of JSON output at the client-side of the measurement


zokl:
* Version of iperf3: master, 787919c4b3614b28f45587753f3a778a0157110c

* Hardware: x86_64

* Operating system (and distribution, if any): Linux, OpenWRT, iperf compiler from github repo

# Bug Report

* Expected Behavior

* Actual Behavior

Non JSON error message at the end of JSON output at the client-side of the measurement. For now, I found only described behavior for this message: "iperf3: error - unable to send control message: Broken pipe". Server-side produces valid JSON output. 

## Wrong JSON output on client-side
[client-4.iperf3.gz](https://github.com/esnet/iperf/files/6350988/client-4.iperf3.gz)

```
			"remote_total":	0,
			"remote_user":	0,
			"remote_system":	0
		},
		"sender_tcp_congestion":	"cubic"
	}
}
iperf3: error - unable to send control message: Broken pipe
```

## Good JSON output on server-side
[server-4.iperf3.gz](https://github.com/esnet/iperf/files/6350992/server-4.iperf3.gz)

```
				"bytes":	0,
				"bits_per_second":	0,
				"omitted":	false,
				"sender":	false
			}
		}],
	"end":	{
	},
	"error":	"error - select failed: Bad file descriptor"
}
```

* Steps to Reproduce

/usr/bin/iperf3 -c 1.2.3.4 --connect-timeout 1000 -t 3600 --logfile test-4.iperf3 -p 5201 -J --parallel 5 -i 1 --window 32k --set-mss 1400
/usr/bin/iperf3 -c 1.2.3.4 --connect-timeout 1000 -t 3600 --logfile test-2.iperf3 -p 5200 -J -R --parallel 5 -i 1 --window 32k --set-mss 1400

davidBar-On:
I tried to run the non-reverse test ("test-4.iperf3" log) on my local machine and there was no problem.  **Maybe try running using the latest iperf3 code** (lately some changes were done that that may be related).  If that won't help, maybe debug info (`-V -d`) can help.

[...]

PR #1146, opened 2021-05-18
Make sure Client logs terminating error message inside the JSON output

davidBar-On:
3.10

* Issues fixed (if any):
#1143

When an active test is terminating with error, make sure the Client logs the error before creating the JSON output, so the error message will be included inside the JSON output.


zokl:
When could this patch commit to master?

bmah888:
Thanks for the fix...looks good!

Issue #1159, opened 2021-06-08
Weird results with speed measured: `iperf3 -s -F file.out `


romanchyla:
# Context

I was following documentation to establish network speed; memory-memory; disk->memory; disk<-memory and in the last one was getting way too low results. 

It turns out that if I mount the drive over NFS I can achieve higher speeds than what iperf3 reports.

I was curious whether I'm doing something wrong or if this is a potentially a bug with iperf3

* Version of iperf3:

iperf 3.7 (cJSON 1.5.2)
Linux rbox3 5.8.0-53-generic #60-Ubuntu SMP Thu May 6 07:46:32 UTC 2021 x86_64
Optional features available: CPU affinity setting, IPv6 flow label, SCTP, TCP congestion algorithm setting, sendfile / zerocopy, socket pacing, authentication

* Hardware:

Please see this SF post to see the context/results/measurements:
https://serverfault.com/questions/1065820/very-slow-write-speed-nvme-drive-on-10g-network

# Bug Report

* Expected Behavior

Sending data over 10G with very fast (single) NVME drive on the target machine, I'd expect reported speed to be close to the 10G network connection.

* Actual Behavior

It was 10% - NVME disk can write >1000MB/s; iperf3 memory-memory shows 1GB/s transfers; but the actual write speeds were  ~100MB/s s

* Steps to Reproduce

server: `iperf3 -s -F /tmp/out`
client: `iperf3 -c server -t 30`

davidBar-On:
It seems that a main cause of the problem is the `fsync` done after each write to the file:
https://github.com/esnet/iperf/blob/3b31c4b6ac40481fb6057ed75b8dd6af9c150b34/src/iperf_api.c#L4279-L4280

Using a local built iperf3 on my local computer with SSD disk, the throughput is about 13MB/sec with `fsync` and over 100MB/sec when the `fsync` is removed.

I will submit a PR with a proposed enhancement to make the `fsync()` optional.

romanchyla:
Thank you! I'll try to test it out and update measured results (probably once a new release is available)

PR #1160, opened 2021-06-09
Allow write to file without fsync to improve performance

davidBar-On:
3.10 local build

* Issues fixed (if any):
 #1159 

Add option `--no-fsync` to allow better throughput when **receiver** output is a disk file (`-F` option).  When the option is set, no `fsync` is done after write to the file.  Default remains `fsync` after each write to the file.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -274,7 +274,8 @@ struct iperf_test
     int       server_port;
     int       omit;                             /* duration of omit period (-O flag) */
     int       duration;                         /* total duration of test (-t flag) */
-    char     *diskfile_name;			/* -F option */
+    char      *diskfile_name;			/* -F option */
+    int       no_fsync;                         /* --no_fsync option */
     int       affinity, server_affinity;	/* -A option */
 #if defined(HAVE_CPUSET_SETAFFINITY)
     cpuset_t cpumask;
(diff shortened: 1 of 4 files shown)


bmah888 (commented):
Thanks for the PR! I am just starting to look at this now. Just making a note that we'll need to do an update to the manual page to document the new `--no-fsync` option.

I've been growing the opinion over the past few years that iperf3 is getting too many command-line options, which can interact in weird and non-intuitive ways. I don't see a problem in this particular case, but it's not really helping this feeling.

bmah888 (commented):
Actually I wonder if we should just take out the fsync(2) call and not even bother having an option.

(I tested the patch, and it performs as advertised, but I'm wondering "wait a minute, why do we need this fsync() thing in the first place?")

davidBar-On:
> Actually I wonder if we should just take out the fsync(2) call and not even bother having an option

I assumed that fsync(2) was added because there where issues that it solved.  If this is not the case, then I agree just removing it is a better approach (especially if this is what iperf2 does).

bmah888:
> I assumed that fsync(2) was added because there where issues that it solved. If this is not the case, then I agree just removing it is a better approach (especially if this is what iperf2 does).

Unfortunately I am not able to tell why that fsync(2) is there. The commit that added it (979405373d872ed08238f279453561c131963264) doesn't really give any explanation.

From what I can tell, iperf2 doesn't even have any functionality to write the output to a file, and there isn't an fsync(2) call anywhere in its source code, so I don't think we have to worry about compatibility.

I'm going to get rid of that fsync(2) in iperf3.

Thanks again for looking into this issue!

bmah888:
See #1176.

PR #1162, opened 2021-06-11
Fix --dscp

presto8:
Maybe I am missing something obvious, but when I call iperf3 with --tos XX, where XX is DSCP << 2, it works fine. But if I call iperf3 with --dscp DSCP (e.g., with the unshifted value) it doesn't work. In looking at the code, I am not seeing where the DSCP is shifted left 2 bits.

### DIFF
--- a/src/dscp.c
+++ b/src/dscp.c
@@ -136,10 +136,11 @@ parse_qos(const char *cp)
 			return ipqos[i].value;
 	}
 	/* Try parsing as an integer */
+    /* Max DSCP value is 2**6 - 1 */
 	val = strtol(cp, &ep, 0);
-	if (*cp == '\0' || *ep != '\0' || val < 0 || val > 255)
+	if (*cp == '\0' || *ep != '\0' || val < 0 || val > 63)
 		return -1;
-	return val;
+	return val << 2;
 }
 
 const char *
--- a/src/iperf3.1
+++ b/src/iperf3.1
@@ -336,7 +336,8 @@ i.e. 52, 064 and 0x34 all specify the same value.
 .TP
 .BR "--dscp " \fIdscp\fR
 set the IP DSCP bits.  Both numeric and symbolic values are accepted. Numeric
-values can be specified in decimal, octal and hex (see --tos above).
+values can be specified in decimal, octal and hex (see --tos above). To set
+both the DSCP bits and the ECN bits, use --tos.
 .TP
 .BR -L ", " --flowlabel " \fIn\fR"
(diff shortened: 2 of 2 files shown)


presto8:
Hi @dtaht , I think you are the original author of the --dscp code? Would appreciate any insight you have on this issue. Thank you!

dtaht:
Yes, it should be shifted left. On the other hand, being able to arbitrarily set the ecn bits for testing for ecn compliant aqms on the path (rfc3168 , or l4s, or SCE) is dang useful. So I would hold off on doing this fix until there is an option to set the ecn bits also. There's a lot of ecn related testing going on at the moment.

https://github.com/heistp/l4s-tests

presto8:
Thanks for the feedback! One thing is that the documentation says:

```
--dscp dscp
              set the IP DSCP bits.
```
Which is not really what is being done? I see the point about wanting to set the ECN bits, but perhaps adding --ecn could be an option? Or update the documentation to clarify that --dscp sets both the IP and ECN bits (in which case, it is identical to --tos).

dtaht:
Ah, the --tos value should suffice and you should patch --dscp

You should also verify that the code is correct for ipv6 (IPV6_TCLASS is a big endian int)

presto8:
Great, thanks for the feedback! I will update the patch, check the ipv6 code correctness, and also update the documentation maybe to point out that --tos can be used to set all the bits, whereas --dscp just sets the dscp bits.

bmah888:
@presto8 Thanks for the PR! Would you say the work for this issue is still ongoing? If so, that's totally fine, I will just delay merging this until it's ready.

presto8:
Thanks @bmah888 ! I still want to check a few things, so please hold off for a bit. I will post a note here when it is ready! I have an IPv4 and IPv6 testbed set up where I can run some tests and see if it works as anticipated with actual hardware testing.

presto8:
Hi, I apologize for the extremely long delay in responding. I was waiting to have access to some test machines.

The good news is that I tested the code and it works perfectly. The test I ran was to set "--dscp 40" which I then inspected with wireshark and observed the dsfield set to 0xa0 and the ds codepoint (dscp) set to 40, exactly as they should be.

Also, in testing it, I realized I think maybe the initial confusion about the behavior. The entire ToS field can also be called the "Differentiated Services Field". The dsfield is further comprised of the DSCP (DS codepoint) plus the ECN.

So --dsfield could be considered an alias for --tos. But definitely --dscp should be fixed as indicated in this pull request.

PR #1166, opened 2021-06-14
Add MPTCP support with the --multipath flag

pabeni:
master

* Issues fixed (if any):
[]

Recent version of the Linux kernel (5.9) gained MPTCP support, this change
 allows easy testing such protocol. 
MPTCP is strongly tied to TCP, and the kernel APIs are almost the same,
so this change does not implement a new 'struct protocol'. Instead it just
extend the TCP support to optionally enable multipath via the --multipath or 
-m switch.

The only required dependency is the 'IPPROTO_MPTCP' protocol number
definition, which should be provided by the netinet/in.h header.
To keep things simple, just conditionally provide the required
protocol, if the system header does not have it yet.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -315,6 +315,7 @@ struct iperf_test
     int	      udp_counters_64bit;		/* --use-64-bit-udp-counters */
     int       forceflush; /* --forceflush - flushing output at every interval */
     int	      multisend;
+    int	      multipath;			/* -m option - multi-path variant */
     int	      repeating_payload;                /* --repeating-payload */
     int       timestamps;			/* --timestamps */
     char     *timestamp_format;
--- a/src/iperf3.1
+++ b/src/iperf3.1
@@ -228,6 +228,10 @@ run in client mode, connecting to the specified server.
 By default, a test consists of sending data from the client to the
 server, unless the \-R flag is specified.
 .TP
+.BR -m ", " --multipath " "
+use multipath variant for the current protocol. This only applies to
+TCP and enables MPTCP usage.
+.TP
 .BR --sctp
 use SCTP rather than TCP (FreeBSD and Linux)
 .TP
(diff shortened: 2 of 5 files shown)


teto:
I've tested it locally and it worked fine. Needs a rebase and maybe reword the commit too.,

bmah888:
A belated thank-you for this pull request. It sounds useful, although I'm not sure where MPTCP would be used in our (ESnet's) network environment. (or for that matter how we would test whether this works or not.)

The main change I can think of at this point is that you should be testing for MPTCP support (I guess you can check for the presence of `IPPROTO_MPTCP`), and only allowing someone to pass `-m` or `--multipath` if it's actually present. The way it works now, it looks like you just assume that the host supports MPTCP regardless of whether `IPPROTO_MPTCP` is defined or not, which isn't the way we detect and enable features across different codebases.

If you look in `configure.ac` you can see an example of this where we test for the presence of various socket options such as `SO_MAX_PACING_RATE` or `SO_BINDTODEVICE`.

teto:
> If you look in configure.ac you can see an example of this where we test for the presence of various socket options such as SO_MAX_PACING_RATE or SO_BINDTODEVICE.

wont that be an issue when cross compiling ? I would think it's not necessary as one needs to check at runtime anyway

ZerxXxes:
This is great! The first use case I can think of is when both server and client is dual stack. If I understand it correctly MPTCP will create a subflow per L3 path so iperf will push traffic over IPv4 and IPv6 at the same time.
In some cases one of them is a bit slower due to CGNAT or just suboptimal routing (when testing long haul links, like cross continents) and then MPTCP will just push more of the packets via the path that works best.
Looking forward to test this!

geliangtang:
https://github.com/esnet/iperf/pull/1659

pabeni:
obsoleted by [#1661](https://github.com/esnet/iperf/pull/1661)

Issue #1167, opened 2021-06-15
JSON output does not include bitrate value (--bitrate)


TheRealDJ:
# Context
* Version of iperf3: 3.9, Also checked with MASTER as client and server.
* Hardware: ODROID-C4 (NA?)
* Operating system (and distribution, if any): 20.04 LTS (NA?)
* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.): NA

# Bug Report
* Expected Behavior
JSON output contains bitrate value passed using --bitrate used for the test.
* Actual Behavior
JSON output does NOT contain the bitrate value.
* Steps to Reproduce
```
# ./iperf3 --client $IPERFSERVER --udp --time 1 --bitrate 10M --json | jq .start.test_start
{
  "protocol": "UDP",
  "num_streams": 1,
  "blksize": 1448,
  "omit": 0,
  "duration": 1,
  "bytes": 0,
  "blocks": 0,
  "reverse": 0,
  "tos": 0
}

# ./iperf3 --client $IPERFSERVER --time 1 --bitrate 10M --json | jq .start.test_start
{
  "protocol": "TCP",
  "num_streams": 1,
  "blksize": 131072,
  "omit": 0,
  "duration": 1,
  "bytes": 0,
  "blocks": 0,
  "reverse": 0,
  "tos": 0
}
```
* Possible Solution
Add test bitrate value to {start}{test_start} JSON.

I'm submitting a PR now.

PR #1168, opened 2021-06-15
Test bandwidth (--bitrate) added to JSON output

TheRealDJ:
Fixes https://github.com/esnet/iperf/issues/1167 ... Bandwidth (--bitrate) parameter missing from JSON output.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -816,7 +816,7 @@ void
 iperf_on_test_start(struct iperf_test *test)
 {
     if (test->json_output) {
-	cJSON_AddItemToObject(test->json_start, "test_start", iperf_json_printf("protocol: %s  num_streams: %d  blksize: %d  omit: %d  duration: %d  bytes: %d  blocks: %d  reverse: %d  tos: %d  bandwidth: %d", test->protocol->name, (int64_t) test->num_streams, (int64_t) test->settings->blksize, (int64_t) test->omit, (int64_t) test->duration, (int64_t) test->settings->bytes, (int64_t) test->settings->blocks, test->reverse?(int64_t)1:(int64_t)0, (int64_t) test->settings->tos, test->settings->rate));
+	cJSON_AddItemToObject(test->json_start, "test_start", iperf_json_printf("protocol: %s  num_streams: %d  blksize: %d  omit: %d  duration: %d  bytes: %d  blocks: %d  reverse: %d  tos: %d  target_bitrate: %d", test->protocol->name, (int64_t) test->num_streams, (int64_t) test->settings->blksize, (int64_t) test->omit, (int64_t) test->duration, (int64_t) test->settings->bytes, (int64_t) test->settings->blocks, test->reverse?(int64_t)1:(int64_t)0, (int64_t) test->settings->tos, test->settings->rate));
     } else {
 	if (test->verbose) {
 	    if (test->settings->bytes)
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -816,7 +816,7 @@ void
 iperf_on_test_start(struct iperf_test *test)
 {
     if (test->json_output) {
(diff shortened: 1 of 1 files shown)


ejv:
Thanks for the contribution.

During TCP tests and when specified by the user, the desired target bitrate is recorded in the JSON dict under the key `target_bitrate` (not nested in `test_start`). In the spirit of consistency, would it not be preferable to standardize around this and add a `target_bitrate` for UDP tests, as well?

Also, I'm not supportive of calling the new key you've added "bandwidth". Bandwidth typically refers to a data link's overall available capacity: `target_bitrate` seems more precise.

TheRealDJ:
The command line option used to be `--bandwidth` (which still works) and is now `--bitrate`. I'm fine with `target_bitrate` but it should be in under `test_start`, and should always be listed in the JSON. PR updated with `target_bitrate` key name.

TheRealDJ:
I did not remove existing `target_bitrate` (not nested) key as to not break any existing implementations.

bmah888:
Hey @TheRealDJ thanks for the PR! First, I'm trying to understand what the bug is here. Is the problem that `target_bitrate` is included in `start` for TCP tests, but it's not included anywhere for UDP tests?  If so, that's consistent with what I'm seeing and I agree that's a bug.

Around line 896 of `src/iperf_api.c` is the following code:
```
    if (test->json_output) {
	cJSON_AddStringToObject(test->json_start, "cookie", test->cookie);
        if (test->protocol->id == SOCK_STREAM) {
	    if (test->settings->mss)
		cJSON_AddNumberToObject(test->json_start, "tcp_mss", test->settings->mss);
	    else {
		cJSON_AddNumberToObject(test->json_start, "tcp_mss_default", test->ctrl_sck_mss);
	    }
        if (test->settings->rate)
            cJSON_AddNumberToObject(test->json_start, "target_bitrate", test->settings->rate);
        }
```
The main bug here is that the `if (test->settings->rate)` conditional is actually included inside the `if (test->protocol-> == SOCK_STREAM)` conditional. So we only put `target_bitrate` into the JSON if we're doing TCP. (As an aside I'm not sure what happens in the case of SCTP.)

A second bug is that we should probably just print `target_bitrate` even if it's `0`.

So without having thought about this very much, I'm inclined to solve the problem by moving the closing brace on the last line of the code snippet up two lines, getting rid of the `if (test->settings->rate)` line, and then fixing up the indentation on the `target_bitrate` line. I think that will fix the problem.

The PR that you submitted puts `target_bitrate` into the `test_start` structure. Is there a particular reason it has to be there? Or would it be OK to have `target_bitrate` in`start`?

TheRealDJ:
I put it in `test_start` because that is where all of the rest of the test parameters are listed. To be honest, I didn't even notice it listed in the json root until pointed out by @ejv ... I was looking in `test_start`. As to not break anyone's existing implimentation, I didn't move the root `target_bitrate`, I only added to the `test_start` section.

bmah888:
> I put it in `test_start` because that is where all of the rest of the test parameters are listed. To be honest, I didn't even notice it listed in `start` until pointed out by @ejv ... I was looking in `test_start`. As to not break anyone's existing implementation, I didn't move the `start`.`target_bitrate`, I only added to the `test_start` section. I guess it is ok to have it in start, but test_start seems like a better place (just my opinion).
> 
> I agree on the other 2 points ... Always list the value in the JSON, and include for all protocols.

Hmmm...I think I see your point, at least I'm about 80% there.

bmah888 (changes requested):
OK. If you can put a (int64_t) cast before `test->settings->rate` in the line you changed, we'll call this good to go. I might want to do a follow-up commit to fix the `target_bitrate` in the `start` structure as I described in comments, just to keep everything consistent. Thanks!

bmah888 (approved):
Looks good, thanks!

PR #1174, opened 2021-06-25
Added {end}{sum_sent} and {end}{sum_received} to UDP JSON output

TheRealDJ:
* Version of iperf3:
**master**

* Issues fixed (if any):
**JSON output missing {end}{sum_sent} and {end}{sum_received} for UDP tests**

Changes made in `iperf_api.c` adding {end}{sum_sent} and {end}{sum_received} JSON objects with appropriate values for UDP tests. I did NOT remove or modify the existing {end}{sum} object to preserve any existing implementations. Existing {end}{sum} always shows sending side's byte count and bits_per_sec which is misleading in many situations.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -3710,9 +3710,14 @@ iperf_print_results(struct iperf_test *test)
                 else {
                     lost_percent = 0.0;
                 }
-                if (test->json_output)
+                if (test->json_output) {
                     cJSON_AddItemToObject(test->json_end, "sum", iperf_json_printf("start: %f  end: %f  seconds: %f  bytes: %d  bits_per_second: %f  jitter_ms: %f  lost_packets: %d  packets: %d  lost_percent: %f sender: %b", (double) start_time, (double) receiver_time, (double) receiver_time, (int64_t) total_sent, bandwidth * 8, (double) avg_jitter * 1000.0, (int64_t) lost_packets, (int64_t) total_packets, (double) lost_percent, stream_must_be_sender));
-                else {
+                    /* Modifications to add sum_sent and sum_received to UDP JSON results - DJ */
+                    /* For historical reasons (and hysterical raisins), "sum" section NOT REMOVED */
+                    cJSON_AddItemToObject(test->json_end, "sum_sent", iperf_json_printf("start: %f  end: %f  seconds: %f  bytes: %d  bits_per_second: %f  jitter_ms: %f  lost_packets: %d  packets: %d  lost_percent: %f  sender: %b", (double) start_time, (double) sender_time, (double) sender_time, (int64_t) total_sent, (double) total_sent * 8 / sender_time, (double) 0.0, (int64_t) 0, (int64_t) sender_total_packets, (double) 0.0, 1));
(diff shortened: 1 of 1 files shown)


TheRealDJ:
I emailed you as well ... When running a reverse UDP test, the existing {end}{sum} is useless. This test was run on an ODROID-C4 with a 1GbE interface as the client, and a server with 10GbE. {end}{sum} indicates a result of 2.83Gbps which is clearly not possible over 1GbE. My modifications clearly show that the sender transmitted 2.83Gbps, but the receiver received 956Mbps.
```
{start}{test_start}
	blksize: 1448
	blocks: 0
	bytes: 0
	duration: 5
	num_streams: 40
	omit: 5
	protocol: UDP
	reverse: 1
	target_bitrate: 250000000
	tos: 0
{end}{sum} - ESNet Master
	bits_per_second: 2858356221.91022
	bytes: 1791813120
	end: 4.999296
	jitter_ms: 0.130597793561623
	lost_packets: 820880
	lost_percent: 66.4866432215017
	packets: 1234654
	seconds: 4.999296
	sender: 0
	start: 0
{end}{sum_sent} - Modifications
	bits_per_second: 2858356221.91022
	bytes: 1791813120
	end: 5.014947
	jitter_ms: 0
	lost_packets: 0
	lost_percent: 0
	packets: 1234654
	seconds: 5.014947
	sender: 1
	start: 0
{end}{sum_received} - Modifications
	bits_per_second: 956521302.19935
	bytes: 597741640
	end: 4.999296
	jitter_ms: 0.130597793561623
	lost_packets: 820880
	lost_percent: 66.4866432215017
	packets: 1233685
	seconds: 4.999296
	sender: 0
	start: 0
```

bmah888 (commented):
Just looking at this for the first time now. I liked that you kept the same basic structure as the original `sum` object. I need to do some testing but so far so good. Part of me wants to throw away everything in this program that generates output and start over, but that isn't likely to happen.

bmah888:
@arlake228 , @mfeit-internet2 do you have any feedback or comments on this PR? It basically fixes up some result reporting for UDP tests by adding new members to the JSON but keeps the old semi-buggy JSON for backwards compatibility. Wondering if this would be helpful or useful for the pScheduler use cases. I'm inclined to commit this as-is but wanted to give you two the chance to weigh in on it.

mfeit-internet2:
@bmah888 That won't cause problems here.  The iperf3 plugin doesn't differentiate between TCP and UDP when parsing the results.  It tries to use use `.end.sum_sent`, then `.end.sum` and, if neither exists, will pitch a fit.

So if you want to incorporate this change, it won't cause us any problems other than more-accurate results.  :-)

bmah888 (approved):
Looks good. I might do a little editing of comments after the merge but the functionality seems like what we want. Thanks for the PR!

TheRealDJ:
I included the "and hysterical raisins" which was a typo of "historical reasons" from [iperf_api.c line 3558](https://github.com/esnet/iperf/blob/master/src/iperf_api.c#L3558). I laughed out loud by myself for about 5 minutes, so I had to include it.

PR #1178, opened 2021-07-03
Reliable receiving (-r) data as sent by sender

hanvari:

* Issues fixed (if any): N/A

**Reliable receiving (-r) data as sent by sender**

Two changes implemented:

1. On client, Following TEST_END signal, it sends the total number of bytes sent for server over control-socket.

2. On server,
    a. following TEST_END it receives the total number of bytes and sets to test->settings->bytes
    b. (if reliability requested) it defers the test termination and cleanup until that exact number of bytes are received (timer-based solution)

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -275,6 +275,7 @@ struct iperf_test
     int       omit;                             /* duration of omit period (-O flag) */
     int       duration;                         /* total duration of test (-t flag) */
     char     *diskfile_name;			/* -F option */
+    int       reliable_receive_in_full;     /* -r option. For reciever only. */
     int       affinity, server_affinity;	/* -A option */
 #if defined(HAVE_CPUSET_SETAFFINITY)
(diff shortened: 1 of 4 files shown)


hanvari:
@bmah888 is this going to get merged?

bmah888:
Hi @havnari, apologies for the delay. Thanks for sending this pull request. I'm going to respectfully decline merging this PR and #1179. This is mostly because the changes introduce some additional complexity, don't really support the main purpose of iperf3 (which is just to be a network benchmark), and seem to introduce some incompatibility with earlier versions of iperf3 by changing interactions over the control channel.

I think this is an interesting feature that others would like, so I encourage you to keep your version alive and available. Because we've gotten to the point in the mainline iperf3 development where we're trying not to make a lot of changes, maintaining a fork of iperf3 shouldn't be difficult. Thanks again!

PR #1179, opened 2021-07-03
Partial file transfer

hanvari:

* Issues fixed (if any): N/A

**Optional sender file seek and receiver append parameters with -F**

File transfer `-F` switch updated to take an optional second parameter, 
`-F, --file name[,(append|#)]`
The parameter after comma will either be,
- `#` the number of bytes to seek in the file and start sending from specific offset (for sender)
- `append` to append newly received data to the end of file in receiver (default behaviour is to overwrite the file)

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1326,6 +1326,23 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 		client_flag = 1;
                 break;
             case 'F':
+                comma = strchr(optarg, ',');
+                if (comma) {
+                    *comma = '\0';
+                    ++comma;
+                    // If present second paramter may either be,
+                    //  - 'aapend' (for receiver)
+                    //  - # to seek in the file (for sender)
+                    if( strcmp(comma, "append")==0 ){
+                        // append to file, not overwriting
+                        test->diskfile_append = 1;
+                    }
+                    else{
+                        // seek in file
+                        test->diskfile_seek = atoi(comma);
+                    }
+                }
+                // process first parameter which is a file path
                 test->diskfile_name = optarg;
                 break;
             case OPT_IDLE_TIMEOUT:
@@ -1572,6 +1589,8 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
         struct stat st;
         if( stat(test->diskfile_name, &st) == 0 ){
             iperf_size_t file_bytes = st.st_size;
+            // if we are seeking in file, decrease file size by seek value
+            file_bytes -= test->diskfile_seek;
             test->settings->bytes = file_bytes;        
             if (test->debug)
                 printf("End condition set to file-size: %d bytes\n", test->settings->bytes);
@@ -2637,6 +2656,8 @@ iperf_defaults(struct iperf_test *testp)
     testp->omit = OMIT;
     testp->duration = DURATION;
     testp->diskfile_name = (char*) 0;
+    testp->diskfile_seek = 0;
+    testp->diskfile_append = 0;
     testp->affinity = -1;
     testp->server_affinity = -1;
     TAILQ_INIT(&testp->xbind_addrs);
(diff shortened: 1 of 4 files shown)


src/iperf_api.c
```diff
@@ -1560,6 +1572,25 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
     if (!rate_flag)
 	test->settings->rate = test->protocol->id == Pudp ? UDP_RATE : 0;
 
+    /* if no bytes or blocks specified, nor a duration_flag, and we have -F,
```
hanvari, line 1575: @bmah888 Bruce, This piece assumes PR #1116 merged before this PR

davidBar-On:
> Adding -K (sender file seek) and -a (receiver append to file)

I like the suggested change.  Per [this comment](https://github.com/esnet/iperf/pull/1160#pullrequestreview-694419724) it may be better to extend the `-F` option instead of adding new options.  Instead of adding `-K` and `-a` the seek-offset and append flag can be added as an **optional** parameters to the `-F`: `-F <name>[/[<seek offset>]/[<append flag>]]`.  This approach is similar to how `-b, --bitrate #[KMG][/#]` is implemented.

hanvari:
> > Adding -K (sender file seek) and -a (receiver append to file)
> 
> I like the suggested change. Per [this comment](https://github.com/esnet/iperf/pull/1160#pullrequestreview-694419724) it may be better to extend the `-F` option instead of adding new options. Instead of adding `-K` and `-a` the seek-offset and append flag can be added as an **optional** parameters to the `-F`: `-F <name>[/[<seek offset>]/[<append flag>]]`. This approach is similar to how `-b, --bitrate #[KMG][/#]` is implemented.

Thank you @davidBar-On  for the feedback. I like that idea, but with `-F` the `optarg`, being a file path, will already include forward slash  in most cases. As such `/` may not be used as the parameter separator. And I'm not sure if that may cause confusion to use any other separator (.e.g `|`) as being inconsistent with the format used in `-b`.
Any thoughts?

davidBar-On:
> `-F` the `optarg`, being a file path, will already include forward slash in most cases.

I didn't thnk about that ... `--affinity` is using a comma to separate between the arguments, so maybe `,` can be used also for `-F`.

hanvari:
> I didn't thnk about that ... `--affinity` is using a comma to separate between the arguments, so maybe `,` can be used also for `-F`.

I have updated the PR to take those as optional parameters with `-F`. 
Thanks again for the feedback.

hanvari:
@bmah888 This PR as it is is a bit contaminated with the proposed changes in #1116. 
If it looks good as it is, please wait until that one is getting merged (or closed otherwise) first and then I could rebase this one.

Also, where manual page and help output are located in the codebase to be updated?

bmah888:
I acknowledge seeing this PR although I haven't really examined it closely. I am getting a little (OK more than a little) concerned about scope creep, and the addition of use cases that have, over the last few years, made iperf3 increasingly complicated and hard to maintain.

bmah888:
> Also, where manual page and help output are located in the codebase to be updated?

The manual page is `src/iperf3.1` and the help text is embedded in `src/iperf_locale.c`.

There's a copy of the manual page embedded in `docs/invoking.rst` but you don't need to worry about that, I usually regenerate that part of the file from the current manual page around the time of a release.

hanvari:
> I acknowledge seeing this PR although I haven't really examined it closely. I am getting a little (OK more than a little) concerned about scope creep, and the addition of use cases that have, over the last few years, made iperf3 increasingly complicated and hard to maintain.

I understand your concern. This PR, along with #1178 (reliable data transfer), would enable users to use iperf3 as before, while being able to transfer files, partially or completely. I have seen the reliable transfer being brought up in the conversations and issues a couple of times in different contexts. I wanted to implement this feature for a my research project, and thought maybe there is an interest in the community for this feature. 

Please consider to merge or close this PR at your own judgement.

If any further context or information about the implementation is needed I'd be happy to provide clarificaiton.

hanvari:
PR Rebased and documentation entries added for the new optional second parameter with -F

hanvari:
@bmah888 any decision on if this is going to get merged?

bmah888:
Thanks for sending this pull request! I'm going to close this as described in comments on #1178.

Issue #1157, opened 2021-06-03
Support address%interface syntax


bmah888:
Enhancement request prompted by an email from Bob McMahon (iperf2 maintainer):

We used the %<dev> for both tx and and rx device binding in iperf 2. Just something to consider for consistency between the two tools.

To affect the physical output interface (e.g. dual homed systems) either use -c <host>%<dev> (requires root) which bypasses this host route table lookup, or configure policy routing per each -B source address and set the output interface appropriately in the policy routes. On the server or receive, only packets destined to -B IP address will be received. It's also useful for multicast. For example, iperf -s -B 224.0.0.1%eth0 will only accept ip multicast packets with dest ip 224.0.0.1 that are received on the eth0 interface, while iperf -s -B 224.0.0.1 will receive those packets on any interface, Finally, the device specifier is required for v6 link-local, e.g. -c [v6addr]%<dev> -V, to select the output interface.

davidBar-On:
I am trying to understand what this change request means and I have some questions:

1. As far as I understand, the main change is an alternative to defining the `--bind-dev` value by defining it as `%` extension to the `--bind` or `-c` IP address.  Is this correct?

2. Looking briefly at iperf2 code, it seems that the main functionality change is for IPv6, in case the IP addr is **linklocal**.  In this case, the `--bind-dev` (if defined) is appended to the IP addr, following a `%`.

3. I didn't understand what the following means and if it requires further changes:
> .... use -c % (requires root) which bypasses this host route table lookup ...

PR #1180, opened 2021-07-06
Fix to support address%interface syntax

davidBar-On:
Latest 3.10.1

* Issues fixed (if any):
#1157

Simple suggested change to support `<address>%<device>` format for both `-c` and `-B` options.  As far as I understand issue #1157 and the related iperf2 implementation, the change supports at least the main suggested functionality.

No change was done to the device binding functionality, assuming current support is sufficient, although it may be that iperf2 support extended binding functionality - e.g. it uses two separate binding devices for Tx (from `-c`) and Rx (from `-B`).

The changes are:

1. **`-c`**: verify that `%device` is included only when the host address is IPv6 link-local.  No real functionality was added/changed.  (I was not able to test IPv6 link-local functionality, but according to iperf2 implementation this should be o.k.  In in any case, since there is no real functionality change there is no risk of iperf3 functionality degradation.)

2.  **`-B`**: allow using the `<ip>%<device>` (only when `SO_BINDTODEVICE` is supported).  When `%<device>` is specified handle it as it was defined by `--bind-dev`.  No check is done about whether device was specified by both `-B` and `--bind-dev` - the last one defined in the command line will be used.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1019,6 +1019,8 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
     char* comma;
 #endif /* HAVE_CPU_AFFINITY */
     char* slash;
+    char *p, *p1;
+    struct in6_addr ipv6_addr;
     struct xbind_entry *xbe;
     double farg;
     int rcv_timeout_in = 0;
@@ -1101,6 +1103,26 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                 }
 		iperf_set_test_role(test, 'c');
 		iperf_set_test_server_hostname(test, optarg);
+                // hostname should be extended with `%<dev>` for IPv6 link-local addresses
+                if ((p = strtok(optarg, "%")) != NULL) {
+                    p1 = strtok(NULL, "%");
+                    if (inet_pton(AF_INET6, p, &ipv6_addr) == 1) {
+                        if (IN6_IS_ADDR_LINKLOCAL(&ipv6_addr)) {
+                            if (!p1) {
+                                i_errno = IEHOSTDEV;
+                            }
+                        } else if (p1) {
+                            i_errno = IEHOSTDEV;
+                        }
+                    } else {    //IPv4
+                        if (p1) {
+                            i_errno = IEHOSTDEV;
+                        }
+                    }
+                    if (i_errno ==  IEHOSTDEV) {
+                        return -1;
+                    }                   
+                }
                 break;
             case 'u':
                 set_protocol(test, Pudp);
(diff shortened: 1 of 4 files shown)


TheRealDJ:
Would this interfere with IPv6 testing using Link Local address? `iperf3 --client fe80::XXX%eth0` works today.

bmah888:
> Would this interfere with IPv6 testing using Link Local address? `iperf3 --client fe80::XXX%eth0` works today.

I'll be darned.  I didn't realize that IPv6 LL worked out of the box.  (Or I forgot...must be old age creeping up on me.)

TheRealDJ:
I do see `if (IN6_IS_ADDR_LINKLOCAL(&ipv6_addr)) {` in the code. Maybe @davidBar-On covered this in the PR.

bmah888:
Just did some more testing. iperf3 on macOS (Big Sur), Ubuntu 20.04, and FreeBSD 12 are all capable of communicating using IPv6 link-local addresses using the `fe80::xxxx%enp0s3` style syntax, using unmodified code from the `master` branch.

davidBar-On:
> I do see if (IN6_IS_ADDR_LINKLOCAL(&ipv6_addr)) { in the code. Maybe @davidBar-On covered this in the PR.

> Just did some more testing. iperf3 on macOS (Big Sur), Ubuntu 20.04, and FreeBSD 12 are all capable of communicating using IPv6 link-local addresses using the fe80::xxxx%enp0s3 style syntax, using unmodified code from the master branch.

My understanding is that works today `-c <IPv6 linklocal>%<device>` because there is no validity check of the host ip address.  Therefore adding `%device>` works for IPV6 linklocal addresses but fails with any other type of ip address (as I wrote I was not able to verify that).  The change I did regarding `-c` is to verify that if `%<device>` was added then the host address is IPv6 linklocal.

(The other change I did is to allow defining the bind device by `%<device>` appended to the `-B` ip address.)

bmah888:
I might have misunderstood what the PR was doing, thanks @davidBar-On for the additional comments. Admittedly I didn't give the code a close reading...I need to go back and look at it more carefully. Thanks!

TheRealDJ:
I had to read a little closer myself. IPv6 link local % interface validation works ... For what it's worth, I like it. Guess I never attempted an IPv6 LL test without the `%interface`.
With esnet master (wrapped with `time` to show timeout duration):
```
# ./iperf3 --client fe80::XXXX%en0
<Test runs as expected>

# time ./iperf3 --client fe80::XXXX
iperf3: error - unable to connect to server: Operation timed out

real	0m51.091s
user	0m0.002s
sys	0m0.006s
```
With @davidBar-On's `TBD__issue-1157-support-address%interface-syntax` branch:
```
# ./iperf3 --client fe80::XXXX%en0
<Test runs as expected>

# ./iperf3 --client fe80::XXXX
iperf3: parameter error - host device name (ip%<dev>) is supported (and required) only for IPv6 link-local address

Usage: ...
```
`Slightly` cleaner exit.

bmah888:
I have a follow-on commit to this PR, which makes the `host%intf` shortcut work similarly for `-c` and `-B`. We also want to support IPv6 link-local addresses in both places. Doing some testing, trying to figure out where's good place to put my revisions for review.

Issue #952, opened 2020-01-10
iperf3 json output unparsable with bidir option set


blaines:
# Context

* Version of iperf3: iperf 3.7 (cJSON 1.5.2)

* Hardware: Darwin Kernel Version 19.2.0

* Operating system (and distribution, if any): macOS 10.15.2 (19C57)

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report

* Expected Behavior

Parsable JSON for bidirectional test with unique keys.

* Actual Behavior

This JSON cannot be processed:
<img width="439" alt="Screen Shot 2020-01-10 at 1 18 55 AM" src="https://user-images.githubusercontent.com/47128/72141581-fab3bf80-3347-11ea-8a48-7903d64f5c28.png">

Note duplicate keys `sum`, `sum_sent`, `sum_received` in the above JSON.

* Steps to Reproduce

`iperf3 -c hostname --bidir -J`

* Possible Solution

Perhaps a client -> server "TX" key, and a server -> client "RX" key?

bmah888:
This is a real bug...someone sent an issue on this awhile back.  I'm trying to find the original instance of this bug being reported.

pmankad96:
I recently ran into this bug. It is doing the right thing by reporting sum_sent and sum_received from client and server perspective. But it is being entered as duplicate entry. 

I was able to fix it by creating a list (sum) of dictionaries (dict1: sum_sent, sum_received, dict2: sum_sent, sum_received) and it worked. 

- Example with a work around:

>>> for key in x['end']:
...     print(key)
...
streams
sum
cpu_utilization_percent
sender_tcp_congestion
receiver_tcp_congestion
>>> type(x['end']['sum'])
<class 'list'>
>>> len(x['end']['sum'])
2
>>> x['end']['sum'][0]
[...]

PR #1200, opened 2021-09-04
Fix for issue #952 - different json object names for bidir reverse ch…

davidBar-On:
Latest master code

* Issues fixed (if any):
#952 (Yest another try to fix it.)

When BIDIR is set, give a different JSON object name to the sum and sent/receive sums.
The different names are given to the data sent from the server, which is the "reverse" channel.  This makes sure that the name reported on the server and client are compatible, and that the names are the same as with non-bdir, except for when reverse mode is used.

With this approach there is no change to JSON output for non-bidir tests.  In addition, and programs that know to read the server and client JSON output for the non-bidir case should be able to work properly also when bidir is specified.

Programs changes are required only if they already analyse the "reverse" channel output for the bidir case, but since currently these JSON outputs has duplicate keys, probably there are very few cases where this is done.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -3246,6 +3246,8 @@ iperf_print_intermediate(struct iperf_test *test)
         double avg_jitter = 0.0, lost_percent;
         int stream_must_be_sender = current_mode * current_mode;
 
+        char *sum_name;
+
         /*  Print stream role just for bidirectional mode. */
 
         if (test->mode == BIDIRECTIONAL) {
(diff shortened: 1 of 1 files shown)


src/iperf_api.c
```diff
@@ -3280,6 +3282,22 @@ iperf_print_intermediate(struct iperf_test *test)
...
+             * the "reverse" channel.  This makes sure that the name reported on the server
+             * and client are compatible, and the names are the same as with non-bdir,
+             * except for when reverse is used.
+             */
+            sum_name = "sum";
+            if (test->mode == BIDIRECTIONAL) {
+                if ((test->role == 'c' && !stream_must_be_sender) ||
+                    (test->role != 'c' && stream_must_be_sender))
+                {
+                    sum_name = "sum_bdir_reverse";
```
bmah888, line 3297: Would it be possible to change this to `sum_bidir_reverse` instead in order to match the spelling of the `--bidir` option?

bmah888:
I'm going to merge this as-is, and then do a follow-up commit to change the identifiers per comments above. Thanks!

PR #1204, opened 2021-09-17
Support zerocopy by Reverse mode

davidBar-On:
Latest master branch

* Issues fixed (if any):
* None

Support `-Z` zerocopy by the server for Reverse mode.
Currently the `-Z` parameter value is not sent to the server, but it seems that this was not done intentionally. The change in this PR works o.k. and significantly improves TCP performance with `-R -Z` (at least for high bandwidth interfaces).

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2163,7 +2163,7 @@ get_parameters(struct iperf_test *test)
 	if ((j_p = cJSON_GetObjectItem(j, "repeating_payload")) != NULL)
 	    test->repeating_payload = 1;
 	if ((j_p = cJSON_GetObjectItem(j, "zerocopy")) != NULL)
-	    test->zerocopy = 1;
+	    test->zerocopy = j_p->valueint;
 #if defined(HAVE_DONT_FRAGMENT)
 	if ((j_p = cJSON_GetObjectItem(j, "dont_fragment")) != NULL)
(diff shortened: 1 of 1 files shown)


src/iperf_api.c
```diff
@@ -2160,6 +2162,8 @@ get_parameters(struct iperf_test *test)
 	    iperf_set_test_udp_counters_64bit(test, 1);
 	if ((j_p = cJSON_GetObjectItem(j, "repeating_payload")) != NULL)
 	    test->repeating_payload = 1;
+	if ((j_p = cJSON_GetObjectItem(j, "zerocopy")) != NULL)
+	    test->zerocopy = 1;
```
bmah888, line 2166: Thanks for the PR! Have not tried running this code yet, but perhaps this should be `test->zerocopy = j_p->valueint`?

bmah888 (commented):
Thanks for the PR! Kinda weird that nobody noticed this before. One possible suggestion on the code...inline.

bmah888:
One feedback item on the PR.

src/iperf_api.c
davidBar-On: Changed to `test->zerocopy = j_p->valueint` as suggested and tested that it works o.k.

bmah888: Great, thanks! I have tested this also and it seems to be working correctly.

bmah888 (approved):

PR #1210, opened 2021-09-24
Issue 1157 take 2

bmah888:

* Issues fixed (if any): #1157 

Builds on @davidBar-On work in PR #1180 to implement functionality described in #1157.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ os:
    - linux
    - osx
    - freebsd
- 
+
 notifications:
   slack:
     secure: ImUmX7hcYotHWCDBfOcIvF6H7kkeGqiaUCy7SVPFtgPbz33ttpbRd94E7oxWVmZMLKb+i6+JCujTEWGwGBimzH+DjL0LLWs0ShzXZIUa1UzEPTc4hgV6VAxucYKFg2WrbXgOPWbulkMG1VZ6pX7GlAEGf0qyNqn44F7S2ay9m18=
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Invoking iperf3
 iperf3 includes a manual page listing all of the command-line options.
 The manual page is the most up-to-date reference to the various flags and parameters.
 
-For sample command line usage, see: 
+For sample command line usage, see:
 
 https://fasterdata.es.net/performance-testing/network-troubleshooting-tools/iperf/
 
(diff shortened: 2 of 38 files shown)


bmah888:
@davidBar-On @TheRealDJ if you could take a look at this and comment, that'd be appreciated!

TheRealDJ:
Code looks good.
Testing Results (masked IPs and Interfaces):
```
./iperf3 --client ${IPERF_SERVER_GLOBAL}
<Test runs as expected>

./iperf3 --client ${IPERF_SERVER_GLOBAL}%${IPERF_SERVER_INT}
p ${IPERF_SERVER_GLOBAL} p1 ${IPERF_SERVER_INT}
<Test runs as expected>

./iperf3 --client ${IPERF_SERVER_LL}%${IPERF_SERVER_INT}
p ${IPERF_SERVER_LL} p1 ${IPERF_SERVER_INT}
IPv6 link local address literal detected
<Test runs as expected>

./iperf3 --client ${IPERF_SERVER_LL}
iperf3: error - unable to connect to server: Invalid argument
```
Looks like the code sets the bind address and interface for `--client GLOBAL%INTERFACE`, but the output doesn't indicate if any binding occurred. Maybe somewhere in the JSON output could be an indication if any specific binding was done (in the case of `--client ADDRESS%DEV`, `--bind ADDRESS%DEV` or `--bind-dev DEV`).

Also, I'd assume that the added iperf_print calls are for testing purposes? They break JSON output.

DJ

TheRealDJ:
I also submitted https://github.com/esnet/iperf/pull/1211 to fix the 2 compile warnings I was getting in the testing process.

davidBar-On:
I sent one comment (including two potential issues). 
There is one difference with iperf2 that I want to point out.  iperf2 is using two different variables for the `device` from `-B` and `-c`.  In addition, setting the device on the client is allowed only by `-c` and on the server only by `-B`.  E.g. see [this warning](https://sourceforge.net/p/iperf2/code/ci/master/tree/src/Settings.cpp#l1657). I didn't find any problem in this PR related code, and I believe that iperf2 uses two separate variables for the device because it doesn't have the `--bind-dev` option. However I am noting this instead I missed something.

swlars (approved):

bmah888:
> I sent one comment (including two potential issues). There is one difference with iperf2 that I want to point out. iperf2 is using two different variables for the `device` from `-B` and `-c`. In addition, setting the device on the client is allowed only by `-c` and on the server only by `-B`. E.g. see [this warning](https://sourceforge.net/p/iperf2/code/ci/master/tree/src/Settings.cpp#l1657). I didn't find any problem in this PR related code, and I believe that iperf2 uses two separate variables for the device because it doesn't have the `--bind-dev` option. However I am noting this instead I missed something.

Good observation @davidBar-On, I think this is basically just differing implementations of the "bind a socket to an interface" functionality between iperf2 and iperf3. Hopefully the addition of the "%" shortcut won't create any confusion, and it would work the same between the two programs. Thanks!

bmah888:
> Also, I'd assume that the added iperf_print calls are for testing purposes? They break JSON output.

@TheRealDJ , yep, those should been (and now are) hidden behind the debug flag. Thanks for catching that!

Have to think a bit about what else we might want to propagate into the JSON, might be a good future enhancement.

PR #1211, opened 2021-09-24
Fixed 2 lines causing warnings in iperf_api.c

TheRealDJ:
Fixed 2 compile warnings

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1574,7 +1574,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
             iperf_size_t file_bytes = st.st_size;
             test->settings->bytes = file_bytes;
             if (test->debug)
-                printf("End condition set to file-size: %d bytes\n", test->settings->bytes);
+                printf("End condition set to file-size: %lu bytes\n", test->settings->bytes);
         }
         // if failing to read file stat, it should fallback to default duration mode
     }
@@ -4306,7 +4306,7 @@ diskfile_recv(struct iperf_stream *sp)
 
     r = sp->rcv2(sp);
     if (r > 0) {
-	(void) write(sp->diskfile_fd, sp->buffer, r);
+	(void) (write(sp->diskfile_fd, sp->buffer, r) + 1);
     }
     return r;
 }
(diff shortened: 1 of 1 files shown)


src/iperf_api.c
```diff
@@ -1574,7 +1574,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
             iperf_size_t file_bytes = st.st_size;
             test->settings->bytes = file_bytes;
             if (test->debug)
-                printf("End condition set to file-size: %d bytes\n", test->settings->bytes);
+                printf("End condition set to file-size: %lu bytes\n", test->settings->bytes);
```
zuckschwerdt, line 1577: actually that needs to be `"%" PRIu64` or `%llu` with as cast to `long long unsigned`

TheRealDJ: I tried `%llu` ... My compiler says it should be `%lu` ... Indicates type is `long unsigned int` not `long long unsigned int`
```
iperf_api.c: In function ‘iperf_parse_arguments’:
iperf_api.c:1651:60: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘iperf_size_t’ {aka ‘long unsigned int’} [-Wformat=]
 1651 |                 printf("End condition set to file-size: %llu bytes\n", test->settings->bytes);
      |                                                         ~~~^           ~~~~~~~~~~~~~~~~~~~~~
      |                                                            |                         |
      |                                                            long long unsigned int    iperf_size_t {aka long unsigned int}
      |                                                         %lu
```

TheRealDJ: How about
```
#ifdef __ILP32__
                printf("End condition set to file-size: %llu bytes\n", test->settings->bytes);
#elif __LP64__
                printf("End condition set to file-size: %lu bytes\n", test->settings->bytes);
#endif
```
to remove warnings from both? Not sure if this would work with compilers other than gcc though.

zuckschwerdt: About the warning you get on `%llu`: did you forget to add the cast?
But really `"%" PRIu64` is terse and standard C99, IMO preferred. (that's `...file-size: %"PRIu64" bytes...`)

TheRealDJ: Updated, and I confirmed no compilation warnings or errors. I was having issues in my test code getting this to work ...  because I was missing `#include <inttypes.h>`.

swlars:
Thanks for the pull request!

PR #1220, opened 2021-10-16
Move sndbuf size warning to stderr

jshen28:
When do perf test with -J, sndbuf warning
is also printed to stdout. Move it to stderr..

Signed-off-by: ushen <yshxxsjt715@gmail.com>

### DIFF
--- a/src/iperf_tcp.c
+++ b/src/iperf_tcp.c
@@ -325,7 +325,7 @@ iperf_tcp_listen(struct iperf_test *test)
 	return -1;
     }
     if (test->debug) {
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the PR! This message only appears when running with the `--debug` flag. We actually have a specific warning about trying to use both debug mode and JSON output, and that this might not work well:
```
warning: Debug output (-d) may interfere with JSON output (-J)
```
Note that there are many cases of debug output going to stdout, other than this particular message. It might be better to send all debug output to stderr (much like we do with `warning()`) but making that change is much more involved. I'm going to close this PR while we figure out what to do.

PR #1251, opened 2021-12-17
Force iperf3 server to exit if it can't open its log file.

### LINKED ISSUE #1225: When logfile is not created, iperf loop indefinitely
# Context
* Version of iperf3:
  iperf3-3.10.1
  kernel-5.14.0

  Fedora Rawhide

# Bug Report
Hi,
when running iperf3 with --logfile and using a path that does not exist, iperf3 won't stop with an error but will actually just loop in the run() function. 

* Expected Behavior
  Stop with an error

* Actual Behavior
  Loops indefinitely

* Steps to Reproduce
  $ iperf3 -s 127.0.0.1 --logfile /this/path/is/nonsense/iperf3-log --forceflush
  iperf3: error - unable to open log file: No such file or directory
  iperf3: error - unable to open log file: No such file or directory
  iperf3: error - unable to open log file: No such file or directory
  iperf3: error - unable to open log file: No such file or directory
  iperf3: error - unable to open log file: No such file or directory
  iperf3: error - unable to open log file: No such file or directory
  iperf3: error - unable to open log file: No such file or directory
  iperf3: error - unable to open log file: No such file or directory
  ...

* Possible Solution
  I think there are two options here:
  a. either the run() function under the daemon part in main.c should test for negative rc like this -  (rc <= -1) on line 155
  or
  b. if the (rc < -1) is intentional to filter out errors where function returns -1 and iperf3 should keep going, 
      but stop when function returns anything lower, the function iperf_open_logfile in iperf_api.c should 
      probably return -2

# Additional question
I ran into this when trying to figure out why the default location for the logfile changed when you don't use absolute path. Since the introduction of https://github.com/esnet/iperf/pull/871 when using --logfile some_file, the file is created in root directory as /some_file. Before this PR, the file was created in the directory where you run the iperf3 command. I was wondering if it would make sense to actually create the logfile before running the daemon() and thus before changing directory to '/' ? Maybe keep the iperf_open_logfile function but call it sooner rather than after the context has changed?

Just a thought, thanks for any insight on this.

Regards,
Michal Ruprich

bmah888:
Fixes #1225.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -437,7 +437,7 @@ iperf_run_server(struct iperf_test *test)
 
     if (test->logfile)
         if (iperf_open_logfile(test) < 0)
-            return -1;
+            return -2;
 
     if (test->affinity != -1)
 	if (iperf_setaffinity(test, test->affinity) != 0)


src/iperf_server_api.c
```diff
@@ -437,7 +437,7 @@ iperf_run_server(struct iperf_test *test)
 
     if (test->logfile)
         if (iperf_open_logfile(test) < 0)
-            return -1;
+            return -2;
```
swlars, line 441: Looks good! Only thing I might mention is changing -2 to be some kind of error code in a define or enum, but completely optional!

swlars (approved):
Looks good! Only thing I might mention is changing -2 to be some kind of error code in a define or enum, but completely optional!

bmah888:
Thanks for the review! There are probably some other places in the where we could / should turn the `-1` and `-2` into preprocessor definitions or constants.

Issue #1249, opened 2021-12-17
Specifying client port (--cport) fails in bi-directional mode (--bidir)


Forst:
# Context

* **Version of iperf3:** 3.9 (deb packages from iperf.fr)
* **Operating system:** Ubuntu 18.04.6, running in LXC; kernel 5.4.0-90-generic

# Bug Report

When trying to specify a fixed client port (`--cport` option) together with bi-directional mode (`--bidir`), the test fails to start with "unable to connect stream: Address already in use", as the client is unable to bind a second connection to the same client port.

Looks like `--bidir` relies on the client port to be ephemeral, and fails when it's explicitly specified.

## Steps to Reproduce

1. Start a server:
    ```console
    user@server:~$ iperf3 -s -p 10000
    ```
2. Try to start a bi-directional client with a fixed source port:
    ```console
    user@client:~$ iperf3 -c 192.168.1.100 -p 10000 --cport 15000 --bidir
    Connecting to host 192.168.1.100, port 10000
    iperf3: error - unable to connect stream: Address already in use
    ```

## Expected Behavior

Performance test happens in both directions obeying the fixed client port.

## Actual Behavior

Performance test fails to start, as it is unable to bind a second stream to the same client port.

## Workaround

Running two instances of `iperf3` on the server side, and two instances on the client side, each with their own direction and client/server ports.

## Possible Solution

* Use a single stream for both directions, instead of establishing two streams
* Allow specifying a client port for each direction `--cport 15000:15001`, or a range of ports iperf3 can pick from `--cport 15000-15999`

bmah888:
I can reproduce this problem. That said I'm not familiar with how TCP ports are allocated / managed in this scenario, mostly because I didn't implement `--bidir`. This will need a little investigation to see exactly what's going on.

(Please note that iperf.fr has no connection to ESnet or the iperf3 development team. The binaries on iperf.fr aren't of current versions of iperf3, and the maintainers have not responded to our attempts to communicate.)

davidBar-On:
Submitted PR #1259 with a proposed fix for the problem.

PR #1259, opened 2022-01-08
Use different Up/Downlink ports when cport and bdir options are set

davidBar-On:
Latest 3.10+

* Issues fixed (if any):
* #1249

When both `--cport` and `--bdir` options are set, make sure that the ports used for one direction are different from the ports used for the other direction.


bmah888:
Thanks for the PR! I'm still exercising the change, but so far so good.

swlars (approved):
Looks good!

Issue #1212, opened 2021-09-27
Extremely high Jitter value reported by iPerf 3.10.1


d3c1978:
# Context

* Version of iperf3: 3.10.1

* Hardware:
  Samsung Galaxy S21 (ARM64) (Client), Intel Xeon 64-bit (Server)

* Operating system (and distribution, if any):
  Android 11 (Client), Ubuntu 18.04 Server (Server)

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):
  iperf3 cross compiled for Android and Ubuntu servers using standard compilers (toolchain provided by Android for client and standard GCC for server)

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report
   Running a UDP Reverse Stream occassionally produces a very high Jitter value which is much longer than the test itself. This does not happen all the time, but about 8% of the runs.
   On first thought, this seemed to be related to bug #

* Expected Behavior
  Jitter value is realistic in all cases

* Actual Behavior
  The initial jitter (and sometimes after the initial interval) shows up as being much longer (occassionally we saw the value representing over 24 hours).

`iperf3 -c iperfserver -R -u -P 1 -b 6M --port 6366
Connecting to host iperfserver, port 6366
Reverse mode, remote host iperfserver is sending
[  5] local 2222:ffff:xxxx:yyyy:aaaa:bbbb:cccc:dddd port 37242 connected to 2222:1111:aaa:bbbb:cccc:dddd:eeee:ffff port 6366
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5]   0.00-1.00   sec   228 KBytes  1.87 Mbits/sec  3408823.110 ms  527590/527761 (1e+02
[  5]   1.00-2.00   sec   365 KBytes  2.99 Mbits/sec  6.670 ms  -273/0 (0
[  5]   2.00-3.00   sec   446 KBytes  3.66 Mbits/sec  3.619 ms  -334/0 (0
[  5]   3.00-4.00   sec   310 KBytes  2.54 Mbits/sec  4.861 ms  -232/0 (0
[...]

Issue #1182, opened 2021-07-08
Very strange jitter value 


raszuk:
Hi, 

When I run Iperf3 UDP testing (in any direction) to go via specific path I am seeing the below output:

root@fuji:~# iperf3 -s
-----------------------------------------------------------
Server listening on 5201
-----------------------------------------------------------
Accepted connection from 10.160.16.54, port 64065
[  5] local 10.133.5.132 port 5201 connected to 10.160.16.54 port 52287
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5]   0.00-1.00   sec   104 KBytes   852 Kbits/sec  46822443975.381 ms  0/13 (0%)
[  5]   1.00-2.00   sec   120 KBytes   983 Kbits/sec  17783745093.089 ms  0/15 (0%)
[  5]   2.00-3.00   sec   128 KBytes  1.05 Mbits/sec  6332331570.581 ms  0/16 (0%)
[  5]   3.00-4.00   sec   128 KBytes  1.05 Mbits/sec  2254779458.383 ms  0/16 (0%)
[  5]   4.00-5.00   sec   128 KBytes  1.05 Mbits/sec  802868635.583 ms  0/16 (0%)
[  5]   5.00-6.00   sec   128 KBytes  1.05 Mbits/sec  285880752.080 ms  0/16 (0%)
[  5]   6.00-7.00   sec   128 KBytes  1.05 Mbits/sec  101794740.820 ms  0/16 (0%)
[  5]   7.00-8.00   sec   128 KBytes  1.05 Mbits/sec  36246474.556 ms  0/16 (0%)
[  5]   8.00-9.00   sec   128 KBytes  1.05 Mbits/sec  12906432.687 ms  0/16 (0%)
[  5]   9.00-10.00  sec   128 KBytes  1.05 Mbits/sec  4595647.335 ms  0/16 (0%)
[  5]  10.00-11.00  sec   128 KBytes  1.05 Mbits/sec  1636391.733 ms  0/16 (0%)
[  5]  11.00-12.00  sec   128 KBytes  1.05 Mbits/sec  582677.244 ms  0/16 (0%)
[  5]  12.00-13.00  sec   128 KBytes  1.05 Mbits/sec  207476.898 ms  0/16 (0%)
[  5]  13.00-14.00  sec   128 KBytes  1.05 Mbits/sec  73878.125 ms  0/16 (0%)
[...]

Issue #1123, opened 2021-02-12
With --udp and --reverse, out-of-order packet reception corrupts sequence numbers / test


aahoward:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the iperf3 mailing list at iperf-dev@googlegroups.com or
question sites such as Stack Overflow
(http://www.stackoverflow.com/).  A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3: 3.9

* Hardware: X86

* Operating system (and distribution, if any): Linux Ubuntu [ multiple ]

# Bug Report
A WAN connection between our datacentre and  AWS exhibits high packet out of order characteristics. When using iperf3 UDP reverse test mode with client on prem and server in AWS we observe corruption of packet counters. The following can be used to reproduce the problem on demand within our network.

```
iperf39 -u -c aws_ec2_ip -l 100 -t 1 -b 10k -R

Note. this will only error when out of order packet delivery is common
```

A test script has been included to automate the reproduction of this bug, In addition a PCAP and JSON output is included that show an incident of the bug.

The bug is caused by frame1 [4 bytes server response to client ] and frame2 [ first test payload ] being recieved in reverse order.  

The client reads 4 bytes from the first frame recieved [ frame2 ] , leaving 96 bytes remaining , then read's 100 bytes [ 96+4 ]. This results in a framing error when the counters are memcpy from the buffer.

Introducing a sleep in iperf_udp.c after the 4 bytes response has been transmitted will allow sufficient time to ensure that the 4 byte response and the first test load are never received out of sequence. A 10ms sleep should be more than enough to cater for most network conditions with respect to out of order packets. 

This bug is only applicable to out of order in the context of the server 4 byte "ping/pong" response and the first testing payload.

* Expected Behavior
4 byte response "987654321" response should always arrive before first test payload packet
* Actual Behavior
Testing AWS to onpremise network links show high amounts of out of order frames. On multiple occasions the 4 byte response from the server to the client arrives after the first payload packet. This creates framing issues.
* Steps to Reproduce
Script/JSON results and PCAP have been attached that show the problem
* Possible Solution
[...]

PR #1260, opened 2022-01-08
Ignore out of order packts during UDP connection in Reverse Mode

davidBar-On:

Latest 3.10+

* Issues fixed (if any):

#1123, #1212 and probably #914, #1182

Based on the analysis done by @aahoward in #1123.  In UDP Reverse mode, the server starts sending the UDP packets immediately after sending the Connect response.  Sometimes, before network packets reordering, the first UDP packet arrives to the client before the Connection response.  In this case the client wrongly handles the first packet as the connection response.

The proposed solution is that the client will wait until receiving the connect response.  Current implementation is that the client allows one packet out of order (hardcoded as it seems that the chance that two packets will arrive before the connection response is minimal).  It is also assumed that the first packets does not include the 4 bytes connection response static value in a 4 bytes alignment offset.

Note that based on a suggestion in a previous issue (I couldn't find it now) I changed the connection and response 4 bytes messages values to textual value, to allow easier Wireshark, etc. readability.

### DIFF
--- a/src/iperf_udp.c
+++ b/src/iperf_udp.c
@@ -365,7 +365,7 @@ int
 iperf_udp_accept(struct iperf_test *test)
 {
     struct sockaddr_storage sa_peer;
-    int       buf;
+    unsigned int buf;
     socklen_t len;
     int       sz, s;
     int	      rc;
@@ -452,7 +452,7 @@ iperf_udp_accept(struct iperf_test *test)
     test->max_fd = (test->max_fd < test->prot_listener) ? test->prot_listener : test->max_fd;
 
     /* Let the client know we're ready "accept" another UDP "stream" */
-    buf = 987654321;		/* any content will work here */
+    buf = UDP_CONNECT_REPLY;
     if (write(s, &buf, sizeof(buf)) < 0) {
         i_errno = IESTREAMWRITE;
         return -1;
@@ -494,11 +494,13 @@ iperf_udp_listen(struct iperf_test *test)
 int
 iperf_udp_connect(struct iperf_test *test)
 {
-    int s, buf, sz;
+    int s, sz;
+    unsigned int buf;
 #ifdef SO_RCVTIMEO
     struct timeval tv;
 #endif
     int rc;
+    int i, max_len_wait_for_reply;
 
     /* Create and bind our local socket. */
     if ((s = netdial(test->settings->domain, Pudp, test->bind_address, test->bind_dev, test->bind_port, test->server_hostname, test->server_port, -1)) < 0) {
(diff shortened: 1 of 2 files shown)


src/iperf_udp.c
```diff
@@ -564,17 +566,35 @@ iperf_udp_connect(struct iperf_test *test)
...
+            i_errno = IESTREAMREAD;
+            return -1;
+        }
+        if (test->debug) {
+            printf("Connect received for Socket %d, sz=%d, buf=%x, i=%d, max_len_wait_for_reply=%d\n", s, sz, buf, i, max_len_wait_for_reply);
+        }
+        i += sz;
+    } while (buf != UDP_CONNECT_REPLY && i < max_len_wait_for_reply);
+
+    if (buf != UDP_CONNECT_REPLY) {
```
bmah888, line 597: Same thing here...suggesting to add `&& buf != 987654321` to the condition in this line.

bmah888 (changes requested):
Thanks for the PR! I did a little quick testing, and it seems good so far, aside from the two comments I had about backward compatibility. I'm going to try to study the various issues this PR is supposed to fix to be sure I understand them.

src/iperf_udp.c
```diff
@@ -564,17 +566,35 @@ iperf_udp_connect(struct iperf_test *test)
...
+    do {
+        if ((sz = recv(s, &buf, sizeof(buf), 0)) < 0) {
+            i_errno = IESTREAMREAD;
+            return -1;
+        }
+        if (test->debug) {
+            printf("Connect received for Socket %d, sz=%d, buf=%x, i=%d, max_len_wait_for_reply=%d\n", s, sz, buf, i, max_len_wait_for_reply);
+        }
+        i += sz;
+    } while (buf != UDP_CONNECT_REPLY && i < max_len_wait_for_reply);
```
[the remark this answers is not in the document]
davidBar-On: Done.  Forgot about the backward compatibility ...

Note that there is a minor potential issue with checking for both the old and new connect reply values.  Potentially, the first packet that is received instead of the connect reply (because of the out of order) can contain the expected reply value on a 4 bytes boundary.  Checking for both values doubles the chance that that will happen.  On the other hand, practically that can probably happen only when sending a file contents.  Since out of order is rare, and this only about the first packet, checking for both values is probably o.k.  If not, then only the legacy value should be used.

src/iperf_udp.c
davidBar-On: Done

bmah888 (approved):

bmah888:
Looks great, thanks again for working on this!

Issue #1215, opened 2021-10-04
--idle-timeout does not work as expected


zokl:
* Version of iperf3: 3.10.1+ (28.9.2021)
* Hardware: x86_64
* Operating system (and distribution, if any): Linux

I am trying to measure a mobile network, where a connection failure may occur during the measurement, during which the IP address of the measurement client is changed. According to the configuration below, I expect that after 30 seconds, the iperf3 application will terminate on the client-side, and 40 seconds the application will terminate on the server-side. 

Expected behavior - Client-side killed by rcv-timeout:
```
root@FTWireless-59906C:~# iperf3 -c 1.2.3.4 -i 1 -b 1M -t 120 --rcv-timeout 30000 -R
Connecting to host 1.2.3.4, port 5201
Reverse mode, remote host 1.2.3.4 is sending
[  5] local 100.112.64.108 port 58414 connected to 1.2.3.4 port 5201
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec   128 KBytes  1.05 Mbits/sec
[  5]   1.00-2.00   sec   128 KBytes  1.05 Mbits/sec
[  5]   2.00-3.00   sec   128 KBytes  1.05 Mbits/sec
[  5]   3.00-4.00   sec   128 KBytes  1.05 Mbits/sec
[  5]   4.00-5.00   sec   128 KBytes  1.05 Mbits/sec
[  5]   5.00-6.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   6.00-7.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   7.00-8.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   8.00-9.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   9.00-10.00  sec  0.00 Bytes  0.00 bits/sec
[  5]  10.00-11.00  sec  0.00 Bytes  0.00 bits/sec
[  5]  11.00-12.00  sec  0.00 Bytes  0.00 bits/sec
[  5]  12.00-13.00  sec  0.00 Bytes  0.00 bits/sec
[  5]  13.00-14.00  sec  0.00 Bytes  0.00 bits/sec
[  5]  14.00-15.00  sec  0.00 Bytes  0.00 bits/sec
[  5]  15.00-16.00  sec  0.00 Bytes  0.00 bits/sec
[  5]  16.00-17.00  sec  0.00 Bytes  0.00 bits/sec
[...]

PR #1282, opened 2022-02-19
Fix for issue 1215 - add TCP  option

davidBar-On:
3.11

* Issues fixed (if any):
#1215

Suggested fix for issue #1215 - adding `snd-timeout` option.  Used by a sender to set TCP `TCP_USER_TIMEOUT ` socket option, so if TCP send for that time, session is terminated.  Applied also for the Control Socket.  (Note: I tried to find existing iperf3 option that can be used, instead of adding a new option, but there doesn't seem to be such option.)

Useful to reduce TCP timeout in cases there is a problem with the interface to the receiver (such as cell phone IP address change as in #1215).

Not sure how to resolve the conflic in `configure`.  The file was changed to add the `HAVE_TCP_USER_TIMEOUT`, but in WSL Ubunto I use, `autoconf 2.69` is available, not 2.71.


bmah888:
Thanks for the pull request! I'll give it a look.

About the `configure` part...don't worry about including changes to `configure` in the pull request. We can regen that with autoconf-2.71 on either macOS or FreeBSD after it's been merged. I was looking for changes to `configure.ac`, which is the source for `configure`, but didn't see that file among the listed files in the pull request?

We probably need to do a manual page update also.

davidBar-On:
Added the modified `configure.ac` (was not included originally by mistake).

bmah888:
Thanks for the PR! I'm going to merge this as it stands now, then do some follow-up commits for regenerating the configure scripts, etc. and some manpage, etc. updates.

zokl:
Thank you very much for your PR.

PR #1316, opened 2022-04-11
suppress -Wunused-variable when HAVE_SENDFILE is undefined

mniestroj:
Suppress following compiler warning when there is no HAVE_SENDFILE
support:

```
  /lib/iperf3/src/net.c: In function 'Nsendfile':
  /lib/iperf3/src/net.c:449:11: warning: unused variable 'offset' [-Wunused-variable]
    449 |     off_t offset;
        |           ^~~~~~
```


swlars:
Thank you for the pull request!

PR #1318, opened 2022-04-13
Added option -U to set a custom test UUID on the client

puchol:
The option allows setting a 36 character UUID as part of the client's test settings, which can then be captured server-side and used to correlate with tests by other tools. As an example, I have written a script which generates a UUID, then runs fping commands, and sends them to a log storage server. That same server can now receive an iperf3 test result, which is correlated by UUID with the fping test.


bmah888:
Thanks for the PR! The `--extra-data` option was designed to solve something like this, where a client needs to send a more-or-less arbitrary string to the server along with a test. Do you need something fundamentally different from this?

puchol:
I was not aware of that flag, it is not mentioned in the documentation at https://iperf.fr/iperf-doc.php. That flag should be fine for my purposes, I will test it, thank you!

bmah888:
Thanks for the follow-up! I'll close this pull request seeing as existing mechanisms meet your needs.

Unfortunately, iperf.fr is not the official iperf3 Web site. https://software.es.net/iperf has the official project Web pages. Apologies for the confusion, we have no control over the content on iperf.fr (and in fact have been trying for years to contact the maintainer of that site).

puchol:
OK thank you again for your attention to this PR, I was not aware about iperf.fr, you learn something new every day!

davidBar-On:
Windows executables of the latest iperf3 versions are maintained [here](https://files.budman.pw/) by [BudMan](https://www.neowin.net/forum/tags/iperf3/) (I don't know who he is).

PR #1325, opened 2022-04-27
configure: support riscv via newer autotools

zx2c4:
Currently this can't easily be cross-compiled for RISC-V because it uses
an ancient autotools. I ran `autoreconf -fi` to generate this commit on
a system with a newer version of autotools, which makes it buildable for
RISC-V.

Before this change, the following happens:

```
checking host system type... Invalid configuration `riscv64-linux-musl': machine `riscv64' not recognized
configure: error: /bin/sh ./config/config.sub riscv64-linux-musl failed
```

After this change, all goes well.

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -1429,10 +1429,10 @@ s390*-*linux*|s390*-*tpf*|sparc*-*linux*)
 	  x86_64-*linux*)
 	    LD="${LD-ld} -m elf_x86_64"
 	    ;;
-	  powerpcle-*linux*)
+	  powerpcle-*linux*|powerpc64le-*linux*)
 	    LD="${LD-ld} -m elf64lppc"
 	    ;;
-	  powerpc-*linux*)
+	  powerpc-*linux*|powerpc64-*linux*)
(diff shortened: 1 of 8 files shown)


bmah888:
Thanks for the pull request! What's a not ancient version of autotools that's suitable for supporting the RISC-V architecture? Ideally that'd be a part of our development environment for iperf3...that's more sustainable than committing the byproducts of you regenerating stuff.

zx2c4:
Looks like it was added in https://github.com/autotools-mirror/autoconf/commit/2ca0d5755f68ea7a99c39a236ea8d02aa207d594, so quite old. Your bootstrap.sh command isn't copying in the new files (-fi flags to autoreconf), so maybe if you changed that you'd get this naturally from the autotools you already have.

bmah888:
OK, I think I understand what you mean. So maybe `bootstrap.sh` needs to invoke `autoconf -f` or `autoreconfig -fi`. Will give this a try. Thanks!

bmah888:
Thanks again for the bug report...master should be good now with #1335 plus the subsequent regeneration of files.

PR #1327, opened 2022-04-29
Add debug level option to --debug

davidBar-On:
master

* Issues fixed (if any):
* None

Add "debug level" option to `--debug` by setting `--debug=<level>`.  This is to allow setting log level to the debug messages, so not all debug messages may be emitted.  4 levels of debug levels are supported (1-4).  If the option is not set than all the debug messages are emitted. (Note that the option does not work for `-d` as `getopt_long()` does not support optional argument to a single digit option.)

The purpose of this change is to allow using the debug mode with UDP, but without the debug messages that are printed per packet.  Therefore, I added a check for the debug level for these debug messages.


swlars:
Thank you for the merge request!

Issue #1314, opened 2022-04-08
--bidir option randomly fails with "iperf3: error - unable to connect stream: No such file or directory" on Windows


madbrain76:
# Context

I'm using the --bidir feature of iperf3 to measure perform between multiple LAN devices on a variety of operating systems.
When using Windows as a client, the option works erratically, often display the message "iperf3: error - unable to connect stream: No such file or directory", but not every time.

* Version of iperf3: 3.10.3

* Hardware: AMD 5950X, 32GB, Aquantia AQN-107 NIC (client)
Server is a variety of Unix platforms. The one in this test was a Raspberry Pi with Realtek 2.5 Gbps USB NIC.

* Operating system (and distribution, if any): Windows 10 21H2 (client side) . Raspberry Pi 64-bit Bullseye OS (server side).

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

I am aware support can't be provided for Windows, but I still think it's worth filing and tracking the issue here. I will attempt to fix the bug myself, if no one else does.

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

I used cygwin for the build. It was a while back and I don't recall which compiler was used. I didn't succeed in building debug binaries, which is why there is no pull request attached.

# Bug Report

```
C:\Users\Julien Pierre\Desktop\iperf3>iperf3 -version
iperf 3.10.1 (cJSON 1.7.13)
CYGWIN_NT-10.0-19044 HIGGS 3.3.3-341.x86_64 2021-12-03 16:35 UTC x86_64
Optional features available: CPU affinity setting, support IPv4 don't fragment

C:\Users\Julien Pierre\Desktop\iperf3>iperf3 -c pi64 --bidir
Connecting to host pi64, port 5201
iperf3: error - unable to connect stream: No such file or directory

C:\Users\Julien Pierre\Desktop\iperf3>iperf3 -c pi64 --bidir
Connecting to host pi64, port 5201
[  5] local 2601:646:8801:9d00:445a:8f46:3dfe:c9f6 port 55100 connected to 2601:646:8801:9d00:53c0:7267:b327:e9c port 5201
[  7] local 2601:646:8801:9d00:445a:8f46:3dfe:c9f6 port 55101 connected to 2601:646:8801:9d00:53c0:7267:b327:e9c port 5201
[...]

Issue #1201, opened 2021-09-05
iperf3: error - unable to create a new stream: No such file or directory


lianghdx:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the iperf3 mailing list at iperf-dev@googlegroups.com or
question sites such as Stack Overflow
(http://www.stackoverflow.com/).  A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3: ver3.1.3

* Hardware: cpu: i5-10500,  memory: samsung 8G*2

* Operating system (and distribution, if any): Windows 10 Enterprise LTSC 1809

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report
    First, I unzipped the zip file into a folder. Then I ran the program in the folder with the command: iperf3 -c 192.168.0.101 -i 1 -t 60 -p 8005. For the record, I had started the server with the ip 192.168.0.101 and the port was 8005. The server started successfully.  However, the program running on the client raised the error which made me feel confused: "iperf3: error - unable to create a new stream: No such file or directory". I have never set any limitation on reading or creating a new folder or file while the error showed something prevent the program from reading the file. 
* Expected Behavior
The client started the test.
* Actual Behavior
The client raised an error "iperf3: error - unable to create a new stream: No such file or directory".
* Steps to Reproduce

* Possible Solution

_Please submit patches or code changes as a pull request._

# Enhancement Request

* Current behavior

* Desired behavior

* Implementation notes

_If submitting a proposed implementation of an enhancement request,
please use the pull request mechanism._

davidBar-On:
It may be that the route from the client's computer to 192.168.0.101 is blocked. Try `ping` to see if this is the case. If ping works, try with different ports, as it may be that 8005 is blocked.

[...]

PR #1346, opened 2022-05-30
Wrap getaddrinfo to try more than once on Windows.

madbrain76:

```
```

* Issues fixed (if any):

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -115,6 +115,28 @@ timeout_connect(int s, const struct sockaddr *name, socklen_t namelen,
 	return (ret);
 }
 
+#ifdef __CYGWIN__
+
+const unsigned int MAX_GETADDRINFO_RETRIES = 10;
+/* getaddrinfo may not be completely reliable on Windows hosts. Try more than once. */
+int iperf_getaddrinfo(const char *restrict node,
+                       const char *restrict service,
+                       const struct addrinfo *restrict hints,
+                       struct addrinfo **restrict res)
+{
+    int retvalue = -1;
+    int i = 0;
+    while (i++ < MAX_GETADDRINFO_RETRIES) {
+        retvalue = getaddrinfo(node, service, hints, res);
+        if (0 == retvalue) break;
+        sleep(1); /* yield thread */
+        fprintf(stderr, "getadddrinfo failed. If iperf still ran successfully, this may be a Windows bug. Please see https://github.com/esnet/iperf/issues/1314 . \n");
+    }
+    return retvalue;
+}
+
+#endif
+
 /* netdial and netannouce code comes from libtask: http://swtch.com/libtask/
  * Copyright: http://swtch.com/libtask/COPYRIGHT
 */
@@ -131,7 +153,7 @@ create_socket(int domain, int proto, const char *local, const char *bind_dev, in
         memset(&hints, 0, sizeof(hints));
         hints.ai_family = domain;
         hints.ai_socktype = proto;
-        if ((gerror = getaddrinfo(local, NULL, &hints, &local_res)) != 0)
+        if ((gerror = iperf_getaddrinfo(local, NULL, &hints, &local_res)) != 0)
             return -1;
     }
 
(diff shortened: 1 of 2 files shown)


src/net.c
```diff
@@ -115,6 +115,28 @@ timeout_connect(int s, const struct sockaddr *name, socklen_t namelen,
...
+                       const struct addrinfo *restrict hints,
+                       struct addrinfo **restrict res)
+{
+    int retvalue = -1;
+    int i = 0;
+    while (i++ < MAX_GETADDRINFO_RETRIES) {
+        retvalue = getaddrinfo(node, service, hints, res);
+        if (0 == retvalue) break;
+        sleep(1); /* yield thread */
+        fprintf(stderr, "getadddrinfo failed. If iperf still ran successfully, this may be a Windows bug. Please see https://github.com/esnet/iperf/issues/1314 . \n");
```
davidBar-On, line 133: Error should be either printer only if debug option was set or using `iperf_err()` (to properly handle the JSON output case).  **Suggest to leave the decision to the iperf3 maintainers** (before merging the change).

In both cases, `test` should be passed as a parameter to the function and `iperf.h` should be included.  In case `iperf_err` is used, also `iperf_api.h` should be included

If printed as debug message, suggest to use the newly added debug level option:
```
if (test->debug_level >= DEBUG_LEVEL_ERROR))
    fprintf(...);
```

davidBar-On:
@bmah888, I believe this PR should be merged into the mainline, as it seems to be important for Windows users.  It fixes issue #1314 and probably also #1201.

Although I did not see similar `getaddrifo` issues for non Windows OSs, the PR may be considered for all all OSs.  The root cause of the problem is not clear, and it may be related only to Cigwin and/or Windows.  However, since the change is just retrying few time `getaddrinfo` in case it fails, the only harm in case of permanent `getaddrinfo` problem is that it will take more time for the client to fail (and the retries count may be smaller than 10).

src/net.c
madbrain76: David, thanks for the feedback. Obviously, I wrote my PR before the option was added. I would prefer for this message to be printed unconditionally if getaddrinfo fails intermittently, regardless of the new debug option. The rationale is that with the loop in place, a fatal error would not be seen anymore. And without being aware of the intermittent failure behind the scene, nobody would think to add -d. The intermittent failure seems rare enough that I wanted anyone with a affected system to see the message. Requiring -d would significantly decrease the odds anyone would ever know about the problem. The wording of the message could be changed, though.

I checked the source to see where debug_level was being set. The iperf_test structure that contains debug_level isn't available in the scope of most of the calls to getaddrinfo . A pointer to it would need to be propagated across the stack. getaddrinfo itself of course, being a low-level call, doesn't take this object as an argument. I am reluctant to change the prototype for getaddrinfo . I could make it a wrapper on all platforms, instead of having the platform #ifdef, though.

iperf_test appears to contain a global configuration state. And there is only ever a single instance, created in main() with a call to iperf_new_test. As long as that remains the case, it strikes me as a bit of busy work to pass this object around the stack down everywhere, not just to iperf_getaddrinfo.  I don't want to address that in this PR.

davidBar-On: > I would prefer for this message to be printed unconditionally if getaddrinfo fails intermittently, regardless of the new debug option.

The new debug option is just the ability to set a debug level.  Using `if (test->debug)` always existed, but it still needs the use of `iperf_test`.

> The rationale is that with the loop in place, a fatal error would not be seen anymore. And without being aware of the intermittent failure behind the scene, nobody would think to add -d.

I agree.  In this case still `iperf_err()` should be used to log the error, to properly handle JSON output (in case `-J` was set), which also needs the use of  `iperf_test`.

> A pointer to it would need to be propagated across the stack. getaddrinfo itself of course, being a low-level call, doesn't take this object as an argument. I am reluctant to change the prototype for getaddrinfo . .....

I agree that adding `iperf_test` to `getaddrinfo` and to net.c in general is a problem.  I therefore suggest to leave the decision to the iperf3 maintainers whether to propagate `iperf_test` down to net.c, not log the error in `getaddrinfo`, add a wrapper or any of the other approaches you suggested.

bmah888:
@davidBar-On @madbrain76 : Sorry for the delay! I understand this PR seems to fix an actual, observed problem when running iperf3 under Cygwin/Windows.

Nevertheless, this isn't a platform that we officially support, and I'm reluctant to commit what's basically a hack around an OS bug (it seems) and have that code change affecting the behavior of our officially-supported platforms. I'm going to claim that if this was an issue on Linux, FreeBSD, or macOS, we likely would have encountered this already.

I'm not inclined to take this patch in its current form, however I might be willing to take a PR in this area if it involved a conditional compilation, such that the code change only applies to Windows and has no effect on our supported systems. How's that sound?

Issue #1363, opened 2022-07-24
--blockcount or --bytes kills reverse mode


michellabbe:
# Context
When the server is hit by a very small packet (I use 1 packet tests as healthcheck in my docker image to keep performance impact lowest on actual tests), then tests in reverse mode always return around 0 Mbits/sec until the server is restarted.

* Version of iperf3:
3.11 (ubuntu 22.04, compiled from sources)
3.11-r0 (from alpine 3.16.1 apk)
3.10.1-r0 (from alpine 3.14.1 apk)
note: 3.9-r1-hc (from alpine 3.13 apk) and earlier is **not** affected by this issue

* Hardware:
intel amd64

# Bug Report

* Expected Behavior
Small tests shouldn't tilt the server.

* Actual Behavior
After receiving a small test, all tests in reverse mode report around 0 bits/sec until the server is restarted:

```
Reverse mode, remote host linux.server.lan is sending
[  5] local 192.168.0.10 port 61075 connected to 192.168.0.33 port 5201
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec   128 KBytes  1.04 Mbits/sec
[  5]   1.00-2.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   2.00-3.01   sec  0.00 Bytes  0.00 bits/sec
[  5]   3.01-4.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   4.00-5.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   5.00-6.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   6.00-7.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   7.00-8.01   sec  0.00 Bytes  0.00 bits/sec
[  5]   8.01-9.02   sec  0.00 Bytes  0.00 bits/sec
[  5]   9.02-10.01  sec  0.00 Bytes  0.00 bits/sec
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.01  sec   128 KBytes   105 Kbits/sec    0             sender
[  5]   0.00-10.01  sec   128 KBytes   105 Kbits/sec                  receiver
```

* Steps to Reproduce

server:
`iperf3 -s`

client:
`iperf3 -c server`  = 944 Mbits/sec
`iperf3 -R -c server`  = 944 Mbits/sec
[...]

PR #1364, opened 2022-07-29
Prevent -k / -n when combined with -R from leaking into future tests.

bmah888:

```
```

* Issues fixed (if any): #1363

The problem was that the -k / -n parameters would get sent to the
server for a reverse tests, but never cleared for future reverse tests,
because these parameters were only sent from the client if they were
non-zero.
.
This is fixed in two ways as a belt-and-suspenders approach.

1. The server always clears the variables used for -k / -n when
receiving parameters, so if the client doesn't send them, they
get cleared (set to zero) anyway.

2. The client sends values for -k / -n even if they're zero. Sending
zero values doesn't hurt anything (except for wasting a few bytes),
and it makes this use case work even for servers that don't have
the fix in item 1 above.


swlars (approved):
Looks good!

Issue #1367, opened 2022-08-01
iperf3 increases packets per seconds for UDP tests after some time


FlyDrag:
# Context

* Version of iperf3:
client (sender): 3.11 (i assume that this is sender issue) - from FreeBSD packages
server (receiver): 3.7 - from FreeBSD packages (upgraded to 3.11 - doesn't help)

* Hardware:
i82576 PCIe-x4 dual NIC.
Intel Pentium G5420 CPU

* Operating system (and distribution, if any):
FreeBSD 13.1-RELEASE (sender)
FreeBSD 12.0-RELEASE-p12 (receiver)

# Bug Report
For some combinations of bandwidth and length iperf3 increases pps (keeping bitrate - so I assume that it decreases the packet size) after some time.

Test setup - two PCs connected by direct ethernet cable. 1 GBps interface speed. 9216 MTU size.

**Test 1.** length 9000, bw 950M - working fine
```
iperf3 -u -b 950M -l 9000 --dont-fragment -c fc2
Connecting to host fc2, port 5201
[  5] local 192.168.254.29 port 41050 connected to 192.168.254.30 port 5201
[ ID] Interval           Transfer     Bitrate         Total Datagrams
[  5]   0.00-1.00   sec   113 MBytes   949 Mbits/sec  13184
[  5]   1.00-2.00   sec   113 MBytes   950 Mbits/sec  13195
[  5]   2.00-3.00   sec   113 MBytes   950 Mbits/sec  13194
[  5]   3.00-4.00   sec   113 MBytes   950 Mbits/sec  13195
[  5]   4.00-5.00   sec   113 MBytes   950 Mbits/sec  13194
[  5]   5.00-6.00   sec   113 MBytes   950 Mbits/sec  13194
[  5]   6.00-7.00   sec   113 MBytes   950 Mbits/sec  13195
[  5]   7.00-8.00   sec   113 MBytes   950 Mbits/sec  13194
[  5]   8.00-9.00   sec   113 MBytes   950 Mbits/sec  13195
[  5]   9.00-10.00  sec   113 MBytes   950 Mbits/sec  13194
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  5]   0.00-10.00  sec  1.11 GBytes   950 Mbits/sec  0.000 ms  0/131934 (0%)  sender
[...]

PR #1379, opened 2022-08-19
Add error message in case of UDP send NET_SOFTERROR

davidBar-On:
master

* Issues fixed (if any):
None

Add debug message in case UDP send failed with `NET_SOFTERROR`.  This is to help evaluating issues such as #1367.

The message is printed on `--debug=3` (or higher level), so it will be possible to see it without seeing a message per UDP packet sent (which is the case when `-d` or `--debug=4` is used).


swlars:
Hi David! Thanks for the merge request! It looks like part of this was merged already, but we're not quite sure how to test the last few lines.

davidBar-On:
Hi Sarah, I don't know what is the change in the indentation of lines 507-510 in `net.c`.  By the way, there is indentations problem in both `Nwrite()` and `nSendfile()`.  The code following the `case/default` is not indented.

Regarding the following that was added in `Nwrite()`:
```c
    if (count == nleft)
        return NET_SOFTERROR;
```

This is copied from `Nsendfile()` and is the suggested fix to #1367 - when a UDP packet was not sent, do not count it as sent and do not increase the UDP packets number.  See [this comment](https://github.com/esnet/iperf/issues/1367#issuecomment-1221254140).

swlars:
Looks good! Thanks for the merge request!

PR #1380, opened 2022-08-20
Do not count UDP messages that were not sent

davidBar-On:
master

* Issues fixed (if any):
#1367

When sending a UDP packet fails, do not increase the packet count.  That should improve the statistics, as the next packet will be sent with the same number - practically retry of sending the failed packet instead of sending a new packet.  Therefore, the client will not count the "retry" as new packet, and more important, the server will not consider the packet that was not sent as a lots packet.

In addition, added debug message in case of `NET_SOFTERROR` when sending UDP messages, to help debugging.

### DIFF
--- a/src/iperf_udp.c
+++ b/src/iperf_udp.c
@@ -250,8 +250,15 @@ iperf_udp_send(struct iperf_stream *sp)
 
     r = Nwrite(sp->socket, sp->buffer, size, Pudp);
 
-    if (r < 0)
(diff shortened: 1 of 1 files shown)


bmah888 (approved):
Thanks for the PR! Looks good to me. Will merge shortly.

As an aside I'm not sure why we increment the sequence number before sending each packet, as opposed to doing it after each successful send.

PR #1390, opened 2022-09-14
Print the shorter usage hint on argument error

jtdor:
* Issues fixed (if any):

Hi there,

in my iperf3 usage I found it a bit unhandy that the whole, long help text is printed when there is an error in the given arguments. I changed the error handling here to only print the short usage hint on such an error. I’m happy about feedback whether this change is welcome.

Citing my commit message below for more reasoning:

When an argument error is printed it is often the result of a simple typo. A user familiar with iperf3 does not need the whole help text to figure out how to fix the command-line. A brief pointer to the erroneous argument is sufficient.

Until now, with the *long* help text printed after an argument error, a user often needs to scroll up to find the message about the erroneous argument.

Another user not yet familiar with iperf3 will still be informed how to access the long help text---or might already have it opened on the side, anyway.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1572,7 +1572,8 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 		usage_long(stdout);
 		exit(0);
             default:
-                usage_long(stderr);
+                fprintf(stderr, "\n");
+                usage();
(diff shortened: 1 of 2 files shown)


bmah888 (approved):
Thanks for the PR! We should have merged this a long time ago.

Issue #1392, opened 2022-09-20
crash when server mode with --logfile


jimying:
server side run with --logfile arg, 
when one client test finish,  server side crash

bmah888:
The crash occurs with the combination of `--logfile` and `--json`.

PR #1393, opened 2022-09-20
Fix crash when server mode with --logfile and --json

jimying:
as tile
#1392 

The reason : log file fd has closed by cleanup_server(), but write it later


bmah888 (approved):
Thanks for the fix!

PR #1394, opened 2022-09-21
Add error handling to the unit_atoX() functions

jtdor:
* Issues fixed (if any):

Hi there,

I added an error handling to the `unit_atoX()` functions here when parsing commandline arguments.

src/units.c needs to include the large iperf_api.h now for i_errno only. It could be cleaner to factor out the declarations of i_errno and related functions to an iperf_error.h instead to reduce the coupling here. I could change that, if preferred?

Citing the commit for more description of the change:

Previously, the unit_atoX() functions did not have any error handling: Calling e.g. `iperf3 -nZ -c ...` would silently hide the fact that -n requires a numeric argument and also not apply the -Z (it was “consumed” by unit_atoi()). The intention was to pass -NZ here.

With this change, iperf3 will refuse options where unit_atoX() did not find any numeric value or detect an invalid suffix (not one of KMGT).


bmah888:
Hi! Thanks for the PR, which I'm finally getting around to looking at. I just committed a change to resolve some merge conflicts against the iperf3 `master` branch.

PR #1396, opened 2022-10-02
Cleanup endian handling on OpenBSD.

brad0:
OpenBSD has had the endian.h header for 8+ years.


bmah888:
Tested on OpenBSD 7.1.

PR #1397, opened 2022-10-02
Add support for OpenBSD systems which have TCP_INFO implemented.

brad0:


bmah888:
Thanks for the PR! We're having a little testing this. On an OpenBSD 7.1/amd64 system (latest available download from openbsd.org) there doesn't appear to be any `struct tcpinfo` structure or `TCP_INFO` socket option. So this feature won't get detected at configure time, and these code changes won't do anything. Are we testing on the wrong platform?

brad0:
TCP_INFO support was added in -current. 7.2 has been tagged and the release is days away. You can also use a -current snap that is now past 7.2 if you want.

https://ftp.openbsd.org/pub/OpenBSD/snapshots/packages/

brad0:
The 7.2 release is now out.

bmah888 (approved):
Compiles and runs correctly (in limited testing) on OpenBSD 7.2. Thanks for the PR!  (I will merge this when I get some iperf3 time.)

brad0:
Thanks.

PR #1400, opened 2022-10-03
Changes in displayed units to their correct form

CodingAlbert:
* Version of iperf3 : master

* Issues fixed (if any): This pull request changes second unit abbreviation `sec` to its SI symbol `s`. Further it fixes data-rate units to their correct form `bits/sec` to `bit/s`.

* source:
https://en.wikipedia.org/wiki/Data-rate_units

Change unit display sec -> s
Change unit display bits/s -> bit/s

### DIFF
--- a/src/iperf_locale.c
+++ b/src/iperf_locale.c
@@ -376,31 +376,31 @@ const char report_bw_format[] =
 "[%3d]%s %6.2f-%-6.2f s  %ss  %ss/s                  %s\n";
 
 const char report_bw_retrans_format[] =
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR! One difficulty with this change is that at least some of the formatting lines you've changed are parts of tables, and the changes disrupt the table formatting. One possible fix is to change "sec" to "s  " so that the new labels take the same amount of space as the old ones. Also there are some merge conflicts because the code has diverged somewhat from when you submitted the PR. Any chance of fixing these up? We'll do a better job of reviewing.

Issue #1414, opened 2022-11-04
Mixed big-endian and little-endian UDP test fails


mriswyth:
* Version of iperf3: 3.12

* Hardware: `x86_64`  and `MIPS 74Kc V5.0`

* Operating system (and distribution, if any): Linux for both systems

# Bug Report

Mixing a big-endian and little-endian systems for the client and server results in UDP tests failing to start.

This is the failed output of an x86_64 PC client command running against a big-endian server 
```
perf3 -c FE80::848D:84FF:FEFE:9165%eno1 -n 1 -u --debug --verbose
iperf 3.12
Linux bthompson.rajant.com 6.0.5-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Oct 26 15:55:21 UTC 2022 x86_64
Control connection MSS 1428
Setting UDP block size to 1428
send_parameters:
{
        "udp":  true,
        "omit": 0,
        "time": 0,
        "num":  1,
        "blockcount":   0,
        "parallel":     1,
        "len":  1428,
        "bandwidth":    1048576,
        "pacing_timer": 1000,
        "client_version":       "3.12"
}
Time: Fri, 04 Nov 2022 17:42:46 UTC
Connecting to host FE80::848D:84FF:FEFE:9165%eno1, port 5201
      Cookie: oc7rjz2iw7yvlmrsxv2gpswowhlcamlqjdna
      Target Bitrate: 1048576
SNDBUF is 212992, expecting 0
RCVBUF is 212992, expecting 0
Setting application pacing to 131072
Sending Connect message to Socket 5
Connect received for Socket 5, sz=4, buf=36373839, i=0, max_len_wait_for_reply=4
iperf3: error - unable to read from stream socket: Invalid argument
```

This is the output of the same client running against a little-endian server
```
iperf3 -c FE80::D0:12FF:FE28:F3BD%eno1 -n 1 -u --debug --verbose
iperf 3.12
Linux bthompson.rajant.com 6.0.5-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Oct 26 15:55:21 UTC 2022 x86_64
Control connection MSS 1428
Setting UDP block size to 1428
send_parameters:
{
        "udp":  true,
        "omit": 0,
        "time": 0,
        "num":  1,
        "blockcount":   0,
        "parallel":     1,
        "len":  1428,
        "bandwidth":    1048576,
        "pacing_timer": 1000,
        "client_version":       "3.12"
[...]

PR #1415, opened 2022-11-04
Address UDP connect message endian issue

mriswyth:
Accept big-endian and little-endian format for both legacy and new messages since the value is sent in host byte order.

```
```

* Issues fixed (if any):
#1414

Accept UDP connect message replies in little-endian or big-endian order.
There was discussion and concern about accepting two values in the original pull request #1260, this makes that problem worse. This is a solution that allows mixed-endian and mixed-version operations if the client is updated.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -430,9 +430,10 @@ struct iperf_test
 extern int gerror; /* error value from getaddrinfo(3), for use in internal error handling */
 
 /* UDP "connect" message and reply (textual value for Wireshark, etc. readability - legacy was numeric) */
-#define UDP_CONNECT_MSG 0x36373839          // "6789" - legacy value was 123456789
-#define UDP_CONNECT_REPLY 0x39383736        // "9876" - legacy value was 987654321
-#define LEGACY_UDP_CONNECT_REPLY 987654321  // Old servers may still reply with the legacy value
+#define UDP_CONNECT_MSG 0x36373839                 // "6789" - legacy value was 123456789
+#define UDP_CONNECT_REPLY 0x39383736               // "9876" - legacy value was 987654321
+#define LEGACY_UDP_CONNECT_REPLY 987654321         // Old servers may still reply with legacy value
+#define LEGACY_UDP_CONNECT_REPLY_ENDIAN 2976439866 // Old servers may still reply with legacy value
 
 /* In Reverse mode, maximum number of packets to wait for "accept" response - to handle out of order packets */
 #define MAX_REVERSE_OUT_OF_ORDER_PACKETS 2
--- a/src/iperf_udp.c
+++ b/src/iperf_udp.c
@@ -592,9 +592,9 @@ iperf_udp_connect(struct iperf_test *test)
             printf("Connect received for Socket %d, sz=%d, buf=%x, i=%d, max_len_wait_for_reply=%d\n", s, sz, buf, i, max_len_wait_for_reply);
         }
         i += sz;
-    } while (buf != UDP_CONNECT_REPLY && buf != LEGACY_UDP_CONNECT_REPLY && i < max_len_wait_for_reply);
+    } while (buf != UDP_CONNECT_REPLY && buf != LEGACY_UDP_CONNECT_REPLY && buf != UDP_CONNECT_MSG && buf != LEGACY_UDP_CONNECT_REPLY_ENDIAN && i < max_len_wait_for_reply);
 
-    if (buf != UDP_CONNECT_REPLY  && buf != LEGACY_UDP_CONNECT_REPLY) {
+    if (buf != UDP_CONNECT_REPLY && buf != LEGACY_UDP_CONNECT_REPLY && buf != UDP_CONNECT_MSG && buf != LEGACY_UDP_CONNECT_REPLY_ENDIAN) {
         i_errno = IESTREAMREAD;
         return -1;
     }


mriswyth:
Here's the output running against the same server binary that shows the failed output in #1414.
```
./iperf3 -c FE80::848D:84FF:FEFE:9165%eno1 -n 1 -u --debug --verbose
iperf 3.12
Linux bthompson.rajant.com 6.0.5-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Oct 26 15:55:21 UTC 2022 x86_64
Control connection MSS 1428
Setting UDP block size to 1428
send_parameters:
{
        "udp":  true,
        "omit": 0,
        "time": 0,
        "num":  1,
        "blockcount":   0,
        "parallel":     1,
        "len":  1428,
        "bandwidth":    1048576,
        "pacing_timer": 1000,
        "client_version":       "3.12"
}
Time: Fri, 04 Nov 2022 18:01:18 GMT
Connecting to host FE80::848D:84FF:FEFE:9165%eno1, port 5201
      Cookie: feepi4jyp4up4bj2fw54nidyqe27qh76x6wk
      Target Bitrate: 1048576
SNDBUF is 212992, expecting 0
RCVBUF is 212992, expecting 0
Setting application pacing to 131072
Sending Connect message to Socket 5
Connect received for Socket 5, sz=4, buf=36373839, i=0, max_len_wait_for_reply=4
[  5] local fe80::bf96:250b:5c8e:66f9 port 56896 connected to fe80::848d:84ff:fefe:9165 port 5201
Starting Test: protocol: UDP, 1 streams, 1428 byte blocks, omitting 0 seconds, 1 bytes to send, tos 0
sent 1428 bytes of 1428, total 1428
send_results
{
        "cpu_util_total":       17.9589679968755,
        "cpu_util_user":        6.8784892135915641,
        "cpu_util_system":      11.080478783283938,
        "sender_has_retransmits":       0,
        "streams":      [{
                        "id":   1,
                        "bytes":        1428,
                        "retransmits":  18446744073709551615,
                        "jitter":       0,
                        "errors":       0,
                        "packets":      1,
                        "start_time":   0,
                        "end_time":     0.000505
                }]
}
get_results
{
        "cpu_util_total":       3.8469305331179324,
        "cpu_util_user":        0,
        "cpu_util_system":      3.836833602584814,
        "sender_has_retransmits":       1.8446744073709552e+19,
        "streams":      [{
                        "id":   1,
                        "bytes":        0,
                        "retransmits":  1.8446744073709552e+19,
                        "jitter":       0,
                        "errors":       0,
                        "packets":      0,
                        "start_time":   0,
                        "end_time":     0.008745
                }]
}
interval_len 0.000505 bytes_transferred 1428
interval forces keep
[ ID] Interval           Transfer     Bitrate         Total Datagrams
[  5]   0.00-0.00   sec  1.39 KBytes  22.6 Mbits/sec  1  
Test Complete. Summary Results:
[  5]   0.00-0.00   sec  1.39 KBytes  22.6 Mbits/sec  0.000 ms  0/1 (0%)  sender
[  5]   0.00-0.01   sec  0.00 Bytes  0.00 bits/sec  0.000 ms  0/0 (0%)  receiver
CPU Utilization: local/sender 18.0% (6.9%u/11.1%s), remote/receiver 3.8% (0.0%u/3.8%s)
```

BrainSlayer:
see my pull request for a better solution. this commit here does not fix the wrong reply message. it just checks for both types on client side. so if you use it on a big endian host as server and the client is still a older version it will still fail. so both client and server side must use the correct endian type message which is little endian for historic reasons

src/iperf.h
```diff
@@ -430,9 +430,10 @@ struct iperf_test
 extern int gerror; /* error value from getaddrinfo(3), for use in internal error handling */
 
 /* UDP "connect" message and reply (textual value for Wireshark, etc. readability - legacy was numeric) */
-#define UDP_CONNECT_MSG 0x36373839          // "6789" - legacy value was 123456789
-#define UDP_CONNECT_REPLY 0x39383736        // "9876" - legacy value was 987654321
-#define LEGACY_UDP_CONNECT_REPLY 987654321  // Old servers may still reply with the legacy value
+#define UDP_CONNECT_MSG 0x36373839                 // "6789" - legacy value was 123456789
+#define UDP_CONNECT_REPLY 0x39383736               // "9876" - legacy value was 987654321
+#define LEGACY_UDP_CONNECT_REPLY 987654321         // Old servers may still reply with legacy value
+#define LEGACY_UDP_CONNECT_REPLY_ENDIAN 2976439866 // Old servers may still reply with legacy value
```
davidBar-On, line 436: Per issue #1608, reverse Endian of `UDP_CONNECT_REPLY` should also be considered. therefore should also add:
```c
#define UDP_CONNECT_REPLY_ENDIAN 0x36373839
```

src/iperf_udp.c
```diff
@@ -592,9 +592,9 @@ iperf_udp_connect(struct iperf_test *test)
             printf("Connect received for Socket %d, sz=%d, buf=%x, i=%d, max_len_wait_for_reply=%d\n", s, sz, buf, i, max_len_wait_for_reply);
         }
         i += sz;
-    } while (buf != UDP_CONNECT_REPLY && buf != LEGACY_UDP_CONNECT_REPLY && i < max_len_wait_for_reply);
+    } while (buf != UDP_CONNECT_REPLY && buf != LEGACY_UDP_CONNECT_REPLY && buf != UDP_CONNECT_MSG && buf != LEGACY_UDP_CONNECT_REPLY_ENDIAN && i < max_len_wait_for_reply);
```
davidBar-On, line 595: Check also for `UDP_CONNECT_REPLY_ENDIAN`.

src/iperf_udp.c
```diff
@@ -592,9 +592,9 @@ iperf_udp_connect(struct iperf_test *test)
...
 
-    if (buf != UDP_CONNECT_REPLY  && buf != LEGACY_UDP_CONNECT_REPLY) {
+    if (buf != UDP_CONNECT_REPLY && buf != LEGACY_UDP_CONNECT_REPLY && buf != UDP_CONNECT_MSG && buf != LEGACY_UDP_CONNECT_REPLY_ENDIAN) {
```
davidBar-On, line 597: Check also for `UDP_CONNECT_REPLY_ENDIAN`.

BrainSlayer: this is already covered by my pull request. but the iperf3 maintainer seem to ignore all pull requests and bug fixes since a long time. i fixed it in my own tree and i'm happy with it. every effort here is pointless. check my pull request. its a very simple solution and covers all problems mentioned here

davidBar-On: OK. Thanks.  I assume you are referring to PR #1468.  By the way, I see that #1564 is another suggested PR to fix the issue.

mriswyth: The following is copied from the first comment in this PR...
>Accept UDP connect message replies in little-endian or big-endian order.
There was discussion and concern about accepting two values in the original pull request https://github.com/esnet/iperf/pull/1260, this makes that problem worse. This is a solution that allows mixed-endian and mixed-version operations if the client is updated.

It's been over a year since I put this together so I really don't remember the specifics of what I found from the original change PR, but it was reasonable enough to make me try to limit this change. I work in an environment where updating the client is much more reasonable and frequent than updating the server. 3.12 is broken for mixed endian no matter what so something has to be updated. I was trying to limit the scope of damage that I was re-introducing that #1260 was attempting to fix. Since the author of #1260 is now part of the conversation I can assume he has a better handle on how many issues fixing everything will introduce.

If those concerns aren't real I also prefer a solution that addresses both client and server. That's an additional patch that I'm carrying on my product on top of this one since I'm unaware of ever having #1260 occur and cross-endian and version compatibility is important to me.

bmah888:
Thanks for the PR! I chose to merge a slightly simpler fix from PR #1468.

PR #1423, opened 2022-11-20
Add control connection keepalive arguments

### LINKED ISSUE #812: Cannot test long UDP sessions
# Context

**Server side**
`iperf 3.6 (cJSON 1.5.2)` on Ubuntu 16.04.1 (built from source code)

Two servers running concurrently on the same `n1-highcpu-2` (2 vCPUs, 1.8 GB memory) VM on Google Cloud Platform:
`iperf3 -s -p5201` (used to test server to client path)
`iperf3 -s -p5202` (used to test client to server path)

**Client side**
`iperf 3.6 (cJSON 1.5.2)` on Windows 10 (binary obtained from [here](https://files.budman.pw/iperf3.6_64bit.zip))

Two clients running concurrently on the same i&-7820HQ @ 2.90GHz machine, one in normal mode and one in reverse mode:
`iperf3 -c X.X.X.X -u -b1M -p5201 -R -t7400` (used to test server to client path)
`iperf3 -c X.X.X.X -u -b1M -p5202 -t7400` (used to test client to server path)

# Bug Report

The tests conclude with the same error on both servers:
```
[  5] 7403.00-7404.00 sec   123 KBytes  1.00 Mbits/sec  86
Server listening on 5201
```
```
[  5] 7403.00-7404.00 sec  0.00 Bytes  0.00 bits/sec  0.039 ms  0/0 (0%)
Server listening on 5202
```

And both clients report a connection reset:
```
[  5] 7398.00-7399.00 sec   123 KBytes  1.00 Mbits/sec  0.199 ms  0/86 (0%)
iperf3: error - unable to receive control message: Connection reset by peer
```
```
[  5] 7398.00-7399.00 sec   121 KBytes   992 Kbits/sec  85
iperf3: error - unable to receive control message: Connection reset by peer
```

This problem is systematic with test duration of 2 hours or above, while I have not noticed the same problem with shorter tests, such as one hour. However, I cannot say that these observations are conclusive.

This situation situation is particularly severe when option `-J` is used and the client in normal mode is launched with the `--get-server-output` option. Upon experiencing the problem, the server:
- Does not produce the JSON output;
- Does not pass the results to the client.

Therefore, the statistics at the server side are completely lost.

# Notes

I see that there are past issues related to cases when the test duration at the server side appears longer than at the client side like above. Where they supposed to be resolved in v3.6?

Issue #735 has definitely some commonalities with this. In  particular, it stands out that also in my case the server error occurs when the test duration at the server side gets 5s longer than what requested by the client. However, in my case, the connection under test is quite fast. 

```
Ping statistics for X.X.X.X:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 12ms, Maximum = 13ms, Average = 12ms
```
There is simply no way that it can hold 4-5 seconds of packets in flight. Why the duration at the server and client side appear different?

Is there any reason why I may see different behavior when the test duration gets longer, say close to 2 hours? I mean, are there timeouts that may affect the TCP control connection while the tests run on UDP? Or is it some sort of timing skew between client and server that gets accumulated?

Thanks!!

### LINKED ISSUE #835: intermittent connection
# Context

When conducting iPerf test, a spotty WiFI connection can be disconnected for X amount of time. Request to add a user configurable TCP control channel timeout to prevent client from closing socket.

* Version of iperf3: 3.6

* Hardware: VM 

* Operating system (and distribution, if any): Ubuntu

# Enhancement Request

* Current behavior

When a client starts a UDP test, if the client loses connection and connects to a different wifi client throws
iperf3: error - unable to write to stream socket: Invalid argument
The server then continues to finish the test until specified length is completed + ~4 second timeout of 3.6 iperf

* Desired behavior

When a client loses connections (brief <45 seconds, due to poor internet connection), client will wait a specified timeout before closing connection so that test can keep continuing.

* Implementation notes

davidBar-On:
master

* Issues fixed (if any):
#812, #835

Adding Control-connection TCP Keepalive arguments: `--cntl-ka [<KEEPIDLE>/<KEEPINTV>/<KEEPCNT>]`.

As suggested in the resolved issues, setting the keepalive period (`TCP_KEEPIDLE`) resolves #812 by allowing to keep the control connection alive longer than the firewall timeout, and resolves #835 by allowing the server to terminate inactive connections after shorter time.

Added `HAVE_TCP_KEEPALIVE`, based on `TCP_KEEPIDLE` availability, as I am not sure if and which OSs support the TCP keepalive.  In principle, there could be a separated "HAVE" for the socket (not TCP) `SO_KEEPALIVE`, but I assume this is not necessary.

_(**Note** based on the discussion below starting with [this comment](https://github.com/esnet/iperf/pull/1423#issuecomment-1437758824): to use this PR before it is merged into main line, `autoconf` and `autoheader` should be run before `./configure`.  If `autoconf` fails with error that version 2.71 should be used, run `autoconf -V` to get its version and change "2.71" in `configure.ac` to that version.)_

I was **not** able to fully test the code, as I was not able to test the keepalive retries.

Also, for some reason, in Windows WSL2 it seems that the keepalive period (`TCP_KEEPIDLE`) must be greater than the retry interval (`TCP_KEEPINTV`).  Otherwise, the keepalive is sent only once. As I couldn't test the retries, I assumed that it may even be that the keepalive period must be greater than the retry period (`TCP_KEEPINTV` * `TCP_KEEPCNT`).  Therefore, in case this condition is not met and only the keepalive period was set, I added code that reduces the retry interval accordingly.


davidBar-On:
@ankitg12, I reply to your [comment](https://github.com/esnet/iperf/issues/812#issuecomment-1375325969) here, as building the change is using the code of this PR.

>Any idea when this change would be merged ? Or how I can get binary with this change?

As this PR is not merged (yet) into the main code, there are probably no available binaries.  You can try building it yourself.  In general, the process is (under Linux):

```sh
git clone https://github.com/davidBar-On/iperf.git --branch issue-812-835-control-connection-keepalive
cd iperf
./confoigure
make
```

The iperf3 executable will be `./src/iperf3`.  To make sure it was built o.k. run `./src/iperf3 -h`.  The help should include a line for the new `--cntl-ka` option.

Note that the PR is not fully tested.  If you were able to build and run iperf3 successfully and you find problems with the implementation of the new option, please report them in this PR. Thanks.

lbsou:
I guess this is probably obvious, but I had to add `#define HAVE_TCP_KEEPALIVE 1` in `./iperf/src/iperf_config.h` before launching `make` to have the `--cntl-ka` option available.

I can confirm with `netstat -tnope` that the new parameter change the control channel behavior correctly: 

KEEPIDLE = Modify the interval of the keepalive
KEEPINTV = Modify the interval of the retry
KEEPCNT = Modify the amount of keepalive lost before dropping the connection

I would be very happy to see this change merge into the master branch!

davidBar-On:
> .... I had to add `#define HAVE_TCP_KEEPALIVE 1` in `./iperf/src/iperf_config` ....

@lbsou, `configure.h` is generated by `./iperf3/configure` per `./iperf3/configure.ac`. Currently, `configure.ac` defines `HAVE_TCP_KEEPALIVE` is defined in `confugure.h` when **`netinet/tcp.h` defines `TCP_KEEPIDLE`**.

Which OS are you using? Can you check if and where `TCP_KEEPIDLE` is defined?

davidBar-On:
Hi, thanks for the info! Hopefully I now understand the problem.  When (if...) the PR will be merged into the mainline, `autoconf` and `autoheader` will be run to updated the `./cofigure` file based on the `./configure.ac`.  Since the PR is not merged, `./configure` does not handle the definition of `HAVE_TCP_KEEPALIVE`.

Can you try running `autoconf`?  It should updated `./configure` file, which defines `HAVE_TCP_KEEPALIVE`.  If `autoconf` fails with error that version 2.71 should be used, run `autoconf -V` to get its version and change "2.71" in `configure.ac` to that version.

Now run  `autoheader` and then `./configure`,  `./src/iperf_config.h` should now define `HAVE_TCP_KEEPALIVE`.

**Was this successful?**  If it did, I will add the above instructions to the PR description

lbsou:
Thank you David, that was it!

I did ``autoreconf --verbose --install --force`` then ``./configure`` and the ``./src/iperf_config.h`` now contain ``#define HAVE_TCP_KEEPALIVE 1``

Any idea when this change will be merged?

davidBar-On:
@lbsou, I have updated the PR description.  Thanks for your efforts and time for clarifying this issue (and helping me to better understand the configuration process ...).

I am not from the team that maintains iperf3, so I don't know if and when this PR will be merged into the mainline.

lbsou:
Thank you David!

@bmah888 Can we be of any further help for this PR to be merged?

Thanks!

davidBar-On:
Hi @lbsou, thanks for testing and finding this issue.
Actually, setting the keepalive only on the client side was intentional, and just the help for the option wrongly showed it for both the client and the server ...

**However**, when evaluating your comment I realized that it is better that the keepalive option will be available for both the client and server.  I enhanced the solution and the option should work now on both sides.

davidBar-On:
Hi @lbsou, thanks for testing and finding this issue.

Actually, setting the keepalive only on the client side was intentional, as the client is initiating the control stream, but I forgot to limit the option for the client.  **However**, when evaluating your comment I realized that it is better that the keepalive option will be available for both the client and the server.  I enhanced the solution and the option should work now on both sides.

lbsou:
Thanks you very much @davidBar-On, may the god of merging be with us!

swlars:
We might look into testing and merging this soon. We're interested in the changes. In the meantime, if you could update the man page for the new command line argument that would be great, otherwise we'll probably get to it once multi-threading is merged.

swlars:
Thanks for the merge request! We might want to consider adding a little more documentation to make it clear that this is client-side only, but other than that it looks great!

swlars:
Tested and looks good!

swlars:
Thanks again for the pull request!

PR #1453, opened 2023-01-10
Add getter for test bidirectional

utahwithak:
`master`

* Issues fixed (if any):
I didn't add an issue but can. 

Adds a simple getter for bidirectional field of the test. Matches the pattern for all other getter functions and follows after reverse as it is laid out in the struct. 

The setter already exists.


bmah888 (approved):
Looks good. Weird that we had a setter but no getter.

PR #1468, opened 2023-02-07
fix endian type of udp connect / reply messages

BrainSlayer:
for example. if you run a iperf3 server on a little endian host, but client is big endian it will result in the following error if you try to start a udp bandwidth test

Connect received for Socket 5, sz=4, buf=36373839, i=0, max_len_wait_for_reply=4 iperf3: error - unable to read from stream socket: Invalid argument

the reason that that all messages are reversed. fix this by swapping the message values to little endian host order on big endian systems

Signed-off-by: Sebastian Gottschall <s.gottschall@dd-wrt.com>

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -67,6 +67,7 @@
 #include "queue.h"
 #include "cjson.h"
 #include "iperf_time.h"
+#include "portable_endian.h"
 
 #if defined(HAVE_SSL)
 #include <openssl/bio.h>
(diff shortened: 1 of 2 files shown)


BrainSlayer:
when is this going to be merged? i mean i see you prepare new releases but serious bugs like this here are remaining unresolved

RayWindRiver:
Please merged this commit.

bmah888 (approved):
OK this looks good, thanks for the PR. I like this solution compared to some others because it keeps all of the byte order dependencies are contained within one file. This works because the values involved are just magic numbers, we're not computing anything on them.

Issue #1482, opened 2023-03-14
Issue of argument  '-i' when omitting the value


sheercoding:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the Discussions section for this project at
https://github.com/esnet/iperf/discussions
or to the iperf3 mailing list at iperf-dev@googlegroups.com.
A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

_Massive performance deviation when attaching or omitting the value 0 after the argument of '-i'._

* Version of iperf3:
_iperf 3.9 (cJSON 1.7.13)_

* Hardware:
_Intel(R) Xeon(R) Platinum 8180 CPU @ 2.50GHz_
* Operating system (and distribution, if any):
* 
_DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS"_

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate. Note that submissions of bug fixes, new
features, etc. should be done as a pull request at
https://github.com/esnet/iperf/pulls_

# Bug Report

* Expected Behavior

$iperf3 -c 10.20.0.3 -i **0** -u -b 0
Connecting to host 10.20.0.3, port 5201
[  4] local 10.20.0.1 port 58558 connected to 10.20.0.3 port 5201
[ ID] Interval           Transfer     Bandwidth       Total Datagrams
[  4]   0.00-10.00  sec  4.73 GBytes  4.06 Gbits/sec  619440
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bandwidth       Jitter    Lost/Total Datagrams
[  4]   0.00-10.00  sec  **4.73 GBytes  4.06 Gbits/sec**  0.000 ms  0/0 (0%)
[  4] Sent 0 datagrams

iperf Done.

* Actual Behavior

$iperf3 -c 10.20.0.3 -i -u -b 0
Connecting to host 10.20.0.3, port 5201
[...]

PR #1485, opened 2023-03-17
Fix issue #1482 - not allow using option as missing argument for an o…

davidBar-On:
master

* Issues fixed (if any):
#1482

Suggested fix for preventing iperf3 from using an option as a missing argument for the previous option.  I.e. in `-i -u` the "-u" was used as the `-i` argument, and the default TCP was running instead of UDP test.

The suggested fix assumes that negative argument is not allowed for all numeric options, and that no string argument can start with "-" (e.g. for `title`, `extra-data`, `username`, `file`).  A more robust fix could be by checking that `optarg` is numeric in each case where `atoi()` is used.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1121,6 +1121,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
     struct xbind_entry *xbe;
     double farg;
     int rcv_timeout_in = 0;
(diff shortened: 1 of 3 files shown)


swlars:
We like the idea behind this, but we're not quite sure if this is the best solution because of the restriction that it places on input parameters.

Issue #1471, opened 2023-02-15
When the file doesnt exists (--rsa-public-key-path), execution results in Segmentation fault


declum:
# Context

* Version of iperf3: 

iperf 3.9 (cJSON 1.7.13)
Linux lcm-worker-vm 5.15.0-60-generic #66-Ubuntu SMP Fri Jan 20 14:29:49 UTC 2023 x86_64
Optional features available: CPU affinity setting, IPv6 flow label, SCTP, TCP congestion algorithm setting, sendfile / zerocopy, socket pacing, authentication

iperf 3.12 (cJSON 1.7.15)
Darwin cjebakumarPF2K7.vmware.com 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:03:51 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T6000 arm64
Optional features available: sendfile / zerocopy, authentication, support IPv4 don't fragment

* Operating system (and distribution, if any): Mac 13.1 (installed by brew) / Ubuntu 22.04

# Bug Report
When the file doesnt exists (--rsa-public-key-path), execution results in Segmentation fault

* Expected Behavior
iperf3 should show a message that the file doesn't exists. Currently, it emits Segmentation fault (as String) even for -J/--json flag. So the caller needs to have special handling.

* Actual Behavior
Segmentation fault

* Steps to Reproduce
iperf3 --get-server-output --json -c <ip> --port 5201 --username user --rsa-public-key-path /tmp/public-missing-file.pem --connect-timeout 3000 --time 5
Password: <password>
Segmentation fault (core dumped)

* Possible Solution
stats the file, and show error if its not found.

PR #1488, opened 2023-03-22
Verify success of BIO_new_file()

davidBar-On:
master

* Issues fixed (if any):
* #1471

Suggested fix crash when public/private key file does not exist - may not be th full fix as the issue didn't happen on my PC.  However, the output of `BIO_new_file()` was not checked, which could lead to such crash in `PEM_read_bio_xxxxKey()` (depending on the library implementation).

In addition moved reading the private key file before prompting for the password, so the user will not have to enter password before the failure because file does not exist.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1626,15 +1626,15 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 
         char *client_password = NULL;
         size_t s;
+        if (test_load_pubkey_from_file(client_rsa_public_key) < 0){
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the PR! While I was setting up to test this, I realized that the original crash doesn't happen on my test setups either (FreeBSD 13 and Ubuntu 20). But I think having the added error handling is a good idea, as well as the workflow change for getting the passwd from the user.

bmah888 (approved):
This looks great, let's get this merged.

PR #1490, opened 2023-03-24
Add Voice Admit DSCP code point from RFC 5865

rpigott:


src/dscp.c
```diff
@@ -90,6 +90,9 @@ const char * iptos2str(int iptos);
 #ifndef IPTOS_DSCP_EF
 # define	IPTOS_DSCP_EF		0xb8
 #endif /* IPTOS_DSCP_EF */
+#ifdef IPTOS_DSCP_VA
```
bmah888, line 93: Should this line begin with `#ifndef` instead of `#ifdef`? This appears to be why this branch fails CI builds.

bmah888 (changes requested):

src/dscp.c
rpigott: Yes, looks like I goofed there.

bmah888 (approved):
Looks good, thanks for the PR!

Issue #1489, opened 2023-03-24
lost_percent greater than 100%


joudinet:
# Context

* Version of iperf3: 3.12

* Hardware: armv7l

* Operating system (and distribution, if any): Linux (Buildroot)

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):
Optional features available: CPU affinity setting, IPv6 flow label, TCP congestion algorithm setting, sendfile / zerocopy, socket pacing, authentication, bind to device, support IPv4 don't fragment

# Bug Report

Running an UDP test with a bitrate of 300Mb/s, Iperf3 sometimes reports a value of lost_percent greater than 100.

* Expected Behavior

I expect lost_percent to always be below 100.

* Actual Behavior

"lost_percent": 179.64904660556834,

* Steps to Reproduce

iperf3 --json --client TARGET_IP --udp --bitrate 300M --omit 5 --time 5 --interval 0

where the link to TARGET_IP reports a TCP throughput below 300 Mb/s.

* Possible Solution

No idea. I guess either the counting of sent packets or the counting of lost packets is wrong.

PR #1498, opened 2023-04-05
Statistics improvements when omitted is used

davidBar-On:
master

* Issues fixed (if any):
#1489

Suggested fixes so Sender statistics when `--omitt` is used:
1. Fix issue #1489: issue was caused because the sender statistics did not ignore the lost packets during the omit period.
2. Receiver statistics on sender side: use both packet count and omitted counts from receiver.  Because of UDP packets delay, the number of ignored packets on sender and receiver are different - more packets are ignored on the sender since some of them did not arrive yet to the receiver at the end of the omit period.  Therefore, the %lost packets that used the sender's ignored cont and the receiver's lost count was wrong.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2410,7 +2410,9 @@ send_results(struct iperf_test *test)
 		    cJSON_AddNumberToObject(j_stream, "retransmits", retransmits);
 		    cJSON_AddNumberToObject(j_stream, "jitter", sp->jitter);
 		    cJSON_AddNumberToObject(j_stream, "errors", sp->cnt_error);
+                    cJSON_AddNumberToObject(j_stream, "omitted_errors", sp->omitted_cnt_error);
 		    cJSON_AddNumberToObject(j_stream, "packets", sp->packet_count);
+                    cJSON_AddNumberToObject(j_stream, "omitted_packets", sp->omitted_packet_count);
 
 		    iperf_time_diff(&sp->result->start_time, &sp->result->start_time, &temp_time);
 		    start_time = iperf_time_in_secs(&temp_time);
@@ -2457,10 +2459,12 @@ get_results(struct iperf_test *test)
     cJSON *j_retransmits;
     cJSON *j_jitter;
     cJSON *j_errors;
+    cJSON *j_omitted_errors;
     cJSON *j_packets;
+    cJSON *j_omitted_packets;
     cJSON *j_server_output;
     cJSON *j_start_time, *j_end_time;
-    int sid, cerror, pcount;
+    int sid, cerror, pcount, omitted_cerror, omitted_pcount;
     double jitter;
     iperf_size_t bytes_transferred;
     int retransmits;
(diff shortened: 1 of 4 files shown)


src/iperf_api.c
```diff
@@ -2513,10 +2517,12 @@ get_results(struct iperf_test *test)
 			j_retransmits = cJSON_GetObjectItem(j_stream, "retransmits");
 			j_jitter = cJSON_GetObjectItem(j_stream, "jitter");
 			j_errors = cJSON_GetObjectItem(j_stream, "errors");
+                        j_omitted_errors = cJSON_GetObjectItem(j_stream, "omitted_errors");
 			j_packets = cJSON_GetObjectItem(j_stream, "packets");
+                        j_omitted_packets = cJSON_GetObjectItem(j_stream, "omitted_packets");
 			j_start_time = cJSON_GetObjectItem(j_stream, "start_time");
 			j_end_time = cJSON_GetObjectItem(j_stream, "end_time");
-			if (j_id == NULL || j_bytes == NULL || j_retransmits == NULL || j_jitter == NULL || j_errors == NULL || j_packets == NULL) {
+			if (j_id == NULL || j_bytes == NULL || j_retransmits == NULL || j_jitter == NULL || j_errors == NULL || j_omitted_errors == NULL || j_packets == NULL || j_omitted_packets == NULL) {
```
bmah888, line 2525: There's a potential interoperability problem here...if an iperf3 with this change is doing a test against an iperf3 before this change, the older iperf3 won't have `omitted_errors` or `omitted_packets` members in the per-stream statistics structure. So this conditional will fail and cause the test to error out. I haven't figured out a good way to do backward compatibility for this yet, but I think it's somewhat important.

bmah888 (changes requested):
Nice diagnosis of the bug, and thanks for the PR!

There's an issue with interoperability with older versions of iperf3, see feedback comment. I'm not sure how to fix this yet but hopefully you have an idea since you've been thinking about this. Thanks!

src/iperf_api.c
davidBar-On: Forgot about backward compatibility ...  I **re-implemented the change**, taking this into account.  When the receiver error count during the omit time is not known (because older version is used), the new `report_bw_udp_format_no_omitted_error` message format is used that shows the error count as "Unknown" (and doesn't show lost percent).

bmah888: OK great, the newest version of this PR seems to restore interoperability. Thanks!

bmah888:
How to test this:

Server: `iperf3 --server`
Client: `iperf3 --client client.example.org --time 5 --udp --bandwidth=1g` (pick some bandwidth that's high enough to cause some packet loss, get a rough baseline of loss rate)
`iperf --client client.example.org --time 5 --omit 3 --udp --bandwidth=1g` (loss rate will probably be inflated before the fix in this PR, but should be similar with the PR)

Test with various combinations of old and new client and server. Also try `--reverse` mode.

Note that with a new client sending to an old server with `--omit`, the client will not be able to display the loss statistics from the server.

bmah888 (approved):

PR #1500, opened 2023-04-07
Do not allow File transfer using UDP

davidBar-On:
master

* Issues fixed (if any):
None

Found that iperf3 crashes when trying to transfer a file (`-F`) using UDP.  This is because UDP block size is fixed, but the last file block is usually smaller than the block size.   Also, the file transfer would fail anyway because of the UDP packet header.

Added an error when trying to transfer a file using UDP.

### DIFF
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -350,6 +350,10 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "send timeout value is incorrect or not in range");
             perr = 1;
             break;
+        case IEUDPFILETRANSFER:
+            snprintf(errstr, len, "cannot transfer file using UDP");
+            perr = 1;
+            break;
         case IERVRSONLYRCVTIMEOUT:
             snprintf(errstr, len, "client receive timeout is valid only in receiving mode");
             perr = 1;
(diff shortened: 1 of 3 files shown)


src/iperf_error.c
```diff
@@ -350,6 +350,10 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "send timeout value is incorrect or not in range");
             perr = 1;
             break;
+        case IEUDPFILETRANSFER:
+            snprintf(errstr, len, "cannot transfer file using UDP");
+            perr = 1;
```
bmah888, line 355: Don't want to set `perr = 1` here, there's no reason to output the result of perror(3) because no library calls were made to cause this error (it's the user giving a bad combination of command-line arguments).

bmah888 (changes requested):
Good idea, thanks for the PR. Another not-well-thought-out case of features that don't interact well together.

Just one change to make before I can commit this, see review comment.

src/iperf_error.c
davidBar-On: Done.  Removed the `perr = 1`.

bmah888: Great, thanks...I think we're good to merge at this point.

bmah888 (approved):

Issue #1233, opened 2021-11-21
iperf3: error - unable to send control message: Bad file descriptor


bharathk-9:
ISSUE: iperf3: error - unable to send control message: Bad file descriptor

Hello Team,

I'm running network testing on my two systems using the iperf3 command, where the testing is getting successful in some time and it's failing in some time with the error message "iperf3: error - unable to send control message: Bad file descriptor".  I assume the connectivity is the problem so to test the connectivity between both systems, I am running a ping test which is passed.  In my next step, I am doing a retry in case of the first attempt failed in the iperf3 test. in my all attempt the results were the same(frequent failure with the error message). 

Iperf Version: iperf 3.9 (cJSON 1.7.13)
Command used: iperf3 --client <serverIP> --json

Help me to identify the issue and fix them, Thanks in advance.

swlars:
Hi, we might need a bit more information to figure out your issue. 

1. What operating system are you using?
2. Are you immediately retrying the test or is there some kind of sleep?
3. Can you try with a more recent version of iperf3; the latest is 3.10.1.

bharathk-9:
Hi @swlars,

Thank you for your response. please find my inline answers.

1. What operating system are you using?
 ANS: Its Ubuntu 18.04.
2. Are you immediately retrying the test or is there some kind of sleep?
 ANS: As of now I'm giving some sleep time but again it depends on the time the server starts up(which we don't check as of now).
3. Can you try with a more recent version of iperf3; the latest is 3.10.1.
 ANS: let me try and give you update.

As per your 2nd question there can be a chance the client start reaching to server as soon as the server starts up, let me try with some check when the server starts and from there ill give some sleep time. But want to know what is the expected sleep time we want to give so that client can reach to server after the server starts up.


Thanks,
Bharath K

davidhan1120:
That message shows up when the port is not available. Check your firewall settings or the port that you are trying to test.

bharathk-9:
I'm not mentioning any port along with command also I don't have any firewall to block, as this is in my own system configured as private network.

davidhan1120:
If you're not mentioning any port with the command, by default it is going to need to use port 5001. Assuming that you are using this on your private network, you need to enable port forwarding on your home router.

ilham-Hamzah:
does it support UPNP?

TheRealDJ:
The default iperf server port is 5201 which must be open on the server's firewall.

brosahay:
[...]

PR #1504, opened 2023-04-16
Enhanced errors text for some control messages errors

davidBar-On:
master

* Issues fixed (if any):
None

Based on [this comment](https://github.com/esnet/iperf/issues/1233#issuecomment-1509369780) I think it will help to enhance two errors messages regarding failure to connect to the server - especially for cases when iperf server is used by mistake (when forgetting the "3"), server and client does not the same port, or server is not running. (I am aware of the issue with long error messages, but I didn't find a shorter version.)

### DIFF
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -228,7 +228,7 @@ iperf_strerror(int int_errno)
             perr = 1;
             break;
         case IECONNECT:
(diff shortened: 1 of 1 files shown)


bmah888 (approved):
Thanks, this looks like a good and useful change.

PR #1505, opened 2023-04-20
Fix unable to create a new stream on android system.

dongfengweixiao:

```
```

  `3.1-STABLE`) to which this pull request applies:master

* Issues fixed (if any):

when run iperf3 on android as server,

```
/data/local/tmp/iperf3 -s                                          
Server listening on 5201
Accepted connection from 192.168.0.2, port 61642
iperf3: error - unable to create a new stream: No such file or directory
Server listening on 5201
```

cause iperf3 use ‘/tmp’ as default tmp folder, but android only support ‘/data/local/tmp’.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4230,6 +4230,11 @@ iperf_new_stream(struct iperf_test *test, int s, int sender)
         if (tempdir == 0){
             tempdir = getenv("TMP");
         }
+#if defined(ANDROID) || defined(_ANDROID_)
+        if (tempdir == 0 && getenv("ANDROID_ART_ROOT")) {
+            tempdir = "/data/local/tmp";
+        }
+#endif
         if (tempdir == 0){
 #if defined(__ANDROID__)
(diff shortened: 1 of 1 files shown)


davidBar-On (commented):
> #if defined(ANDROID) || defined(\_ANDROID_\)

Why not use `#if defined(__ANDROID__)`?  Per [this site](https://iq.opengenus.org/detect-operating-system-in-c/) for example, `__ANDROID__` is used to detect Android..  Not ethat `__ANDROID__` is already used in _flowlabel.h_.

> if (tempdir == 0 && getenv("ANDROID_ART_ROOT")) {
```
What is `ANDROID_ART_ROOT`?  Is it general env. variable used for all Android apps?  Why the "if defined android" check is not enough?

davidBar-On:
Thanks for the clarifications.  I think that the **first "build time check" approach** is better, assuming that you will verify that it works o.k.  Can you modify the PR accordingly?

PR #1513, opened 2023-05-09
add build test.

anarkiwi:
Add two build CI tests for Ubuntu latest and Ubuntu 20.04.


bmah888 (commented):
Let's get this merged...it's possible that the test_commands.sh invocation might not work right but at least we'll know if the GitHub Actions configuration works right.

Issue #1534, opened 2023-06-29
UDP loss data not calculated correctly? 


malanmoon:
iperf 3.12 (cJSON 1.7.15)

I created an UDP session for 24h, the server output: 
```
....
[  5] 2060.00-2080.00 sec  2.33 GBytes  1.00 Gbits/sec  0.006 ms  0/1801149 (0%)
[  5] 2080.00-2100.00 sec  2.33 GBytes   999 Mbits/sec  0.006 ms  1290/1801157 (0.072%)
[  5] 2100.00-2120.00 sec  2.33 GBytes  1000 Mbits/sec  0.004 ms  0/1801116 (0%)
....
[  5] 7000.00-7020.00 sec  2.33 GBytes  1.00 Gbits/sec  0.003 ms  0/1801158 (0%)
[  5] 7020.00-7040.00 sec  2.33 GBytes  1000 Mbits/sec  0.004 ms  705/1801138 (0.039%)
[  5] 7040.00-7060.00 sec  2.26 GBytes   969 Mbits/sec  0.002 ms  55673/1801169 (3.1%)
[  5] 7060.00-7080.00 sec  2.33 GBytes  1000 Mbits/sec  0.007 ms  0/1801094 (0%)
....
[  5] 86400.00-86400.04 sec  5.24 MBytes   995 Mbits/sec  0.007 ms  0/0 (0%)
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[SUM]  0.0-86400.0 sec  1338478610 datagrams received out-of-order
[  5]   0.00-86400.04 sec  9.82 TBytes  1000 Mbits/sec  0.007 ms  0/2147483647 (0%)  receiver

```
`Lost/Total Datagrams` showed `0/2147483647 (0%)` while there is packet loss, albeit minimal.

davidBar-On:
@malanmoon, in principle, what you see is possible when UDP packets are received out of order.  E.g. if in interval x last two packets received are 90 and 100.  The report for this interval will show that 9 packets (91-99) are lost.  However, if these packets are received in interval x+1 than in total no packet is lost.  These lost packets will be counted in the "received out-of-order" in the server output.

[...]

PR #1536, opened 2023-07-02
Prevent UDP packet count and operations overflow

davidBar-On:
master

* Issues fixed (if any):
* #1534

Make all variables related to UDP packets numbers and related calculations `int64_t`.  Although `--udp-counters-64bit` options allows UDP packet number to be larger than 2^32 bits, several of these related variables were defined as `int` which is 32 nits in several environments.  Note that `int64_t` is used and not `uint64_t` to keep sign compatibility with the replaced `int` (assuming that the 63 bits limit is large enough).

In addition, `report_bw_udp/report_sum_bw_udp...` definitions in `iperf_locale.c` were changed to use `%PRIu64` instead of `%d` when printing packets counts.   Note that the related `reportCSV_bw_udp_format` was **not** changed, as I am not sure how it is used.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -85,13 +85,13 @@ struct iperf_interval_results
     float     interval_duration;
 
     /* for UDP */
-    int       interval_packet_count;
-    int       interval_outoforder_packets;
-    int       interval_cnt_error;
-    int       packet_count;
+    int64_t   interval_packet_count;
+    int64_t   interval_outoforder_packets;
+    int64_t   interval_cnt_error;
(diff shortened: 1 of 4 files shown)


bmah888:
Thanks for the PR, this looks good to me. I've done some testing on it, but not enough to get the counters past 32-bit integers. Regarding your comment about using signed 64-bit integers, I think if we were writing this code from scratch using uint64_t variables would be best, but I agree with your rationale of a least-impact change and just sticking with signed variables for now.

bmah888 (approved):

PR #1564, opened 2023-08-22
udp connect message use network endian

yihuanZhang:
base-commit:e3bdb5e27e03cf048ae584380448b358053fa2ca

when server and client run different endian machine in udp mode, then `unable to read from stream socket` return.
run client with -d, will get `Connect received for Socket 23, sz=4, buf=36373839, i=0, max_len_wait_for_reply=4`.
macro is in file `iperf.h ` like `#define UDP_CONNECT_REPLY 0x39383736`.
different endian cause iperf_udp_connect return error.


bmah888:
Thanks for the PR! I chose to adopt a slightly different solution in PR #1468.

PR #1574, opened 2023-09-08
Add support to detect packet corruption

shixinlovey1314:

```
```

* Issues fixed (if any):

In this patch, add support to detect packet corruption during test. In order to enable this feature, the receiver has to enable the new option `--integrity-check`. 

The idea is to let the sender sends out known payload, which we use 1 byte to identify the flow, and the other byte to identify the data sequence.  Receiver check the payload pattern and report corruption when detects unexpected payload.

In order to not causing huge performance dip, receiver use a worker thread to check packet, worker thread and receiver thread uses a very simple lockless ring to transfer packet.

### DIFF
--- a/configure
+++ b/configure
@@ -15099,6 +15099,67 @@ then :
 fi
 
 
+# Check for phread support
+{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for library containing pthread_create" >&5
+printf %s "checking for library containing pthread_create... " >&6; }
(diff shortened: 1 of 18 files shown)


bmah888:
Thank you for this pull request! The idea of integrity-checking the test data is an interesting one and probably most useful for UDP users. However we're uncomfortable maintaining this amount of new code for something that's not a high priority for our use cases. We'd have no objections however if you want to maintain a fork of iperf3 that includes this functionality and make that available.

Issue #1583, opened 2023-10-19
iperf3 authentication not working


cskowronnek:
# Context

* Version of iperf3: 3.15

* Hardware: VMs

* Operating system (and distribution, if any): Ubuntu 20.04

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

# Bug Report

* Expected Behavior
iperf works with auth

* Actual Behavior
iperf doesn't work

* Steps to Reproduce

Server side output:
```
user@server:~/iperf-3.15/src$ ./iperf3 -s --authorized-users-path /home/user/credentials.csv --rsa-private-key-path /home/user/private.pem -d
-----------------------------------------------------------
Server listening on 5201 (test #1)
-----------------------------------------------------------
get_parameters:
{
        "tcp":  true,
        "omit": 0,
        "time": 10,
        "num":  0,
        "blockcount":   0,
        "parallel":     1,
        "len":  131072,
        "pacing_timer": 1000,
        "authtoken":    "At25shPOYZW1T3D607ebM3ZvdDUaK1vtrIISXwedoqb72xbaSkkIJNdNN0imLIeVervc3wv3lj0rrl70jjTkfaonTP8zCGFLSPoputt7pg6OGRY4tH63YCPGdv8yIYGXBunPfUQpZ1kfYCfVaub0x6G5Ta5RGj7To/yYyQjX2hAA02B8rS8N7VyhqFPC8bwLNEU0udkWF7WaSIZGq+MeHVetj2OmGgNFzwpoTZHwz+AiFxGQxVuOL4m8vWnzh2mGU42efBUchppn8IW1xnB41RfgRNcghpY2/azxRCEsI9DDi07Qck1htbZ5zQf/b6jhuFuVuPX2xLUWQ1wsmIGRhw==",
        "client_version":       "3.15"
}
Auth Token Content:
user: mario
pwd:  rossi
ts:   1697705498
Auth Token Credentials:
--> mario rossi
Authentication failed with return code 2 for user 'mario' ts 1697705498
iperf3: error - test authorization failed
-----------------------------------------------------------
Server listening on 5201 (test #2)
-----------------------------------------------------------
```

Client side
```
user@client:~/iperf-3.15/src$ ./iperf3 -c myserver.com --rsa-public-key-path /home/user/public.pem --username mario -d
Password: send_parameters:
{
        "tcp":  true,
        "omit": 0,
        "time": 10,
        "num":  0,
        "blockcount":   0,
[...]

PR #1585, opened 2023-10-21
Validate that server authorized users file is accessible

davidBar-On:
master

* Issues fixed (if any): #1583

Add a server validation after reading the command line options, that the authorized user file is accessible for reading.  Otherwise the check is done only when a client connects to the server.

Note that the assumption is that the file exists when the server starts, although it may be modified later, as the server is reading it again for each client connection.  Without this PR change, the file may not exist when the server starts and it may be created later, but it is not clear why such usecase is useful.  If the users are not known when the server starts, an empty file can be created.


swlars (approved):
Looks good! Thank you for the pull request!

Issue #1300, opened 2022-03-19
OpenSSL 3.0 support (deprecated APIs)


rwmacleod:
Cross-compile build using poky + meta-oe

* Version of iperf3: 3.11
* Hardware: x86-64
* Operating system (and distribution, if any): Host is ubuntu-18.04.3
* Other relevant information (for example, non-default compilers,  libraries, cross-compiling, etc.):
x-compile is master branch

# Bug Report

* Expected Behavior
There are no build warnings
* Actual Behavior
Many warnings, mostly about openssl-3 deprecated APIs.

libtool: compile:  x86_64-poky-linux-gcc -m64 -march=core2 -mtune=core2 -msse3 -mfpmath=sse -fstack-protector-strong -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/ala-lpgg
p31/rmacleod/src/distro/yocto/b/up-march22/tmp/work/core2-64-poky-linux/iperf3/3.11-r0/recipe-sysroot -DHAVE_CONFIG_H -I. -I../../git/src -I/ala-lpggp31/rmacleod/src/distro/yocto/b/up-march22/tmp/work/core2-64-p
oky-linux/iperf3/3.11-r0/recipe-sysroot/usr/include -O2 -pipe -g -feliminate-unused-debug-types -fmacro-prefix-map=/ala-lpggp31/rmacleod/src/distro/yocto/b/up-march22/tmp/work/core2-64-poky-linux/iperf3/3.11-r0=
/usr/src/debug/iperf3/3.11-r0 -fdebug-prefix-map=/ala-lpggp31/rmacleod/src/distro/yocto/b/up-march22/tmp/work/core2-64-poky-linux/iperf3/3.11-r0=/usr/src/debug/iperf3/3.11-r0 -fdebug-prefix-map=/ala-lpggp31/rmac
leod/src/distro/yocto/b/up-march22/tmp/work/core2-64-poky-linux/iperf3/3.11-r0/recipe-sysroot= -fdebug-prefix-map=/ala-lpggp31/rmacleod/src/distro/yocto/b/up-march22/tmp/work/core2-64-poky-linux/iperf3/3.11-r0/r
ecipe-sysroot-native= -D_GNU_SOURCE -Wall -c ../../git/src/units.c  -fPIC -DPIC -o .libs/units.o
[...]

PR #1589, opened 2023-10-27
Bmah openssl3

bmah888:

* Issues fixed (if any): #1300 

Support OpenSSL 3.0. If OpenSSL 3.0 (or later) is detected, use newer APIs instead of OpenSSL 1 APIs that are causing deprecation warnings.

It's possible that the newer API calls would just work even on OpenSSL 1, but that possibility is not addressed in this PR.


swlars (approved):
Looks good!

Issue #444, opened 2016-07-23
easily parseable real time output option


ili101:
Hi
I made a GUI for iperf 2 in Powershell:
https://github.com/ili101/PPerf

I want to add iperf 3 support but iperf 3 don't have a parseable real time output option

iperf 2 have CSV output 
iperf 3 have JSON output but it's only updates at the end

I don't think JSON can support real time updates because of it's structure, probably why you are updating it only at the end.

Can you add CSV support or something with similar functionality?

Thank you

ili101:
I have managed to parse the text output
by cutting the data by the words position on the [ ID] line
but you have a small bug that some of the lines sometimes start before the header position

like the total and the 5.01-5.01 lines in this example

```
-----------------------------------------------------------
Server listening on 5201
-----------------------------------------------------------
Accepted connection from 127.0.0.1, port 63781
[  6] local 127.0.0.1 port 5201 connected to 127.0.0.1 port 64003
[ ID] Interval           Transfer     Bandwidth       Jitter    Lost/Total Datagrams
[  6]   0.00-1.01   sec   120 KBytes  0.98 Mbits/sec  0.136 ms  0/15 (0%)  
[  6]   1.01-2.01   sec   128 KBytes  1.05 Mbits/sec  0.137 ms  0/16 (0%)  
[  6]   2.01-3.01   sec   128 KBytes  1.05 Mbits/sec  0.169 ms  0/16 (0%)  
[  6]   3.01-4.01   sec   128 KBytes  1.05 Mbits/sec  0.126 ms  0/16 (0%)  
[  6]   4.01-5.01   sec   128 KBytes  1.04 Mbits/sec  0.113 ms  0/16 (0%)  
[  6]   5.01-5.01   sec  0.00 Bytes  0.00 Mbits/sec  0.113 ms  0/0 (0%)  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bandwidth       Jitter    Lost/Total Datagrams
[  6]   0.00-5.01   sec  0.00 Bytes  0.00 Mbits/sec  0.113 ms  0/79 (0%)  
```

bmah888:
I wonder if some of the output fields are justified in a funny way.  Admittedly I haven't looked into this at all.

If I were going to try parsing this output, I might try instead breaking the line up into "words" separated by some amount of whitespace (rather than by position on the line).  Think of the Perl or Python split commands, that's what I'm envisioning.

[...]

PR #1098, opened 2021-01-03
Implement streaming json output

kleinweby:
This takes another try at implementing "real-time" json output as requested in #444 and #923.

By specifying `--json-stream` on cli, instead of outputting a whole json document at the end multiple json "events" are emitted. These events are json objects that are unformatted (and therefore never contain a newline) and are then separated by a newline.

This can easily be parsed by simply reading until the next newline and then deserialise the read line with any json parser.

Example output from `iperf3 -c 127.0.0.1 --json-stream`:

```json
{"event":"start","data":{"connected":[{"socket":5,"local_host":"127.0.0.1","local_port":62151,"remote_host":"127.0.0.1","remote_port":5201}],"version":"iperf 3.9+","system_info":"<snip>","timestamp":{"time":"Sun, 03 Jan 2021 10:00:43 UTC","timesecs":1609668043},"connecting_to":{"host":"127.0.0.1","port":5201},"cookie":"bm5yvcnf4lhitapzhq5cbcnzppdiahy4giue","tcp_mss_default":16332,"sock_bufsize":0,"sndbuf_actual":131072,"rcvbuf_actual":131072,"test_start":{"protocol":"TCP","num_streams":1,"blksize":131072,"omit":0,"duration":10,"bytes":0,"blocks":0,"reverse":0,"tos":0}}}
{"event":"interval","data":{"streams":[{"socket":5,"start":0,"end":1.000077,"seconds":1.00007700920105,"bytes":6359613440,"bits_per_second":50872989831.698044,"omitted":false,"sender":true}],"sum":{"start":0,"end":1.000077,"seconds":1.00007700920105,"bytes":6359613440,"bits_per_second":50872989831.698044,"omitted":false,"sender":true}}}
{"event":"interval","data":{"streams":[{"socket":5,"start":1.000077,"end":2.000101,"seconds":1.0000239610672,"bytes":6450053120,"bits_per_second":51599188588.374786,"omitted":false,"sender":true}],"sum":{"start":1.000077,"end":2.000101,"seconds":1.0000239610672,"bytes":6450053120,"bits_per_second":51599188588.374786,"omitted":false,"sender":true}}}
[...]

PR #1605, opened 2023-11-22
Add `--json-immediate` option.

tom-cook-veea:

* Issues fixed (if any): This addresses #444, though possibly not as the filer intended.

This change adds a new boolean flag, `--json-immediate`.  This flag has three effects:
* It implies `--json/-J` even if it is not also set.
* It causes the JSON output to be printed without whitespace formatting, so that a complete JSON structure appears on one output line.
* It causes the JSON structure for each interval to be printed out at the end of that interval.  The usual JSON output is still printed at the end of the test.

The motivation is, as requested in #444, to provide machine-readable output in real time.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -339,6 +339,7 @@ struct iperf_test
     int       bidirectional;                    /* --bidirectional */
     int	      verbose;                          /* -V option - verbose mode */
     int	      json_output;                      /* -J option - JSON output */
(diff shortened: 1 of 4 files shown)


tom-cook-veea:
Just noticed that this has considerable overlap with #1098, which is probably a better implementation of the same idea.  It would be good to get some action on _one_ of these, so that you don't keep getting PRs that duplicate the same work.

swlars:
Just merged the other one! Thank you for the pull request!

PR #1609, opened 2023-11-30
Correct format specifier for printing int64_t

kjander0:
Version: iperf 3.16-beta1
OS: fedora 35

Negative packet loss (out of order packet) results in large positive packet loss count (percentage value is correct though).
The problem is unsigned format specifier when printing out UDP results.

```
Server listening on 5201 (test #2)
Accepted connection from 192.168.10.2, port 53474
[  5] local 192.168.10.1 port 5201 connected to 192.168.10.2 port 39361
[  5]   0.00-5.00   sec   631 KBytes  1.03 Mbits/sec  25.143 ms  2/448 (0.45%)  
[  5]   5.00-10.00  sec   641 KBytes  1.05 Mbits/sec  22.475 ms  18446744073709551614/451 (-0.44%)  
[  5]  10.00-15.00  sec   642 KBytes  1.05 Mbits/sec  24.487 ms  1/455 (0.22%)  
[  5]  15.00-20.00  sec   639 KBytes  1.05 Mbits/sec  23.750 ms  0/452 (0%)  
[  5]  20.00-25.00  sec   641 KBytes  1.05 Mbits/sec  19.466 ms  18446744073709551615/452 (-0.22%)  
[  5]  25.00-30.00  sec   641 KBytes  1.05 Mbits/sec  30.371 ms  0/453 (0%)
```

Commands:
iperf3 -s -i 5
iperf3 -c 192.168.10.1 -u -i 5

Testing:
Simulated out of order packets with netem (high jitter). Fix seems to be working fine.

### DIFF
--- a/src/iperf_locale.c
+++ b/src/iperf_locale.c
@@ -84,11 +84,11 @@
 #if defined(HAVE_INTTYPES_H)
 # include <inttypes.h>
 #else
-# ifndef PRIu64
+# ifndef PRId64
 #  if sizeof(long) == 8
-#   define PRIu64		"lu"
+#   define PRId64		"ld"
 #  else
-#   define PRIu64		"llu"
+#   define PRId64		"lld"
 #  endif
 # endif
 #endif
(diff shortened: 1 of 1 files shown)


bmah888 (approved):
Thanks for the PR! It looks good by visual inspection, and to the extent that I can test it, it seems OK. I'll merge his shortly.

Noting a couple of things for somebody's future reference:

1. There are several other (probably copy-pasted) copies of definitions like this in the iperf3 sources. It'd be great if we could get them into a single header (maybe `iperf.h`? Not sure if this is feasible or not given how some source files include headers)
2. Some of the compatiblity goop does definitions for `PRIu64` and some for `PRId64`. We should probably standardize on a single version that does both.
3. `src/iperf_auth.c` contains an unconditional inclusion of `<inttypes.h>`, which will break if someone actually tries to compile with this feature.

PR #1612, opened 2023-12-03
Check and link libatomic if needed

### LINKED ISSUE #1611: Undefined symbols in `libiperf` when building for ARMv6
# Context

* Version of iperf3: `3.16`

* Hardware: `ARMv6`

* Operating system (and distribution, if any): PLD Linux

# Bug Report

* Expected Behavior

`libiperf` does not have unresolved symbols.

* Actual Behavior

`libiperf` has unresolved symbols related to 64-bit atomics:
```
	__atomic_store_8
	__atomic_load_8
	__atomic_fetch_add_8
```

* Steps to Reproduce

1. Compile `iperf` targetting `ARMv6`

* Possible Solution

Link `libatomic`

jpalus:
* Version of iperf3 to which this pull request applies: `3.16`

* Issues fixed (if any): Fixes #1611

Some architectures without native support for 64-bit atomics need linking with libatomic.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,7 +92,19 @@ CXX="$PTHREAD_CXX"
 ])
 
 # Atomics
-AC_CHECK_HEADERS([stdatomic.h])
+AC_CHECK_HEADERS([stdatomic.h],
+    [AC_MSG_CHECKING([whether libatomic is required])
+    AC_LINK_IFELSE([AC_LANG_PROGRAM([[#include <stdatomic.h>]], [[atomic_uint_fast64_t i; i++;]])],
+        [AC_MSG_RESULT([no])],
+        [save_LIBS="$LIBS"
+        LIBS="$LIBS -latomic"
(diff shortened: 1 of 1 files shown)


bmah888:
I haven't really looked at this in detail yet but thanks for sending a PR that _only_ includes the `configure.ac` change, that's exactly what we need. When we merge it I'll regenerate `configure` after that.

bmah888 (approved):
I was able to confirm that this change doesn't cause any regression with our officially supported platforms. I don't think we have immediate access to a platform where `libatomic` is required, but I figure you've tested that part. Merging shortly and then will regenerate the `configure` file(s). Thanks again for the PR!

jpalus:
> I was able to confirm that this change doesn't cause any regression with our officially supported platforms. I don't think we have immediate access to a platform where `libatomic` is required, but I figure you've tested that part. Merging shortly and then will regenerate the `configure` file(s). Thanks again for the PR!

Just in case you'd like to test it for yourself and you have access to x86 32-bit environment please try to compile with either `-march=i386` or `-march=i486`. They both lack support for 64-bit atomics and hence issue should be the same there. x86 gained support for 64-bit atomics with i586.

PR #1615, opened 2023-12-06
Add MPTCP support with the --multipath flag

marcosfsch:
  master

* Issues fixed (if any):
Rebase of PR [#1166](https://github.com/esnet/iperf/pull/1166), including testing for MPTCP support.

Recent version of the Linux kernel (5.9) gained MPTCP support, this change
allows easy testing such protocol.
MPTCP is strongly tied to TCP, and the kernel APIs are almost the same,
so this change does not implement a new 'struct protocol'. Instead it just
extend the TCP support to optionally enable multipath via the --multipath or
-m switch.

The only required dependency is the 'IPPROTO_MPTCP' protocol number
definition, which should be provided by the netinet/in.h header.
To keep things simple, just conditionally provide the required
protocol, if the system header does not have it yet.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -325,6 +325,18 @@ if test "x$iperf3_cv_header_tcp_info_snd_wnd" = "xyes"; then
   AC_DEFINE([HAVE_TCP_INFO_SND_WND], [1], [Have tcpi_snd_wnd field in tcp_info.])
 fi
 
+# Check for IPPROTO_MPTCP support
+AC_CACHE_CHECK([IPPROTO_MPTCP support],
+[iperf3_cv_header_ipproto_mptcp],
+AC_COMPILE_IFELSE(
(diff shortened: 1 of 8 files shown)


matttbe:
@marcosfsch thank you for this rebase. @geliangtang did another one in #1661 with some small modifications discussed with the MPTCP devs:
- no build check: better to check at run-time, the kernel could be different
- `--mptcp` (shorter) instead of `--multipath` (too vague, could be multipath QUIC, etc.)
- place the new `proto` arg after `type` in the modified function signatures.

Are the modifications OK for you? Would it be OK to close this PR?
Hopefully IPerf3 devs will be able to review #1661 soon :)

marcosfsch:
I'm ok with closing this PR, thanks for the rebase and improvements.

Em qua., 19 de jun. de 2024 05:16, Matthieu Baerts ***@***.***>
escreveu:

PR #1619, opened 2023-12-12
Add error handling for close socket

maks-mishin:
* Development branch: master 

* Issues fixed: add error handling for correct closing of socket `s` because of return with close socket leads to memory leak. 

* Brief description of code changes: added `goto` label with correct exit from function if necessary.


ProjectMutilation:
@maks-mishin this isn't correct, mark of error_handling must be after return 0, now the function always return -1
probably you should added variable int ret and initialize it 0, and in error cases 
```
    ret = -1;
    goto error_handling;
    ...
    error_handling:
        close(s);
        return ret;
 ```

File: src/iperf_server_api.c [hunk 1]
```diff
@@ -174,18 +174,18 @@ iperf_accept(struct iperf_test *test)
...
+            goto error_handling;
         }
 	FD_SET(test->ctrl_sck, &test->read_set);
 	if (test->ctrl_sck > test->max_fd) test->max_fd = test->ctrl_sck;
 
 	if (iperf_set_send_state(test, PARAM_EXCHANGE) != 0)
-            return -1;
-        if (iperf_exchange_parameters(test) < 0)
-            return -1;
+        goto error_handling;
```
ProjectMutilation, line 183: incorrect indent

src/iperf_server_api.c
```diff
@@ -174,18 +174,18 @@ iperf_accept(struct iperf_test *test)
...
+    if (iperf_exchange_parameters(test) < 0)
```
ProjectMutilation, line 184: the same

src/iperf_server_api.c
```diff
@@ -174,18 +174,18 @@ iperf_accept(struct iperf_test *test)
...
+        goto error_handling;
 	if (test->server_affinity != -1)
 	    if (iperf_setaffinity(test, test->server_affinity) != 0)
-		return -1;
+		    goto error_handling;
```
ProjectMutilation, line 188: extra indent

maks-mishin:
@ProjectMutilation, thanks for comments and advice.
I will fix it in the next commit.

bmah888 (approved):
Thanks for the PR and comments! Plugging this file descriptor leak is definitely a good thing. I'm thinking about the use of the variable `ret`, it's only defined in one place (and only ever initialized to `-1`), and we only reference the value of `ret` in the two error exit cases. I'm not sure that's much better than just returning with the numeric literal `-1`.

I'm going to approve this for merging as-is, and going to sit on it a little bit to see if we want to iterate this a bit or not. The code change is definitely an improvement.

PR #1622, opened 2023-12-14
Fix memory leak with addrinfo

maks-mishin:
* Development branch: master

* Issues fixed: fix memory leak 

* Brief description of code changes: now structure `server_res` is free before return from function.


bmah888 (approved):
Thanks for the PR, good catch!

PR #1623, opened 2023-12-14
Fix descriptor leak in some return paths

maks-mishin:
* Development branch: master

* Issues fixed: descriptor leak

* Brief description of code changes: close socket `s` in some cases when force return from function.


bmah888 (approved):
Thanks for the PR, approved!

PR #1636, opened 2024-01-20
Define SIZE_OF_LONG_INT to replace sizeof(long) in #if

davidBar-On:
master

* Issues fixed (if any):

Define SIZE_OF_LONG_INT by `./config.ac` with the value of `sizeof(long)` (required running  ./bootstrap; ./configuration).  This is to fix [the issue](https://github.com/esnet/iperf/commit/ff458b2e736ad301ae82624f08742165255671bd#r136874913) identified by @yumkam that `sizeof()` is not valid for the preprocessor in `#if`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -180,6 +180,16 @@ else
     fi
 fi
 
+# Get the size of long int
+AC_CACHE_CHECK([Size of long int],
+[iperf3_cv_sizeof_long],
+AC_RUN_IFELSE(
+  [AC_LANG_PROGRAM([[#include <stdio.h>]],
+                   [[fprintf(stderr, "%d", sizeof(long)); return(0)]])],
+  [iperf3_cv_sizeof_long=`./conftest$EXEEXT 2>&1`],
+  [AC_MSG_ERROR(Failed to compile and run size of long test)]))
+AC_DEFINE_UNQUOTED([SIZE_OF_LONG_INT], $iperf3_cv_sizeof_long, [Size of long int.])
+
 # Check for TCP_CONGESTION sockopt (believed to be Linux and FreeBSD only)
 AC_CACHE_CHECK([TCP_CONGESTION socket option],
 [iperf3_cv_header_tcp_congestion],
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -180,6 +180,16 @@ else
     fi
 fi
 
+# Get the size of long int
+AC_CACHE_CHECK([Size of long int],
+[iperf3_cv_sizeof_long],
+AC_RUN_IFELSE(
+  [AC_LANG_PROGRAM([[#include <stdio.h>]],
+                   [[fprintf(stderr, "%d", sizeof(long)); return(0)]])],
```
bmah888, line 188: Thanks for the pull request! I got a compiler warning with macOS / Clang on this line (line 188) where it complained of a mismatch between the `%d` format specifier (which takes an int) and `sizeof(long)`, which is a long. If I changed the `%d` to `%zd` then this seems to work correctly.

bmah888 (changes requested):
One minor change requested, thanks for the PR!

After this gets merged, I'll regenerate the top-level `configure` file myself.

configure.ac
yumkam: That's all kinda pointless.
`<inttypes.h>` is in c99, c99 is supported everywhere (that's why issue gone unnoticed for 10 years), and `"%zd"` is from c99 too. Besides, `AC_RUN_IFELSE` breaks on cross-compilation (you need check that fails on compilation stage), etc.
Just use `<inttypes.h>' unconditionally and be done.

davidBar-On:
Changed "%d" to "%zd".  However, I tend to agree with @yumkam [comment](https://github.com/esnet/iperf/pull/1636#discussion_r1491827219) that just using `<inttypes.h>` unconditionally is the right approach.

bmah888:
OK I'm just going to close this PR. I've got a code branch that gets rid of a bunch of this conditional compilation stuff, I'm going to push that branch and make a PR for it.

PR #1640, opened 2024-02-03
Make use of arc4random_buf()

brad0:
Make use of arc4random_buf() if it exists. Supported on
FreeBSD/NetBSD/OpenBSD/Solaris/Linux/Android.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -241,6 +241,9 @@ AC_CHECK_FUNCS([sendfile])
 # connections.
 AC_CHECK_FUNCS([getline])
 
+# Check for arc4random_buf().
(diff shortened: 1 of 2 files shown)


brad0:
Ping.

swlars:
Thank you for the pull request! Considering the readentropy() function is only used to randomize the payload in iperf3 streams to avoid compression, I'm not sure what benefit using arc4random_buf() would bring. Could you explain the advantages?

swlars:
Thank you for the pull request. If you'd like to discuss the benefits of this change, we can revisit it in the future.

PR #1649, opened 2024-02-08
minor error message correction for openssl includes

dopheide-esnet:
* Version of iperf3:
master

* Issues fixed (if any):
minor confusion during build debugging

Just matching the error message to what the test is actually doing.


bmah888 (approved):
Looks good. I'm going to quash the two commits in the PR together and then merge that. Thanks!

Issue #1641, opened 2024-02-04
Usage of pthread_cancel makes Android build fail


abraaocaldas:
# Context
Because of a undeclared function 'pthread_cancel' introduced at: https://github.com/esnet/iperf/commit/2e80758ad49dfaeae3970d09227e3b28e412f06c

* Version of iperf3:
Iperf 3.16

* Hardware:
N/A

* Operating system (and distribution, if any):
Android 14

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):
Build using NDK or native compiling inside device using Termux

# Bug Report

* Expected Behavior
Build is done correctly

* Actual Behavior
Build fail with message:
`error: call to undeclared function 'pthread_cancel'; ISO C99 and later do not support implicit function declarations`

* Steps to Reproduce
 Inside termux clone repository on latest commit or at least `2e80758ad49dfaeae3970d09227e3b28e412f06c`
./configure
make

* Possible Solution
---

abraaocaldas:
Hello, It seems stable with the #1651 , compiling and tests are runnning fine.

PR #1651, opened 2024-02-18
Add pthread missing functions in Android

davidBar-On:
master

* Issues fixed (if any): #1641

Add the related pthread functions that are missing in Android.  This is instead of the current practice to makes changes to the source files before building for Android  (see for example [here](https://github.com/davidBar-On/android-iperf3), which is the source for the changes).

The changes should not impact a build for any other OS, as the added/modified code in under `#ifdef __ANDROID__`.

Note that the PR includes two new files, so configure (and bootstrap?) is needed (this is why the automatic checks fail).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -33,6 +33,8 @@ libiperf_la_SOURCES     = \
                         iperf_util.h \
                         iperf_time.c \
                         iperf_time.h \
+                        iperf_pthread.c \
+                        iperf_pthread.h \
 			dscp.c \
                         net.c \
                         net.h \
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -74,9 +74,7 @@
 #include <openssl/evp.h>
 #endif // HAVE_SSL
 
-#ifdef HAVE_PTHREAD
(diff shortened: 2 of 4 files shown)


bmah888 (commented):
Thanks! This looks good to me in general. Have you (or someone else) tested whether this actually works correctly on an Android host? (I have no way to test this, but it doesn't seem to cause regression on the officially supported platforms, which is good.)

bmah888:
> > Have you (or someone else) tested whether this actually works correctly on an Android host?
> 
> I did a basic test using Android simulator and it is working.

Great, thanks!

bmah888 (approved):
Thanks for the PR! I'm hoping to merge this soon and then regenerate the automake/autoconf files.

hajoha:
> Thanks! This looks good to me in general. Have you (or someone else) tested whether this actually works correctly on an Android host? (I have no way to test this, but it doesn't seem to cause regression on the officially supported platforms, which is good.)

Yes we integrated that changes into our app with iPerf3.16 and its working, thanks!

PR #1655, opened 2024-02-29
Assume we always have inttypes.h and stdint.h (both from C99)

bmah888:
This eliminates some compile-time tests that didn't really work as desired and aren't easy to fix. It also removes conditional compilation in a number of places.

Note that there were already instances of unconditionally including `<inttypes.h>` and `<stdint.h>` in the iperf3 source code, so in practice we had somehow already made this assumption.

Inspired by comments on PR #1636.

* Issues fixed (if any):

### DIFF
--- a/examples/mic.c
+++ b/examples/mic.c
@@ -4,9 +4,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <string.h>
-#ifdef HAVE_STDINT_H
 #include <stdint.h>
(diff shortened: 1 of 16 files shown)


bmah888:
Tagging @davidBar-On and @yumkam in case they have any comments.

swlars (approved):
Looks good! We might want to take a closer look at used/unused includes at some point.

PR #1661, opened 2024-03-12
add MPTCPv1 support

geliangtang:
The Multipath TCP (MPTCP) protocol (v1 / RFC 8684) has been added in the upstream Linux kernel since v5.6.

MPTCP is strongly tied to TCP, and the kernel APIs are almost the same. The only required dependency is the 'IPPROTO_MPTCP' protocol number definition, which should be provided by the netinet/in.h header if it is recent enough.

This patch adds a new flag '-m' or '--mptcp' to support MPTCPv1. It can be used like this:

 > iperf3 -m -s
 > iperf3 -m -c 127.0.0.1

There is no need to check for IPPROTO_MPTCP support in configure.ac at build time, it is at runtime we will see if the kernel being use supports or not MPTCP.

If IPPROTO_MPTCP is not supported by the kernel being tested, it is normal to fail because the feature is not available and the user explicitly asked to use MPTCP.

Closes: https://github.com/esnet/iperf/pull/1659
Co-developed-by: Paolo Abeni <pabeni@redhat.com>

```
```

* Issues fixed (if any):

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -337,6 +337,18 @@ if test "x$iperf3_cv_header_tcp_info_snd_wnd" = "xyes"; then
   AC_DEFINE([HAVE_TCP_INFO_SND_WND], [1], [Have tcpi_snd_wnd field in tcp_info.])
 fi
 
+# Check for IPPROTO_MPTCP (Linux)
+AC_CACHE_CHECK([MPTCP protocol],
+[iperf3_cv_header_ipproto_mptcp],
+AC_COMPILE_IFELSE(
+  [AC_LANG_PROGRAM([[#include <netinet/in.h>]],
+		   [[int foo = IPPROTO_MPTCP;]])],
+  iperf3_cv_header_ipproto_mptcp=yes,
+  iperf3_cv_header_ipproto_mptcp=no))
+if test "x$iperf3_cv_header_ipproto_mptcp" = "xyes"; then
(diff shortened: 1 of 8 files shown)


sskras:
Any news about adding the support?

Currently I am trying to launch my MPTCP research, and not having to run `iperf3` with the `mptcpize` wrapper script would be a clean way of going forward.

matttbe:
@bmah888: sorry for the ping. Do you think such "smallish" modification could be accepted in IPerf3? No hurry, but it is just to know which direction we take regarding tests in the Linux kernel :)

matttbe (approved):
@geliangtang thank you for the rebase and the last modifications!

The check for `IPPROTO_MPTCP` at build time should not be needed, but good to have added it to be consistent with how the other kernel constants are being used/checked here. Hopefully this new version can be accepted :)

bmah888:
Thanks @geliangtang for the recent changes and @matttbe for the out-of-band emails! I've just started a CI run. At this point I think we just need to do a little testing for our own due diligence and then this PR is probably good to merge.

bmah888 (approved):
Thanks again for the PR! This looks good to the extent that I can tell. We still don't have a good environment within ESnet for testing MPTCP (also no expertise with it because it's not a part of our use cases), but I've at least verified that there are no regressions in iperf3 behavior. If there's any fixes or improvements to be made, we can take care of that later. So...let's get this merged!

Issue #1662, opened 2024-03-12
Client parameter `-t -1` has unexpected results


ed-at-protectli:
# Context
When I was trying to run iperf3 tests for an indefinite amount of time, I tried the client parameter `-t -1` instead of the supported and documented `-t 0`. Because the tests "ran" and the CLI tool didn't complain about the input, I assumed this was "ok". Furthermore, the test indeed seemed to run "indefinitely" instead of the standard 10. However, the application quits with failure after precisely 38 seconds on my test hardware.

* Version of iperf3:
iperf 3.9 (cJSON 1.7.13)
from apt package version 3.9-1+deb11u1build0.22.04.1 for Ubuntu 22.04

* Hardware:
Custom hardware device, using i226-V NICs and an Intel Celeron.

* Operating system (and distribution, if any):
Ubuntu 22.04 Desktop

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):
Installed from apt repo on a vanilla Ubuntu install.

# Bug Report

* Expected Behavior

The test would either run indefinitely, or the application would reject the `-t -1` parameter.

* Actual Behavior

The test initiates, but will eventually crash.

* Steps to Reproduce

1. I simply connect together two NICs (enp1s0 and enp2s0) on the device with an ethernet cable (not going through a switch)
2. Namespace the NICs so that the data actually flows over the cable and not shortcutting through the kernel:
```
# In one terminal for the iperf server:
ip net add iperfserver
ip link set enp1s0 netns iperfserver
ip netns exec iperfserver ip address add 169.254.1.1/20 dev enp1s0
ip netns exec iperfserver ip link set enp1s0 up
ip netns exec iperfserver iperf3 -s -B 169.254.1.1

# In another terminal for the iperf client:
ip net add iperfclient
ip link set enp2s0 netns iperfclient
ip netns exec iperfclient ip address add 169.254.1.2/20 dev enp2s0
ip netns exec iperfclient ip link set enp2s0 up
```
3. On the client terminal, run the iperf3 command like so, using `-t -1` or `-t=-1` or `--time -1` or `--time=-1`:
```
ip netns exec iperfclient iperf3 -c 169.254.1.1 -t -1
```
4. Observe the results in the server terminal after (in my case) 38 seconds
```
[  5]  37.00-38.00  sec   235 Mbytes  1.97 Gbits/sec
iperf3: error - select failed: Bad file descriptor
----------------------------------------
Server listening on 5201
[...]

PR #1666, opened 2024-03-14
Not allow negative test duration

davidBar-On:
master
* Issues fixed (if any): #1662

Make the test `duration` unsigned, to prevent the ability to set negative duration values.

Alternative, and maybe more standard, solution may be to change `duration` and the variables in `struct iperf_time` to `int64_t` (signed, instead of `uint32_t`), which is compatible with `struct timeval`.  I therefore added a check that the `period` option is not negative, although this is redundant when it is unsigned.

The issue itself was caused because of the following in `create_server_timers()`:
```c
    int max_rtt = 4; /* seconds */
    int state_transitions = 10; /* number of state transitions in iperf3 */
    .....
        test->timer = tmr_create(&now, server_timer_proc, cd, (test->duration + test->omit + grace_period) * SEC_TO_US, 0);
```
When period was -1 to -40, the `test->duration + test->omit + grace_period` calculation overflowed and caused the server to call `server_timer_proc()` within 0  to 39 seconds (depending on `duration`) and close all sockets to the client.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1313,7 +1313,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 	        break;
             case 't':
                 test->duration = atoi(optarg);
-                if (test->duration > MAX_TIME) {
+                if (test->duration > MAX_TIME || test->duration < 0) {
                     i_errno = IEDURATION;
                     return -1;
                 }
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -166,7 +166,7 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "some option you are trying to set is client only");
             break;
         case IEDURATION:
-            snprintf(errstr, len, "test duration too long (maximum = %d seconds)", MAX_TIME);
+            snprintf(errstr, len, "test duration valid values are 0 to %d seconds", MAX_TIME);
             break;
         case IENUMSTREAMS:
             snprintf(errstr, len, "number of parallel streams too large (maximum = %d)", MAX_STREAMS);


TheRealDJ:
@davidBar-On I had the same checks issue. I merged Bruce's latest master commits to my branch, and the checks now pass. Failures appear to have been due to the fqrate not being reset between tests (iperf_reset_test).

davidBar-On:
@TheRealDJ, thanks.  I rebased to the latest master and the checks passed.

bmah888:
> @davidBar-On I had the same checks issue. I merged Bruce's latest master commits to my branch, and the checks now pass. Failures appear to have been due to the fqrate not being reset between tests (iperf_reset_test).

Yeah sorry about that...a little while back we fixed an issue with `--fq-rate` not working with `--reverse` tests, but that uncovered this other problem, which made the tests run too long and time-out in the Github actions framework.

src/iperf.h
```diff
@@ -294,7 +294,7 @@ struct iperf_test
     int       bind_port;                        /* --cport option */
     int       server_port;
     int       omit;                             /* duration of omit period (-O flag) */
-    int       duration;                         /* total duration of test (-t flag) */
+    uint32_t  duration;                         /* total duration of test (-t flag) */
```
bmah888, line 297: I think that `duration` can and should remain an `int`, right? You check in line +1316 below should make this value safe...in fact if you change `duration` to be unsigned, then `test->duration < 0` below will never be true. Also the return value of `atoi()` is an `int`, not an `unsigned int`.

bmah888 (changes requested):
Thanks for the PR! Just one change to request and we should be good.

src/iperf.h
davidBar-On: Agree `duration` should be changed back to signed.  I didn't take into account that atoi() returns int ... I first changed to unsigned `duration` but than added the `test->duration < 0` check to ensure that if it will be changed back to signed, negative values will still fail.

Changed now back to signed, but to `int64_t` instead of `int` and `atol()` instead of `atoi()`.  This is because for a very large absolute value, e.g. `-99999999999999999999999`, atoi() returns 0 which is the failure value, but 0 is a valid timeout.  The change does not solve the issue, but much larger absolute values should be used to cause it.

If the above change is an overkill, and `int` / `atoi()` should be used, please let me know.

bmah888: I think I'd like to stick to `int` / `atoi()`. I'm not sure if `int64_t` and `atol()` are always comparable without conversions, and if someone needs a test to run for longer than 2^(32-1) seconds well, maybe they might need to rethink that a bit.

davidBar-On: O.k Changed back to `int`/ `atoi()` (with rebase).

bmah888: Thanks!

bmah888 (approved):
Thanks! Will merge this shortly.

PR #1667, opened 2024-03-14
Properly reset the --fq-rate parameter on the server between tests.

bmah888:
Without this change, an --fq-rate setting would persist on the server, which could adversely slow down future --reverse tests. This bug was exposed by PR #1643, which allows --fq-rate to work on the server. One annoying side-effect of this bug was that GitHub Actions scripts were timing out and throwing errors.

Pet file copyright date while here.

* Issues fixed (if any): None

* Brief description of code changes (suitable for use as a commit message): See above.


swlars (approved):
Looks good!

PR #1670, opened 2024-03-19
iperf_api: memset entire malloc

RABijl:

```
```

* Issues fixed (if any):

in the function iperf_new_test
the  bitrate_limit_intervals_traffic_bytes array
was only memset for the size of the sizeof
return type, instead of the entire array.

while we're on this topic, is there are reason why calloc is not used instead of malloc + memset?


bmah888:
Thanks for the PR! Yeah that's a pretty silly bug isn't it.

PR #1672, opened 2024-03-22
Don't chdir when becoming a daemon.

bmah888:
This fixes some non-intuitive behavior when using the iperf3 authentication feature, where iperf3 was able to use a relative path to locate the credentials file when being run "normally" but not if it was being run as a --daemon (the workaround was to use only absolute pathname arguments).

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -154,7 +154,7 @@ run(struct iperf_test *test)
         case 's':
 	    if (test->daemon) {
 		int rc;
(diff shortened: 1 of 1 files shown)


dopheide-esnet (commented):
Yup, that's exactly what I did and then verified it worked.

bmah888:
> Yup, that's exactly what I did and then verified it worked.

Cool, thanks for confirming / testing!

Issue #1095, opened 2020-12-16
Bandwidth performance tool for chisel tunnel based on socks5


ivanv6:
Hi, I developed a tunnel system based on chisel which exposes on client-side, a SOCKS5 proxy to allow users to perform requests. Moreover, I'm interested to test the bandwidth performance of the tunnel system. There is a way to do this performance test by using iperf through a socks5? Thanks a lot

bmah888:
As long as your proxy presents a TCP port to bind to, it should be possible to run iperf3 through it. You would tell the iperf3 client to connect to the proxy (use the `-c` and `-p` options to point to the SOCKS proxy host and port). I think this would work for the most common iperf3 use case of doing TCP tests from the client to server. I'm not sure how UDP or reverse-mode tests would work (it doesn't feel to me like either of this would work well, but I don't really have any first-hand experience with this sort of thing).

ivanv6:
dear @bmah888 thanks a lot for your response, I'm, trying to test the iperf3 tool with your suggestions. Another question, there is a way to run iperf3 always on socks5 also if the socks5 requires username and password to accept the request? In a few words, the socks5 requires credentials.

bmah888:
It needs credentials? Yikes that's hard. So there must be some other API other than the standard socket API then?

(What I mean is that there are more parameters/credentials needed than what the normal socket API needs to be able to transfer data, then a program like iperf3 would need to have some explicit SOCKS support. It doesn't currently. My knowledge in this area is really hazy though.)

BigSully:
Another option is to run `iperf3` client with `proxychains`, which proxies traffics via socks5 tunnel to the remote iperf3 server.

mister-goo:
Added socks5 client in my fork, only TCP protocol is implemented. https://github.com/mister-goo/iperf/commit/532d483b37a070f15cf96dc0370ed7aa63bc26b1

rezapoorzare:
I have one more question:
Is it possible to use iperf3 through socks5? Does it support options such as --socks5?
Thanks a lot

davidBar-On:
> Is it possible to use iperf3 through socks5? Does it support options such as --socks5?

[...]

PR #1676, opened 2024-03-26
Add SOCKS5 Proxy support for TCP

davidBar-On:
master

* Issues fixed (if any): #1095 (partly)

 SOCKS5 Proxy support for TCP.  `--socks5` option is added. Based on @mister-goo code suggested [here](https://github.com/esnet/iperf/issues/1095#issuecomment-1146514695).

The implementation is standalone, so there should be no change to functionality when the `--socks5` is not used.

UDP is not supported since it requires added a header to each UDP packet sent.  Since this header has to be taken into consideration in the iperf3 packet size (`-l`) it make the implementation more complex.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -343,6 +343,14 @@ struct iperf_test
     int       timestamps;			/* --timestamps */
     char     *timestamp_format;
 
+    char     *socks5_host;                      /* --socks5 option */
+    uint16_t  socks5_port;                      /* --socks5 option optional value */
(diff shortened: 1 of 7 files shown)


bmah888:
Thanks for the pull request! SOCKS5 support is not something that we've seen as a requirement in the R&E networking space, so it's unfortunately a bit out of scope for iperf3's core functionality. We're also not able to test or maintain this functionality in our environment. This might be a good candidate for someone to integrate to a fork of iperf3.

sergeevabc:
We need this!

Issue #1631, opened 2024-01-09
Server side JSON problem


zokl:
* Version of iperf3: 3.16+ from 4.1.2024
* Hardware: x86_64
* Operating system (and distribution, if any): Linux, OpenWRT x86_64

```
root@cli1:~# iperf3 -v
iperf 3.16+ (cJSON 1.7.15)
Linux cli1 5.10.201 #0 SMP Thu Jan 4 21:29:46 2024 x86_64
Optional features available: CPU affinity setting, IPv6 flow label, TCP congestion algorithm setting, sendfile / zerocopy, socket pacing, bind to device, support IPv4 don't fragment, POSIX threads
```

# Bug Report

* Expected Behavior

Server side report will be without unformated messages.

* Actual Behavior

Server side structure is valid but at the and is unformated message : `iperf3: error - unable to write to stream socket: Invalid argument` or `iperf3: error - no error`. This error is very similar to this closed issue: https://github.com/esnet/iperf/issues/1143 but on server side (@davidBar-On ).

```
		"sum_sent":	{
			"start":	0,
			"end":	90.000614,
			"seconds":	90.000614,
			"bytes":	10557718528,
			"bits_per_second":	938457466.790171,
			"retransmits":	1,
			"sender":	true
		},
		"sum_received":	{
			"start":	0,
			"end":	90.000614,
			"seconds":	90.000614,
			"bytes":	0,
			"bits_per_second":	0,
			"sender":	true
		},
		"cpu_utilization_percent":	{
			"host_total":	36.1468918421911,
			"host_user":	0.734266323616089,
			"host_system":	35.412626629677185,
			"remote_total":	0,
			"remote_user":	0,
			"remote_system":	0
		},
		"sender_tcp_congestion":	"cubic"
	}
}
iperf3: error - unable to write to stream socket: Invalid argument
```

or sometime:

  ```
              "sum_received": {
                        "start":        0,
                        "end":  10.00105,
                        "seconds":      10.00105,
                        "bytes":        7484473344,
                        "bits_per_second":      5986950045.44523,
                        "sender":       false
                },
                "cpu_utilization_percent":      {
                        "host_total":   85.012575052611453,
                        "host_user":    2.79302852241707,
                        "host_system":  82.219516533647,
                        "remote_total": 0,
                        "remote_user":  0,
                        "remote_system":        0
                },
                "receiver_tcp_congestion":      "bbr"
        }
}
iperf3: error - no error
```

[...]

PR #1677, opened 2024-03-27
Prevent server termination with JSON and remove redundant code with error

davidBar-On:
master

* Issues fixed (if any): #1631

Fix to two issues when JSON is used, one of them caused **server termination**:

1. `iperf_errexit()` locked the mutex, but called `iperf_json_finish()` that also lock/unlock the mutex, so when the client terminated before the end of the test, the mutex unlock failed and the server terminated.
2. Removed a redundant code, which was probably forgotten after copying it some lines above.  The redundant code caused redundant client error when it terminated.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4889,16 +4889,7 @@ iperf_json_finish(struct iperf_test *test)
         cJSON_Delete(test->json_top);
         test->json_top = NULL;
     }
-    // Get ASCII rendering of JSON structure.  Then make our
-    // own copy of it and return the storage that cJSON allocated
-    // on our behalf.  We keep our own copy around.
-    char *str = cJSON_Print(test->json_top);
-    if (str == NULL)
-	return -1;
-    test->json_output_string = strdup(str);
-    cJSON_free(str);
-    if (test->json_output_string == NULL)
-        return -1;
+
     if (test->json_stream) {
         cJSON *error = cJSON_GetObjectItem(test->json_top, "error");
         if (error) {
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -94,10 +94,6 @@ iperf_errexit(struct iperf_test *test, const char *format, ...)
     struct tm *ltm = NULL;
     char *ct = NULL;
 
-    if (pthread_mutex_lock(&(test->print_mutex)) != 0) {
-        perror("iperf_errexit: pthread_mutex_lock");
-    }
-
     /* Timestamp if requested */
     if (test != NULL && test->timestamps) {
 	time(&now);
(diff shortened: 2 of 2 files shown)


bmah888:
Thanks for the PR! Initial tests with your patch are looking good so far, but I want to go back through the code (along with your commentary at the top of this PR) to understand how we (I) managed to mess this up without noticing. :-p

src/iperf_api.c
```diff
@@ -4889,16 +4889,7 @@ iperf_json_finish(struct iperf_test *test)
...
-    // own copy of it and return the storage that cJSON allocated
-    // on our behalf.  We keep our own copy around.
-    char *str = cJSON_Print(test->json_top);
-    if (str == NULL)
-	return -1;
-    test->json_output_string = strdup(str);
-    cJSON_free(str);
-    if (test->json_output_string == NULL)
-        return -1;
+
```
bmah888, line 4892: This duplicated code, which appears to have come when we merged in PR #1098, is responsible for the error exit (because `test->json_top` may have been cleared in line 4890). So deleting this is good.

src/iperf_error.c
```diff
@@ -113,7 +109,11 @@ iperf_errexit(struct iperf_test *test, const char *format, ...)
 	    cJSON_AddStringToObject(test->json_top, "error", str);
         }
 	iperf_json_finish(test);
-    } else
+    } else {
+        if (pthread_mutex_unlock(&(test->print_mutex)) != 0) {
+            perror("iperf_errexit: pthread_mutex_unlock");
```
bmah888, line 114: This code should be locking the mutex, not unlocking it, because we're trying to protect the fprintf() calls in the next few lines.

src/iperf_error.c
```diff
@@ -127,8 +127,9 @@ iperf_errexit(struct iperf_test *test, const char *format, ...)
 	    fprintf(stderr, "iperf3: %s\n", str);
 	}
 
-    if (pthread_mutex_unlock(&(test->print_mutex)) != 0) {
-        perror("iperf_errexit: pthread_mutex_unlock");
+        if (pthread_mutex_lock(&(test->print_mutex)) != 0) {
+            perror("iperf_errexit: pthread_mutex_lock");
```
bmah888, line 131: And then here we want to unlock the mutex, not lock it, before we exit the function.

src/iperf_error.c
```diff
@@ -94,10 +94,6 @@ iperf_errexit(struct iperf_test *test, const char *format, ...)
     struct tm *ltm = NULL;
     char *ct = NULL;
 
-    if (pthread_mutex_lock(&(test->print_mutex)) != 0) {
```
bmah888, line 97: OK you want to move the mutex operation below the (potential) call to `iperf_json_finish()`. This is good.

bmah888 (changes requested):
Thanks again for the PR! OK, I understand the original problems now. One change I think we need is to interchange the order of the mutex lock/unlock operations that you just moved/added. I think fixing these bugs uncovered some other problems (locking in the `--json-stream` codepaths and duplicated JSON output in the non `--json-stream` output codepath). I'm going to try to figure out fixes for these, not sure if I want to try to add to this PR or create a new PR for them.

src/iperf_error.c
davidBar-On: Oops! At least I hope that I mean to lock here.  Changed.

src/iperf_error.c
davidBar-On: Changed

davidBar-On:
I switched between the lock and unlock, but if other related changes are needed, then I agree it is better to have a new PR (I don't have a problem with this).

bmah888 (commented):
Thanks, OK I think this PR is good to go now! I'm hoping to have a fix for the other two problems shortly.

Issue #615, opened 2017-07-20
Server-side limiting switches


mfeit-internet2:
In perfsonar/pscheduler#518, I opined that bandwidth curbs imposed by the limits in perfSONAR are on the honor system because the destination (iperf3 server-side) participant has to trust that the source (client-side) will throttle itself to the agreed-upon maximum speed.

I'd like to propose the addition/re-use of these switches for the server mode as a way for the pScheduler iperf3 plugin to enforce some of the limits 

  * `-b` - Forcibly terminates the test if the measured bandwidth during any interval exceeds the specified value.  (I'd also be interested in hearing your feedback on what kind of measures could or should be taken to soften this in light of the fact that every so often the results contain an outlier.)

  * `-t` - Forcibly terminates the test if it runs longer than the specified time.

  * `-u` - Allows the client to request UDP.

While not a perfect solution, this would give the receiving end a way to end its involvement if it thinks the client is holding up its end of the bargain.

bmah888:
OK, I'm pondering this.  I think I understand your concern for protecting the server process.

You're quite right that there can be transient intervals where the bitrate exceeds the target.  We can also have tests running longer (but not much longer) than specified with `-t` due to network delays and possibly weird timing races (#278 is a quasi-example of this).  Couple questions:

1. In the perfSONAR use case, I understand that the main stopping condition is time-based (`-t`).  Is that the only one?  iperf3 supports a couple of other stopping conditions on the client, e.g. `-n` (total number of bytes) or `-k` (number of sends).

2.  Is the expectation that pSchedule will use the same flags to specify server limits as what the client wants?

mfeit-internet2:
0.  The bitrate and time limis don't have to be exact, just something where the server can reasonably  decide the client has overstayed its welcome and put a stop to the test.  The bandwidth end of it is more important than time for reasons I'll outline below.

1.  pScheduler runs tools based on time because scheduling is done on a timeline instead of queues like BWCTL did.  We build in some extra time at the end in case the tool we're running doesn't end exactly on time, but after that it will be killed off.  If we know that the time tolerance is going to be, say, 5%, we can adjust the iperf3 plugin's expectations to accommodate it.

1a.  The other reason I'd like the tool to exit on its own is for diagnostic reasons.  If it exits and explains why, that's more useful than our ham-fisted "we killed it because it ran longer than expected."  The latter is in place as a watchdog so we can be positive the test has been shut down before its scheduled time ends.

[...]

Issue #354, opened 2016-02-16
Enable '-t X' option in server mode


TheRealDJ:
I would like the server side to dictate the maximum duration for a test. Maybe also add a '--force-duration X' to override a client's requested test duration also.

Ex 1: Server called with '-t 30' and client called with '-t 30' runs 30s test
Ex 2: Server called with '-t 30' and client called with '-t 15' runs 15s test
Ex 3: Server called with '-t 15' and client called with '-t 30' runs 15s test
Ex 4: Server called with '--force-duration 30' and client called with '-t 15' runs 30s test

bmah888:
Solved with the merge of #1684

PR #999, opened 2020-05-18
Issue #937 - Add a server option to limit total allowed throughput 

davidBar-On:
* Version of iperf3 (or development branch, such as `master` or
  `3.1-STABLE`) to which this pull request applies:
Master

* Issues fixed (if any): 
#937 

* Brief description of code changes (suitable for use as a commit message):
Add --btotal server option to limit total server's throughput - send and received.
Limit is checked both during a new Client connection - based on the test data rate limit and number of streams, and during the test - based on the actual total average data rate (average is calculated over specified number of throughput report intervals).

PR #1684, opened 2024-04-19
Add Time Flag For The Iperf Server Process

manedurphy:

```
```

* Issues fixed (if any): #354, #615 

These changes add the functionality as requested by the linked issues. Specifically, this functionality allows the server to determine the maximum duration of an `iperf` measurement. The implementation is similar to the `--server-bitrate-limit` flag, where the measurement is forcibly stopped when an interval has exceeded the server's configured bitrate. In this case, the measurement is forcibly stopped when the duration exceeds the server's configured time, which is the value of `--server-time`.

### Example

From the dropdowns below, we can see that the server has enforced a measurement time of `3` seconds. The client attempts to run the measurement for `4` seconds. When the 3rd second is exceeded, the socket is forcibly closed.

<details>

<summary>server</summary>

```txt
./src/iperf3 -s --server-time 3
Server listening on 5201 (test #1)
Accepted connection from 127.0.0.1, port 44938
[  6] local 127.0.0.1 port 5201 connected to 127.0.0.1 port 44946
[ ID] Interval           Transfer     Bitrate
[  6]   0.00-1.00   sec  3.08 GBytes  26.5 Gbits/sec                  (omitted)
[  6]   1.00-2.00   sec  3.35 GBytes  28.8 Gbits/sec                  (omitted)
[  6]   2.00-3.00   sec  3.18 GBytes  27.3 Gbits/sec                  (omitted)
[  6]   0.00-1.00   sec  3.20 GBytes  27.5 Gbits/sec
[  6]   1.00-2.00   sec  3.34 GBytes  28.7 Gbits/sec
Server listening on 5201 (test #2)
```

</details>

<details>

<summary>client</summary>

```txt
./src/iperf3 --client 127.0.0.1 --port 5201 -t 4 --omit 3
Connecting to host 127.0.0.1, port 5201
[  5] local 127.0.0.1 port 44946 connected to 127.0.0.1 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec  3.08 GBytes  26.5 Gbits/sec    0   1.06 MBytes       (omitted)
[  5]   1.00-2.00   sec  3.35 GBytes  28.8 Gbits/sec    0   1.06 MBytes       (omitted)
[  5]   2.00-3.00   sec  3.15 GBytes  27.1 Gbits/sec    0   1.06 MBytes       (omitted)
[  5]   0.00-1.00   sec  3.20 GBytes  27.5 Gbits/sec    0   1.12 MBytes
[  5]   1.00-2.00   sec  3.34 GBytes  28.7 Gbits/sec    0   1.25 MBytes
[  5]   1.00-2.00   sec  3.34 GBytes  28.7 Gbits/sec    0   1.25 MBytes
```

</details>

<details>

<summary>client_json</summary>

```json
{
        "start":        {
                "connected":    [{
                                "socket":       5,
                                "local_host":   "127.0.0.1",
                                "local_port":   53314,
                                "remote_host":  "127.0.0.1",
                                "remote_port":  5201
                        }],
                "version":      "iperf 3.16+",
                "system_info":  "Linux windows-nexus 5.15.146.1-microsoft-standard-WSL2 #1 SMP Thu Jan 11 04:09:03 UTC 2024 x86_64",
                "timestamp":    {
                        "time": "Thu, 18 Apr 2024 04:38:09 GMT",
                        "timesecs":     1713415089
                },
                "connecting_to":        {
                        "host": "127.0.0.1",
                        "port": 5201
                },
                "cookie":       "wtuhwwetnqayseccmqno6dutzjf3zly3op5p",
                "tcp_mss_default":      32768,
                "target_bitrate":       0,
                "fq_rate":      0,
                "sock_bufsize": 0,
                "sndbuf_actual":        16384,
                "rcvbuf_actual":        131072,
                "test_start":   {
                        "protocol":     "TCP",
                        "num_streams":  1,
                        "blksize":      131072,
                        "omit": 3,
                        "duration":     4,
                        "bytes":        0,
                        "blocks":       0,
                        "reverse":      0,
                        "tos":  0,
                        "target_bitrate":       0,
                        "bidir":        0,
                        "fqrate":       0,
                        "interval":     1
                }
        },
        "intervals":    [{
                        "streams":      [{
                                        "socket":       5,
                                        "start":        0,
                                        "end":  1.001066,
                                        "seconds":      1.0010659694671631,
                                        "bytes":        3578265600,
                                        "bits_per_second":      28595642717.968742,
                                        "retransmits":  0,
                                        "snd_cwnd":     2029973,
                                        "snd_wnd":      3145088,
                                        "rtt":  51,
                                        "rttvar":       5,
                                        "pmtu": 65535,
                                        "omitted":      true,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        0,
                                "end":  1.001066,
                                "seconds":      1.0010659694671631,
                                "bytes":        3578265600,
                                "bits_per_second":      28595642717.968742,
                                "retransmits":  0,
                                "omitted":      true,
                                "sender":       true
                        }
                }, {
                        "streams":      [{
                                        "socket":       5,
                                        "start":        1.001066,
                                        "end":  2.001046,
                                        "seconds":      0.99997997283935547,
                                        "bytes":        3452698624,
                                        "bits_per_second":      27622142185.078888,
                                        "retransmits":  0,
                                        "snd_cwnd":     2029973,
                                        "snd_wnd":      3145088,
                                        "rtt":  88,
                                        "rttvar":       67,
                                        "pmtu": 65535,
                                        "omitted":      true,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        1.001066,
                                "end":  2.001046,
                                "seconds":      0.99997997283935547,
                                "bytes":        3452698624,
                                "bits_per_second":      27622142185.078888,
                                "retransmits":  0,
                                "omitted":      true,
                                "sender":       true
                        }
                }, {
                        "streams":      [{
                                        "socket":       5,
                                        "start":        2.001046,
                                        "end":  3.001049,
                                        "seconds":      1.0000029802322388,
                                        "bytes":        3500015616,
                                        "bits_per_second":      28000041481.373692,
                                        "retransmits":  0,
                                        "snd_cwnd":     2029973,
                                        "snd_wnd":      3145088,
                                        "rtt":  63,
                                        "rttvar":       17,
                                        "pmtu": 65535,
                                        "omitted":      true,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        2.001046,
                                "end":  3.001049,
                                "seconds":      1.0000029802322388,
                                "bytes":        3500015616,
                                "bits_per_second":      28000041481.373692,
                                "retransmits":  0,
                                "omitted":      true,
                                "sender":       true
                        }
                }, {
                        "streams":      [{
                                        "socket":       5,
                                        "start":        0.012641,
                                        "end":  0.988366,
                                        "seconds":      1.0010069608688354,
                                        "bytes":        3449421824,
                                        "bits_per_second":      27567615082.364941,
                                        "retransmits":  0,
                                        "snd_cwnd":     2029973,
                                        "snd_wnd":      3145088,
                                        "rtt":  58,
                                        "rttvar":       7,
                                        "pmtu": 65535,
                                        "omitted":      false,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        0.012641,
                                "end":  0.988366,
                                "seconds":      1.0010069608688354,
                                "bytes":        3449421824,
                                "bits_per_second":      27567615082.364941,
                                "retransmits":  0,
                                "omitted":      false,
                                "sender":       true
                        }
                }, {
                        "streams":      [{
                                        "socket":       5,
                                        "start":        0.988366,
                                        "end":  1.988398,
                                        "seconds":      1.0000319480895996,
                                        "bytes":        3414818816,
                                        "bits_per_second":      27317677780.382618,
                                        "retransmits":  0,
                                        "snd_cwnd":     2029973,
                                        "snd_wnd":      3145088,
                                        "rtt":  61,
                                        "rttvar":       9,
                                        "pmtu": 65535,
                                        "omitted":      false,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        0.988366,
                                "end":  1.988398,
                                "seconds":      1.0000319480895996,
                                "bytes":        3414818816,
                                "bits_per_second":      27317677780.382618,
                                "retransmits":  0,
                                "omitted":      false,
                                "sender":       true
                        }
                }, {
                        "streams":      [{
                                        "socket":       5,
                                        "start":        0.988366,
                                        "end":  1.988398,
                                        "seconds":      1.0000319480895996,
                                        "bytes":        3414818816,
                                        "bits_per_second":      27317677780.382618,
                                        "retransmits":  0,
                                        "snd_cwnd":     2029973,
                                        "snd_wnd":      3145088,
                                        "rtt":  61,
                                        "rttvar":       9,
                                        "pmtu": 65535,
                                        "omitted":      false,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        0.988366,
                                "end":  1.988398,
                                "seconds":      1.0000319480895996,
                                "bytes":        3414818816,
                                "bits_per_second":      27317677780.382618,
                                "retransmits":  0,
                                "omitted":      false,
                                "sender":       true
                        }
                }],
        "end":  {
        },
        "error":        "control socket has closed unexpectedly"
}
```

</details>

### Implementation Details

With this implementation, the server overrides the value of `test->duration` with the value that was passed in via the `--server-time` flag. We can see from the drop downs that the client's experience is similar to when the bitrate limit is exceeded with the error: `iperf3: error - control socket has closed unexpectedly`. However, the server side is less friendly, with an error of `iperf3: error - select failed: Bad file descriptor`. I can already envision myself encountering this error and not knowing why I am getting it, as well as forgetting that I had set the `--server-time` flag to begin with.

Another thing to note is that nothing is stopping the user from setting a high value for `--omit`. Given that this feature is meant to protect the server, is it reasonable to include a `--server-omit` option to enforce a maximum value?

### Update (04/18/2024)

Since the first commit, I've introduced a solution for the error `iperf3: error - select failed: Bad file descriptor` which provides a more friendly experience. Let's consider the scenario where the `--server-time` flag has been set on the server to a value of `3`:

1. The client connects to the server
2. The server accepts the connection, receives the client's cookie, and sends the client that value of `3` for the server time
3. When the client's timers are created, it compares its duration to the server's duration of `3`. If the client's duration is greater than the server's, then the server's value is selected for the timer.
4. The same process occurs when the server's timers are created.

The JSON output also has a new field which specifies the server's value for duration: `server_duration`

<details>

<summary>client_json</summary>

```json
{
        "start":        {
                "connected":    [{
                                "socket":       5,
                                "local_host":   "127.0.0.1",
                                "local_port":   36906,
                                "remote_host":  "127.0.0.1",
                                "remote_port":  5201
                        }],
                "version":      "iperf 3.16+",
                "system_info":  "Linux windows-nexus 5.15.146.1-microsoft-standard-WSL2 #1 SMP Thu Jan 11 04:09:03 UTC 2024 x86_64",
                "timestamp":    {
                        "time": "Mon, 22 Apr 2024 02:23:53 GMT",
                        "timesecs":     1713752633
                },
                "connecting_to":        {
                        "host": "127.0.0.1",
                        "port": 5201
                },
                "cookie":       "m4n5p26auofjvyda2x2ba56kexzsly7mz5k7",
                "tcp_mss_default":      32768,
                "target_bitrate":       5000000,
                "fq_rate":      0,
                "sock_bufsize": 0,
                "sndbuf_actual":        16384,
                "rcvbuf_actual":        131072,
                "test_start":   {
                        "protocol":     "TCP",
                        "num_streams":  1,
                        "blksize":      131072,
                        "omit": 0,
                        "duration":     6,
                        "server_duration":      3,
                        "bytes":        0,
                        "blocks":       0,
                        "reverse":      0,
                        "tos":  0,
                        "target_bitrate":       5000000,
                        "bidir":        0,
                        "fqrate":       0,
                        "interval":     1
                }
        },
        "intervals":    [{
                        "streams":      [{
                                        "socket":       5,
                                        "start":        0,
                                        "end":  1.000087,
                                        "seconds":      1.0000870227813721,
                                        "bytes":        655360,
                                        "bits_per_second":      5242423.78970069,
                                        "retransmits":  0,
                                        "snd_cwnd":     654830,
                                        "snd_wnd":      1375232,
                                        "rtt":  38,
                                        "rttvar":       24,
                                        "pmtu": 65535,
                                        "omitted":      false,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        0,
                                "end":  1.000087,
                                "seconds":      1.0000870227813721,
                                "bytes":        655360,
                                "bits_per_second":      5242423.78970069,
                                "retransmits":  0,
                                "omitted":      false,
                                "sender":       true
                        }
                }, {
                        "streams":      [{
                                        "socket":       5,
                                        "start":        1.000087,
                                        "end":  2.000084,
                                        "seconds":      0.99999701976776123,
                                        "bytes":        655360,
                                        "bits_per_second":      5242895.6250465661,
                                        "retransmits":  0,
                                        "snd_cwnd":     654830,
                                        "snd_wnd":      2684928,
                                        "rtt":  45,
                                        "rttvar":       22,
                                        "pmtu": 65535,
                                        "omitted":      false,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        1.000087,
                                "end":  2.000084,
                                "seconds":      0.99999701976776123,
                                "bytes":        655360,
                                "bits_per_second":      5242895.6250465661,
                                "retransmits":  0,
                                "omitted":      false,
                                "sender":       true
                        }
                }, {
                        "streams":      [{
                                        "socket":       5,
                                        "start":        2.000084,
                                        "end":  3.000148,
                                        "seconds":      1.0000640153884888,
                                        "bytes":        655360,
                                        "bits_per_second":      5242544.39648379,
                                        "retransmits":  0,
                                        "snd_cwnd":     654830,
                                        "snd_wnd":      3112448,
                                        "rtt":  62,
                                        "rttvar":       44,
                                        "pmtu": 65535,
                                        "omitted":      false,
                                        "sender":       true
                                }],
                        "sum":  {
                                "start":        2.000084,
                                "end":  3.000148,
                                "seconds":      1.0000640153884888,
                                "bytes":        655360,
                                "bits_per_second":      5242544.39648379,
                                "retransmits":  0,
                                "omitted":      false,
                                "sender":       true
                        }
                }],
        "end":  {
                "streams":      [{
                                "sender":       {
                                        "socket":       5,
                                        "start":        0,
                                        "end":  3.000148,
                                        "seconds":      3.000148,
                                        "bytes":        1966080,
                                        "bits_per_second":      5242621.36401271,
                                        "retransmits":  0,
                                        "max_snd_cwnd": 654830,
                                        "max_snd_wnd":  3112448,
                                        "max_rtt":      62,
                                        "min_rtt":      38,
                                        "mean_rtt":     48,
                                        "sender":       true
                                },
                                "receiver":     {
                                        "socket":       5,
                                        "start":        0,
                                        "end":  3.000258,
                                        "seconds":      3.000148,
                                        "bytes":        1966080,
                                        "bits_per_second":      5242429.1510930061,
                                        "sender":       true
                                }
                        }],
                "sum_sent":     {
                        "start":        0,
                        "end":  3.000148,
                        "seconds":      3.000148,
                        "bytes":        1966080,
                        "bits_per_second":      5242621.36401271,
                        "retransmits":  0,
                        "sender":       true
                },
                "sum_received": {
                        "start":        0,
                        "end":  3.000258,
                        "seconds":      3.000258,
                        "bytes":        1966080,
                        "bits_per_second":      5242429.1510930061,
                        "sender":       true
                },
                "cpu_utilization_percent":      {
                        "host_total":   100.87966904530634,
                        "host_user":    100.8797023673657,
                        "host_system":  0,
                        "remote_total": 0.037496475331318856,
                        "remote_user":  0.037496475331318856,
                        "remote_system":        0
                },
                "sender_tcp_congestion":        "cubic",
                "receiver_tcp_congestion":      "cubic"
        }
}
```

</details>

### Update (06/05/2024)

A valid question was brought up by @TheRealDJ: `What happens when the server has this feature but the client doesn't?`. With the current implementation, there is no backwards compatability. Further discussions will need to be had in order to handle this case properly.

### Update (05/17/2025)

A new implementation has been made to address the backwards compatibility issue mentioned above.

#### Implementation Details

The server has a new flag, `--server-max-duration`, represents the maximum permitted duration for an `iperf3` test in seconds. When the client initiates a test, it sends its desired duration to the server. The server then compares the client's requested duration with its own maximum duration. If the client's requested duration exceeds the server's maximum, the server rejects the test and sends an error message back to the client. Since older client's do not have the newly added `IEMAXSERVERDURATIONEXCEEDED` error, the handler for `int_errno` falls into the default case. The output for the server and both outputs for the client are shown below.

<details>

<summary>server</summary>

```txt
Server listening on 5201 (test #1)
iperf3: error - client's max duration exceeds the server's maximum permitted duration
```

</details>

<details>

<summary>client (new)</summary>

```txt
Connecting to host 127.0.0.1, port 5201
iperf3: error - client's max duration exceeds the server's maximum permitted duration
```

</details>

<details>

<summary>client (old)</summary>

```txt
Connecting to host 127.0.0.1, port 5201
iperf3: error - int_errno=37: No locks available
```

</details>

#### Questions

Is this output acceptable? The error message is not very user friendly, but I'm not sure of a way to relay more information when the error message is based on the `int_errno` value. Ideally, the server and the client should exchange information on their respective versions, but since that functionality does not currently exist, I am uncertain of the best way to handle this.

### Update (05/19/2025)

@TheRealDJ mentioned that the sum of the client's duration and omits must not exceed the server's maximum duration. This change has been implemented.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1114,7 +1114,8 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
         {"udp", no_argument, NULL, 'u'},
         {"bitrate", required_argument, NULL, 'b'},
         {"bandwidth", required_argument, NULL, 'b'},
-	{"server-bitrate-limit", required_argument, NULL, OPT_SERVER_BITRATE_LIMIT},
+        {"server-bitrate-limit", required_argument, NULL, OPT_SERVER_BITRATE_LIMIT},
+        {"server-max-duration", required_argument, NULL, OPT_SERVER_MAX_DURATION},
         {"time", required_argument, NULL, 't'},
         {"bytes", required_argument, NULL, 'n'},
         {"blockcount", required_argument, NULL, 'k'},
@@ -1555,6 +1556,13 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                 }
 		server_flag = 1;
 	        break;
+            case OPT_SERVER_MAX_DURATION:
+                test->max_server_duration = atoi(optarg);
+                if (test->max_server_duration < 0 || test->max_server_duration > MAX_TIME) {
+                    i_errno = IEDURATION;
+                    return -1;
+                }
+                break;
             case OPT_RCV_TIMEOUT:
                 rcv_timeout_in = atoi(optarg);
                 if (rcv_timeout_in < MIN_NO_MSG_RCVD_TIMEOUT || rcv_timeout_in > MAX_TIME * SEC_TO_mS) {
@@ -2286,6 +2294,20 @@ iperf_exchange_parameters(struct iperf_test *test)
                 i_errno = IECTRLWRITE;
                 return -1;
             }
+            return -1;
+        }
+
+        if (test->max_server_duration > 0 && test->duration > test->max_server_duration) {
+            if (iperf_set_send_state(test, SERVER_ERROR) != 0)
+                return -1;
+            
+            i_errno = IEMAXSERVERDURATIONEXCEEDED;
+            err = htonl(i_errno);
+            if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
+                i_errno = IECTRLWRITE;
+                return -1;
+            }
+
             return -1;
         }
 
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2302,7 +2302,7 @@ iperf_exchange_parameters(struct iperf_test *test)
             if (iperf_set_send_state(test, SERVER_ERROR) != 0)
                 return -1;
 
-            i_errno = IEMAXSERVERDURATIONEXCEEDED;
+            i_errno = IEMAXSERVERTESTDURATIONEXCEEDED;
             err = htonl(i_errno);
             if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
                 i_errno = IECTRLWRITE;
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2308,6 +2308,12 @@ iperf_exchange_parameters(struct iperf_test *test)
                 i_errno = IECTRLWRITE;
                 return -1;
             }
+            
+            err = htonl(errno);
+            if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
+                i_errno = IECTRLWRITE;
+                return -1;
+            }
 
             return -1;
         }
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2297,7 +2297,7 @@ iperf_exchange_parameters(struct iperf_test *test)
             return -1;
         }
 
-        if (test->max_server_duration > 0 && ((test->duration + test->omit) > test->max_server_duration) || test->duration == 0) {
+        if ((test->max_server_duration > 0) && (((test->duration + test->omit) > test->max_server_duration) || (test->duration == 0))) {
             if (iperf_set_send_state(test, SERVER_ERROR) != 0)
                 return -1;
 
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2297,7 +2297,7 @@ iperf_exchange_parameters(struct iperf_test *test)
             return -1;
         }
 
-        if (test->max_server_duration > 0 && (test->duration + test->omit) > test->max_server_duration) {
+        if (test->max_server_duration > 0 && ((test->duration + test->omit) > test->max_server_duration) || test->duration == 0) {
             if (iperf_set_send_state(test, SERVER_ERROR) != 0)
                 return -1;
             
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2570,13 +2570,6 @@ get_parameters(struct iperf_test *test)
 	    test->repeating_payload = 1;
 	if ((j_p = iperf_cJSON_GetObjectItemType(j, "zerocopy", cJSON_Number)) != NULL)
 	    test->zerocopy = j_p->valueint;
-
-    /* Ensure that the client does not request to run longer than the server's configured max */
-    if ((test->max_server_duration > 0) && (((test->duration + test->omit) > test->max_server_duration) || (test->duration == 0))) {
-        i_errno = IEMAXSERVERTESTDURATIONEXCEEDED;
-        r = -1;
-    }
-
 #if defined(HAVE_DONT_FRAGMENT)
 	if ((j_p = iperf_cJSON_GetObjectItemType(j, "dont_fragment", cJSON_Number)) != NULL)
 	    test->settings->dont_fragment = j_p->valueint;
@@ -2592,6 +2585,13 @@ get_parameters(struct iperf_test *test)
 	if (test->settings->rate)
 	    cJSON_AddNumberToObject(test->json_start, "target_bitrate", test->settings->rate);
 	cJSON_Delete(j);
+
+    /* Ensure that the client does not request to run longer than the server's configured max */
+    if ((test->max_server_duration > 0) && (((test->duration + test->omit) > test->max_server_duration) || (test->duration == 0))) {
+        i_errno = IEMAXSERVERTESTDURATIONEXCEEDED;
+        r = -1;
+    }
+
     }
     return r;
 }
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2297,7 +2297,7 @@ iperf_exchange_parameters(struct iperf_test *test)
             return -1;
         }
 
-        if (test->max_server_duration > 0 && test->duration > test->max_server_duration) {
+        if (test->max_server_duration > 0 && (test->duration + test->omit) > test->max_server_duration) {
             if (iperf_set_send_state(test, SERVER_ERROR) != 0)
                 return -1;
             
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2300,7 +2300,7 @@ iperf_exchange_parameters(struct iperf_test *test)
         if (test->max_server_duration > 0 && ((test->duration + test->omit) > test->max_server_duration) || test->duration == 0) {
             if (iperf_set_send_state(test, SERVER_ERROR) != 0)
                 return -1;
-            
+
             i_errno = IEMAXSERVERDURATIONEXCEEDED;
             err = htonl(i_errno);
             if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2298,26 +2298,6 @@ iperf_exchange_parameters(struct iperf_test *test)
             return -1;
         }
 
-        if ((test->max_server_duration > 0) && (((test->duration + test->omit) > test->max_server_duration) || (test->duration == 0))) {
-            if (iperf_set_send_state(test, SERVER_ERROR) != 0)
-                return -1;
-
-            i_errno = IEMAXSERVERTESTDURATIONEXCEEDED;
-            err = htonl(i_errno);
-            if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
-                i_errno = IECTRLWRITE;
-                return -1;
-            }
-            
-            err = htonl(errno);
-            if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
-                i_errno = IECTRLWRITE;
-                return -1;
-            }
-
-            return -1;
-        }
-
 #if defined(HAVE_SSL)
         if (test_is_authorized(test) < 0){
             if (iperf_set_send_state(test, SERVER_ERROR) != 0)
@@ -2590,6 +2570,13 @@ get_parameters(struct iperf_test *test)
 	    test->repeating_payload = 1;
 	if ((j_p = iperf_cJSON_GetObjectItemType(j, "zerocopy", cJSON_Number)) != NULL)
 	    test->zerocopy = j_p->valueint;
+
+    /* Ensure that the client does not request to run longer than the server's configured max */
+    if ((test->max_server_duration > 0) && (((test->duration + test->omit) > test->max_server_duration) || (test->duration == 0))) {
+        i_errno = IEMAXSERVERTESTDURATIONEXCEEDED;
+        r = -1;
+    }
+
 #if defined(HAVE_DONT_FRAGMENT)
 	if ((j_p = iperf_cJSON_GetObjectItemType(j, "dont_fragment", cJSON_Number)) != NULL)
 	    test->settings->dont_fragment = j_p->valueint;
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1562,6 +1562,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                     i_errno = IEDURATION;
                     return -1;
                 }
+                server_flag = 1;
                 break;
             case OPT_RCV_TIMEOUT:
                 rcv_timeout_in = atoi(optarg);
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -334,7 +334,7 @@ iperf_handle_message_client(struct iperf_test *test)
     switch (test->state) {
         case PARAM_EXCHANGE:
             if (iperf_exchange_parameters(test) < 0)
-            return -1;
+                return -1;
             if (test->on_connect)
                 test->on_connect(test);
             break;
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -334,7 +334,7 @@ iperf_handle_message_client(struct iperf_test *test)
     switch (test->state) {
         case PARAM_EXCHANGE:
             if (iperf_exchange_parameters(test) < 0)
-                return -1;
+            return -1;
             if (test->on_connect)
                 test->on_connect(test);
             break;
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -551,6 +551,9 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "unable to set/get socket keepalive TCP number of retries (TCP_KEEPCNT) option");
             perr = 1;
             break;
+        case IEMAXSERVERDURATIONEXCEEDED:
+            snprintf(errstr, len, "client's max duration exceeds the server's maximum permitted duration");
+            break;
 	default:
 	    snprintf(errstr, len, "int_errno=%d", int_errno);
 	    perr = 1;
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -551,7 +551,7 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "unable to set/get socket keepalive TCP number of retries (TCP_KEEPCNT) option");
             perr = 1;
             break;
-        case IEMAXSERVERDURATIONEXCEEDED:
+        case IEMAXSERVERTESTDURATIONEXCEEDED:
             snprintf(errstr, len, "client's max duration exceeds the server's maximum permitted duration");
             break;
 	default:
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -552,7 +552,7 @@ iperf_strerror(int int_errno)
             perr = 1;
             break;
         case IEMAXSERVERTESTDURATIONEXCEEDED:
-            snprintf(errstr, len, "client's max duration exceeds the server's maximum permitted duration");
+            snprintf(errstr, len, "client's requested duration exceeds the server's maximum permitted limit");
             break;
 	default:
 	    snprintf(errstr, len, "int_errno=%d", int_errno);
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -312,6 +312,7 @@ struct iperf_test
     int       server_port;
     int       omit;                             /* duration of omit period (-O flag) */
     int       duration;                         /* total duration of test (-t flag) */
+    int       max_server_duration;               /* maximum possible duration of test as enforced by the server (--max-server-duration flag) */
     char     *diskfile_name;			/* -F option */
     int       affinity, server_affinity;	/* -A option */
 #if defined(HAVE_CPUSET_SETAFFINITY)
--- a/src/iperf3.1
+++ b/src/iperf3.1
@@ -251,7 +251,7 @@ In one-off mode, this is the number of seconds the server will wait
 before exiting.
 .TP
 .BR --server-max-duration " "
-max time, in seconds, that an iperf client can run against the server.
+The maximum time, in seconds, that an iperf client can run against the server.
 When the sum of the client's time and omit values exceeds the max duration set by the server
 or the client's time value is 0, the measurement is rejected.
 .TP
--- a/src/iperf3.1
+++ b/src/iperf3.1
@@ -250,6 +250,11 @@ Restart the server after \fIn\fR seconds in case it gets stuck.
 In one-off mode, this is the number of seconds the server will wait
 before exiting.
 .TP
+.BR --server-max-duration " "
+max time, in seconds, that an iperf client can run against the server.
+When the sum of the client's time and omit values exceeds the max duration set by the server
+or the client's time value is 0, the measurement is rejected.
+.TP
 .BR --server-bitrate-limit " \fIn\fR[KMGT][/\fCn\fR]"
 Set a limit on the server side, which will cause a test to abort if
 the client specifies a test of more than \fIn\fR bits per second, or
--- a/src/iperf_api.h
+++ b/src/iperf_api.h
@@ -105,6 +105,7 @@ typedef atomic_uint_fast64_t atomic_iperf_size_t;
 #define OPT_CNTL_KA 31
 #define OPT_SKIP_RX_COPY 32
 #define OPT_JSON_STREAM_FULL_OUTPUT 33
+#define OPT_SERVER_MAX_DURATION 33
 
 /* states */
 #define TEST_START 1
@@ -440,6 +441,7 @@ enum {
     IEUDPFILETRANSFER = 34, // Cannot transfer file using UDP
     IESERVERAUTHUSERS = 35,  // Cannot access authorized users file
     IECNTLKA = 36,          // Control connection Keepalive period should be larger than the full retry period (interval * count)
+    IEMAXSERVERDURATIONEXCEEDED = 37, // Client's duration exceeds server's maximum duration
     /* Test errors */
     IENEWTEST = 100,        // Unable to create a new test (check perror)
     IEINITTEST = 101,       // Test initialization failed (check perror)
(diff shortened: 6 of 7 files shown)


manedurphy:
Update (05/22/2025)

@bmah888 expressed concern over the value for the added `OPT_SERVER_MAX_DURATION`, as it is approaching a value close to those of other short-hand flags. The way I see it, these should be okay to increment as is up until ASCII `47`, as these values are unlikely to be used for flags. We can also increase it to `128` now and never think about it again. I'll leave that decision up to Bruce. My only concern left for the current implementation is the user-unfriendly message when the `iperf3` client's version is older. Is this a known limitation that users are familiar enough with to know what to do when then encounter the message `iperf3: error - int_errno=37: No locks available` which provides the error code from the server that the client does not have a case for? And if this is a well-known limitation and we can merge this change as is, is there room for an enhancement in the future? The way that I would approach this is for the client and server to exchange their respective versions, and when the server encounters an error for a feature that the client doesn't have awareness of, it will send a special code which indicates that the error has to do with a feature on a newer version of `iperf3`..... or something like that.

src/iperf_api.c
```diff
@@ -1534,6 +1535,13 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                 }
 		server_flag = 1;
 	        break;
+            case OPT_SERVER_MAX_DURATION:
```
davidBar-On, line 1538: `server_flag = 1;` should be added to this case.

manedurphy:
@bmah888 have you had some time to consider the approach we should take here given the constraints? @davidBar-On your input on this would also be appreciated.

It appears that it has always been an issue that an `iperf3` client could run into a case where it cannot provide more context for what happened on the server. Here is the output for a `3.8.1` client that is trying to exceed the duration of the `--server-max-duration` flag:

```
iperf3 -c 127.0.0.1 -t 10
Connecting to host 127.0.0.1, port 5201
iperf3: error -
```

So while a more sophisticated approach for providing more context to older versions of the client would be ideal, I'm not sure that it is absolutely necessary given the historical context. At least not necessary to move forward with this feature. An enhancement for this can be added to the code base separately.

src/iperf_client_api.c
```diff
@@ -334,7 +334,7 @@ iperf_handle_message_client(struct iperf_test *test)
     switch (test->state) {
         case PARAM_EXCHANGE:
             if (iperf_exchange_parameters(test) < 0)
-                return -1;
+            return -1;
```
davidBar-On, line 337: Was this removal of leading spaces done by mistake?  Note that iperf3 indentation is 8 spaces.

src/iperf_error.c
```diff
@@ -548,6 +548,9 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "unable to set/get socket keepalive TCP number of retries (TCP_KEEPCNT) option");
             perr = 1;
             break;
+        case IEMAXSERVERDURATIONEXCEEDED:
+            snprintf(errstr, len, "client's max duration exceeds the server's maximum permitted duration");
```
davidBar-On, line 552: Suggest to replace "duration" with "test duration".

src/iperf_api.c
```diff
@@ -1095,7 +1095,8 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
         {"udp", no_argument, NULL, 'u'},
         {"bitrate", required_argument, NULL, 'b'},
         {"bandwidth", required_argument, NULL, 'b'},
-	{"server-bitrate-limit", required_argument, NULL, OPT_SERVER_BITRATE_LIMIT},
+        {"server-bitrate-limit", required_argument, NULL, OPT_SERVER_BITRATE_LIMIT},
+        {"server-max-duration", required_argument, NULL, OPT_SERVER_MAX_DURATION},
```
davidBar-On, line 1099: There is no need to add a new option, as `--time` can be used (the iperf3 team don't like new options).  Client will store the value in `duration` and server will store it in `max_server_duration`.  Note that when getting the `--time` value it should be stored in temporary variable, and only after all option values are received and it is known whether this is a server or client, it can be stored into the test structure.

This requires that the `--time` will become server and client feature, and the help message should state both the client and the server messages.  Note that

davidBar-On (changes requested):

src/iperf_error.c
manedurphy: Changed

src/iperf_api.c
manedurphy: The approach I've taken was to be consistent with [iperf/#999](https://github.com/esnet/iperf/pull/999), where the client's `--bitrate` flag and the server's `--server-bitrate-limit` flag are handled separately under different options even though they are both related to configuring bit rate. I'm in favor of maintaining consistency for the sake of the user's experience, but I also think that `--server-max-duration` makes it more clear that this flag has to do with test duration enforcement. We can even change it to `--server-max-test-duration` and `OPT_SERVER_MAX_TEST_DURATION` to be more consistent with the previous recommendation for changing `IEMAXSERVERDURATIONEXCEEDED` to `IEMAXSERVERTESTDURATIONEXCEEDED`. What do you think?

bmah888: I think adding a new flag is OK this case.

bmah888:
> @bmah888 have you had some time to consider the approach we should take here given the constraints? @davidBar-On your input on this would also be appreciated.
> 
> It appears that it has always been an issue that an `iperf3` client could run into a case where it cannot provide more context for what happened on the server. Here is the output for a `3.8.1` client that is trying to exceed the duration of the `--server-max-duration` flag:
> 
> ```
> iperf3 -c 127.0.0.1 -t 10
> Connecting to host 127.0.0.1, port 5201
> iperf3: error -
> ```
> 
> So while a more sophisticated approach for providing more context to older versions of the client would be ideal, I'm not sure that it is absolutely necessary given the historical context. At least not necessary to move forward with this feature. An enhancement for this can be added to the code base separately.

I agree with you here...iperf3 wasn't designed to support perfect forward/backward compatibility between arbitrary versions of code. It was intended for deployments within a single administrative domain (or at least between cooperating administrative domains), where the requirements for compatibility are lessened a bit. In this case all we're missing is the ability to display a meaningful error message to the end-user on the client side.

bmah888:
I've been testing this and it seems to be working as designed. 

One thing just occurred to me that if you run the server as `iperf3 --server --server-max-duration`, I can still force a really long test to run by doing something like `iperf3 --client iperf3.example.com --bytes 1T` or something like that (in other words, have my client specify a number of bytes or a number of blocks to transmit, both of which override the time duration). I'm not sure how to handle these cases, short of adding more limiting options.

manedurphy:
@bmah888 I'm open to adding those protections. I think having a flag per protection is fine since the goal is to give the administrator of the `iperf3` server more control over preventing a single client's ability to flood traffic. If I have a 10 Gbps server that is supposed to allow some pool of 30 devices to test their 1 Gbps speeds, then I can better ensure that no device is able to hog the server's bandwidth by:

1. Setting `--server-max-bitrate` to `1Gb/1s`.
2. Setting `--max-server-duration` to `15` since that's plenty of time to verify speed. The client can set `--omit 5` to account for TCP ramp up and `--time 10` for the test duration.
3. This would mean the theoretical max bytes transferred would be `1875000000 bytes`, to which I could set `--server-max-bytes` to protect the server from any device that wants to run a longer test. This would mean that devices with slower bit rates could run for longer than the configured `--server-max-duration`, but would never be able to exceed the maximum configured number of bytes that the server will accept via the `--server-max-bytes` flag.

For me, personally, as an `iperf3` server administrator, I care the most for my server's availability to meaningfully serve as many clients as possible, and preventing clients from dominating the server's resources. If someone wants to run a long `iperf3` client with `--bitrate 1Kb/1s --bytes 1875000000`, that's fine with me because it won't interfere with other devices.

I'd want to know if these additional changes should be implemented as part of these changes, or if I should raise them in a separate PR. I'm open to either, but tend to prefer raising separate PRs for separation of concern for the PR template.

davidBar-On:
> ... as an iperf3 server administrator, I care the most for my server's availability to meaningfully serve as many clients as possible, and preventing clients from dominating the server's resources.

I assume that means that the duration that a specific client is using the server is the most important limit.  In this case, instead of limiting bytes/blocks, at the beginning of the test the server can start a timer that will expire after the `--server-max-duration` and will allow the server to terminate the test.

Actually, the server already does something similar, in `create_server_timers()`.  Based on the test duration `test->duration + test->omit + grace_period`.  If that timer expires, `server_timer_proc()` is called and it terminates the test (I have submitted PR #1914 to enhance this function error handling).  The suggested approach is that when `--server-max-duration` is set, the timer will be set to the minimum between the parameter value and the value currently used for the timer.  Something like this:
```c
        max_duration = test->duration + test->omit + grace_period;
        if (test->max_server_duration > 0 && test->max_server_duration < max_duration) {
            max_duration = test->max_server_duration;
        }
        test->timer = tmr_create(&now, server_timer_proc, cd, max_duration * SEC_TO_US, 0);
```

bmah888:
> For me, personally, as an `iperf3` server administrator, I care the most for my server's availability to meaningfully serve as many clients as possible, and preventing clients from dominating the server's resources. If someone wants to run a long `iperf3` client with `--bitrate 1Kb/1s --bytes 1875000000`, that's fine with me because it won't interfere with other devices.

Thanks @manedurphy! Herein we're running into a difference in use cases between what iperf3 was originally designed for, and what people would (quite reasonably) like it to do. :-) iperf3 was originally designed to test between two specific hosts (i.e. client and server) at a time. In the networking environment it was designed for, it's very possible that the end hosts, rather than the network, are the bottlenecks to performance. So only one test per server process can be running at once (and ideally only one test per host at a time), to avoid interference between tests. There was also a general assumption that the client and server were loosely affiliated, or at least co-operating.

This is very different from people who want to run large-scale iperf3 speed-test type services that serve dozens or hundreds of potentially unknown clients, some of which might need some guardrails. I understand that the PR we're discussing is an attempt to adapt iperf3 to this use case, which is great. I'm just trying to poke (in a friendly way, or at least I'm trying!) at the solution to see what else might be needed to fit the requirements.

I appreciate wanting to support (server-side) limits on parameters as well, although I am not sure how these options should interact, particularly when the client side can only specify one of the ending criteria. So...my question back to you is do you think this PR is still useful as-is? I haven't quite made up my mind on this. (I see where @davidBar-On has some feedback also, which I haven't read yet.)

manedurphy:
@bmah888 I appreciate the additional context for `iperf3's` original design and use case. You are correct, this feature stems from a use case of large-scale speed testing. We have pools of `10Gb/s`, `40Gb/s`, and `100Gb/s` servers that are assigned to physical locations for serving `iperf3` clients that reside in those areas. We have 1 client process that interacts with 1 server process as `iperf3` was designed to do, but we cannot have only 1 server process per host due to our scale. Depending on the location, it can be either tens or hundreds of `iperf3` server processes on a single host, each of which interact with a single client process. I believe that this server-protecting feature falls in line with the `--server-bitrate-max` feature in [iperf/#999](https://github.com/esnet/iperf/pull/999), which I think has more usefulness when the client and server are not affiliated as does the proposed `--server-max-duration`.

@davidBar-On your proposed change of using the value of `--server-max-duration` to set the value of the timer actually falls more in line with the original implementation that @TheRealDJ and I discussed. However, I believe that the server should not allow the client to run at all when it can determine that it is destined to fail based on the client's input parameters. The same can be (conditionally) said for the `--server-bitrate-max` flag in the sense that if the server has this flag set to `1Gbit/1` and the client has `--bitrate` set to `2Gbit/1`, then the server should reject the test because it knows that it will exceed its configured maximum, and it should provide a message to the client which makes it clear why the test was rejected. In the cases where the server cannot determine this in advance, such as when the `--bitrate` in the client is not set, then the server should ideally send a user-friendly message to the client. As of now, there is a nice message on the server, `iperf3: error - total required bandwidth is larger than server limit`, but the client has no visibility into what went wrong, `iperf3: error - control socket has closed unexpectedly`. I believe the enhancement in [iperf/#1914](https://github.com/esnet/iperf/pull/1914) intends to do the same, where the server would print `iperf3: error - server test duration expired - test is terminated by the server`, but the client would still have no visibility into what went wrong. I'm bringing up the `--server-bitrate-max` here because I believe it's usefulness in protecting the server aligns with the purpose of the proposed `--server-duration-max`, and I believe that all server-protecting flags, current and future, should follow the same set of guidelines for implementation.

Guidelines:
- When the server has a protection flag set, like `--server-bitrate-limit`, `--server-max-duration`, or `--server-max-bytes`, etc, it should be prepared to evaluate and react to the client's respective input parameters: `--bitrate`, `--time`, `bytes`, etc.
- Should the server determine that the client's input parameters would effectively violate the protections it has set, then it should reject the test with a user-friendly message to the client which describes why the test was rejected: 
  - `client's requested bit rate exceeds the server's maximum permitted limit`
  - `client's requested duration exceeds the server's maximum permitted limit`
  - `client's requested bytes exceed the server's maximum permitted limit`
- Should the server not be able to determine by the client's parameters that the test will violate the server's protections, then the server should stop the test upon violation and send a user-friendly message to the client to explain why the test was stopped:
  - `client's bit rate exceeded the server's maximum permitted limit`
  - `client's bytes exceeded the server's maximum permitted limit`

Since it is always possible for the server to determine if the client's duration will violate it's protection, it will either reject or permit with success. For other protections such as bit rate or bytes, the server may not be able to determine if its protections will be violated, and will have to react accordingly.

davidBar-On:
@manedurphy, thanks for the detailed response.  In general, I fully agree with what you wrote.

> I believe that the server should not allow the client to run at all when it can determine that it is destined to fail based on the client's input parameters. The same can be (conditionally) said for the --server-bitrate-max flag ...

I agree that both checks should be done , both when receiving the client's parameters and during the test and that the parameters check should be added regarding `--server-bitrate-max`.

> ... then the server should ideally send a user-friendly message to the client. As of now, there is a nice message on the server, `iperf3: error - total required bandwidth is larger than server limit`, but the client has no visibility into what went wrong, `iperf3: error - control socket has closed unexpectedly` ...

I believe that I found a reasonable solution (actually a workaround) to this problem in PR #1909.  Once this PR (or a variant of it) will be merged, the client will (also) print the correct error message.  The idea is that if the server finds a problem with a parameter received from the client, it will send the error code as the next State to client (after making sure that no State and error code have the same value).  When the client receive the "State", instead of just printing `Received unknown state`, it suggests that maybe this is an error code and prints `Received unknown state <state code> from server, which may be the error: <the error string>`.

manedurphy:
@davidBar-On I was thinking of following the same pattern to convey the message. I've pasted a snippet below, but I'll open a new PR which has both the error handling and input validation for the sake of keeping this PR focused on handling the time parameter.

<details>

<summary>snippet</summary>

```c
// Check if average transfer rate was exceeded (condition set in the callback routines)
if (test->bitrate_limit_exceeded) {
    i_errno = IETOTALRATE;
    if (iperf_set_send_state(test, SERVER_ERROR) != 0) {
        cleanup_server(test);
        return -1;
    }

    err = htonl(i_errno);
    if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
        cleanup_server(test);
        i_errno = IECTRLWRITE;
        return -1;
    }
    err = htonl(errno);
    if (Nwrite(test->ctrl_sck, (char*) &err, sizeof(err), Ptcp) < 0) {
        cleanup_server(test);
        i_errno = IECTRLWRITE;
        return -1;
    }

    cleanup_server(test);
    return -1;
}
```

</details>

My latest push here includes writing `errno` as per your requested change.

bmah888:
@manedurphy thanks for your patience in dealing with feedback on this PR! I'm pretty happy with this in the state it's in, except there's a merge conflict with `master` caused by an unrelated PR. Would you be up for fixing that merge conflict on your branch? I can merge it then.

Also I noticed your PR #1931, you probably want that in as well before we do iperf-3.20, right?

manedurphy:
I think all of the PRs that were asked for `3.20` have already been merged, so this and #1931 can be merged whenever you feel they are ready. I've rebased both, but think I'll have to rebase again when one of them is merged.

PR #1687, opened 2024-04-23
Fix clang compilation failure on Android

tqfx:

* Issues fixed (if any):

Fix aarch64-linux-android-clang compilation failure due to patch d80b914141582dc0c0cf3f71cc5799061074be66 #1651

### DIFF
--- a/src/iperf_pthread.c
+++ b/src/iperf_pthread.c
@@ -6,8 +6,27 @@
  * as Android NDK does not support `pthread_cancel()`.
  */
 
+#include <string.h>
 #include <signal.h>
 #include "iperf_pthread.h"
+
+void iperf_thread_exit_handler(int sig)
+{
+    pthread_exit(0);
+}
+
+int iperf_set_thread_exit_handler() {
+    int rc;
+    struct sigaction actions;
+
+    memset(&actions, 0, sizeof(actions));
+    sigemptyset(&actions.sa_mask);
+    actions.sa_flags = 0;
+    actions.sa_handler = iperf_thread_exit_handler;
+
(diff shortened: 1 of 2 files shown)


src/iperf_pthread.h
```diff
@@ -10,12 +10,12 @@
  */
 
 #define PTHREAD_CANCEL_ASYNCHRONOUS 0
-#define PTHREAD_CANCEL_ENABLE NULL
```
tqfx, line 13: error: incompatible pointer to integer conversion passing 'void *' to parameter of type 'int' [-Wint-conversion]

src/iperf_pthread.c
```diff
@@ -6,19 +6,10 @@
  * as Android NDK does not support `pthread_cancel()`.
  */
 
+#include <string.h>
```
tqfx, line 9: error: call to undeclared library function 'memset' with type 'void *(void *, int, unsigned long)'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]

src/iperf_pthread.c
```diff
@@ -6,19 +6,10 @@
...
 #include <signal.h>
 #include "iperf_pthread.h"
 
-int pthread_setcanceltype(int type, int *oldtype) { return 0; }
-int pthread_setcancelstate(int state, int *oldstate) { return 0; }
-int pthread_cancel(pthread_t thread_id) {
-    int status;
-    if ((status = iperf_set_thread_exit_handler()) == 0) {
```
tqfx, line 16: error: call to undeclared function 'iperf_set_thread_exit_handler'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]

Issue #1078, opened 2020-11-24
Iperf3 single-stream low bandwidth with small message sizes (1KB, 1500B, 2000B, etc.)


noamsto:
# Context

* Version of iperf3: iperf 3.5 

* Hardware: 
Nvidia ConnectX-6 Dx 100GbE
Intel  XXV710 25GbE

* Operating system (and distribution, if any): 
Red Hat Enterprise Linux Server release 7.5 (Maipo) - kernel 5.10.0-rc4
CentOS Linux release 8.1.1911 (Core) - kernel 5.2.9

# Bug Report
We are observing low performance with iperf3 while sending single-stream traffic with small message sizes.
Other benchmarks reports significant better results for same scenario, we assume it's Iperf3 that's limiting the performance we observe.

For example:
Running `iperf3 -c  fd12:fd07::200:105  -t 10 -l1500B` we see ~7.9Gbps.
![iperf3_1stream_1500B](https://user-images.githubusercontent.com/17932324/100090710-44bc6b80-2e5c-11eb-8a3d-19f2b09da1c2.jpg)

Running the same test with 1 stream and 1500B with **iperf2** we see 23Gbps which is what we would expect with current system.
cmd: `iperf -c fd12:fd07::200:105 -V -l1500 -P1`
![iperf2_1stream_1500B](https://user-images.githubusercontent.com/17932324/100090915-95cc5f80-2e5c-11eb-9dc1-db7bc4126637.jpg)

* Expected Behavior

Iperrf3 should be able to achieve at least the same performance reported by iperf2 for a single-stream.

* Actual Behavior

Iperf3 small message size performance is low. Iperf3 is the limiting factor.

* Steps to Reproduce

Send traffic with message size 1500B (On a high performance NIC over 10/25Gbps).

* Possible Solution

I've noted iperf2 suffer a similiar issue when running with '-i' flag, might be related to the reporting/results gathering flow?

davidBar-On:
@noamsto, please re-run the test with the following options and let us know the results:
1. Server verbose and debug (`-d -V`) option to see the size of the send/receive buffers.
2. If you suspect that the statistics reports may be an issue, run both server and client with no interval statistics: `-i 0`.
3. What is the throughput if you use the default message size `-l 128K`?

noamsto:
Hi @davidBar-On,

1. Attached the following output when running server command `iperf3 -Vsd`
[...]

PR #1691, opened 2024-04-30
Increase TCP bandwidth for small messages

davidBar-On:
master

* Issues fixed (if any): #1078

Suggested enhancement to resolve the iperf3 TCP low bandwidth with small message sizes, compared to iper2 and netperf.  This is by receiving all the sent burst messages as one read message.  In my environment, for `-l1500` throughput is increased by about 35% for a single stream and more than 50% for multi streams tests.

It seems that the main reason for iperf2 and netperf higher bandwidth for small messages is that iperf3 is sending and receiving the same message size, while in iperf2, and probably netperf, they are different.  For example, iperf2 default receive TCP message size is 128K.

Notes:
1. Receiving all the burst messages as one message is assumed to be o.k. (and not "cheating"), based of iperf2 (and netperf) behavior.
2. The TCP receive message size is extended (by multiplying `blksize` by the burst size) to maximum `MAX_BLOCKSIZE` (1MB).  If this is too large, it may be limited to `MAX_TCP_BUFFER` (512KB)or `DEFAULT_TCP_BLKSIZE` (128KB).
3. One of the advantages of using the sent burst size for the received message is that sending and receiving are in sync.  However, other approaches may be used, e.g. adding additional value to `-l` with the receiver message length, i.e. the first value is for the sender size (and is the receiver default).  This will be similar to the iperf2 approach, but I believe that using burst is better.
4. There is an issue with using the burst size, which actually already exists in iperf3 before this PR.  When setting the the test bandwidth using `-b`, the default burst size is set to 1.  Assume that the network bandwidth is 1Gbps.  For test that does not set `-b` the burst size is 10, and for test with `-b10G` the burst size is 1.  Therefore, the first test will have higher bandwidth, although practically the second test didn't put a real limit to the bandwidth.  A workaround is also setting the burst size by `-b10G/10`, but still users may not understand this burst size difference.  (By the way, with multi-thread the sending burst may be redundant, so it may be possible to remove the burst loop when sending.  With this approach, burst size will only be applicable to TCP receives, maybe allowing to not changing the default burst to 1 when `-b` is set.)

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -146,7 +146,7 @@ struct iperf_settings
 {
     int       domain;               /* AF_INET or AF_INET6 */
     int       socket_bufsize;       /* window size for TCP */
-    int       blksize;              /* size of read/writes (-l) */
+    int       blksize;              /* size of read/writes (-l); specific stream read size may be different */
     iperf_size_t  rate;                 /* target data rate for application pacing*/
     iperf_size_t  bitrate_limit;   /* server's maximum allowed total data rate for all streams*/
     double        bitrate_limit_interval;  /* interval for averaging total data rate */
@@ -190,6 +190,7 @@ struct iperf_stream
     int       socket;
     int       id;
     int       sender;
+    int       blksize;         /* Same as Test blksize, except for send TCP read size may be extended by burst */
 	/* XXX: is settings just a pointer to the same struct in iperf_test? if not,
 		should it be? */
     struct iperf_settings *settings;	/* pointer to structure settings */
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1935,12 +1935,7 @@ iperf_send_mt(struct iperf_stream *sp)
     int no_throttle_check;
 
     /* Can we do multisend mode? */
-    if (test->settings->burst != 0)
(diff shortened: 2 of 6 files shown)


davidBar-On:
Few comments that may help the evaluation:
> it's important to consider if we're pushing bytes as quickly as possible or measuring the performance of the entire system with sending and receiving these differently-sized messages. We tend to prefer the latter.

The suggested change is only in the **receiving side**.  There is no change to the way the bytes a pushed.  If I understood iperf2 code correctly, this is how it works, i.e. sending each message separately but receiving bytes as they arrive.  The approach is also similar to the `SKIP-RX-COPY` approach that the overhead of getting the messages at the receiving side may be ignored.

> For example, we removed most of the select calls from the send and receiving loops, since they were interfering with the measurement.

As I mentioned in the description, because of this change the burst size has no meaning practically.  Therefore, it seems that the **default burst size can be changed to 1 with no impact on iperf3 performance**.  In this case, by default this PR will not change iperf3 behavior (because the change is to read "burst" number of messages).  Only if the user will set the burst size, iperf3 behavior and performance will change.  If this approach is desired I can enhance the PR code accordingly.

PR #1708, opened 2024-06-02
Do not listen to old UDP prot_listener

davidBar-On:
master

* Issues fixed (if any): none

Clear old UDP `prot_listener` from `read_set`, as currently the server's `select()` always return when data (that is relevant only to the worker thread) is received.  This makes the main thread consume a lot of unnecessary CPU, which reduces the total throughput - especially when only one stream is used.

In my single machine environment, with the this fix the throughput is increased by about 40% for one stream test.

### DIFF
--- a/src/iperf_udp.c
+++ b/src/iperf_udp.c
@@ -446,6 +446,7 @@ iperf_udp_accept(struct iperf_test *test)
     /*
      * Create a new "listening" socket to replace the one we were using before.
      */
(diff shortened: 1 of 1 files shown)


bmah888 (approved):
Thanks for the pull request! This change looks good, approved. I had mixed results trying to reproduce the performance improvements you saw, but I think this PR makes the functionality of the program more correct regardless.

PR #1709, opened 2024-06-02
Json read check json size

davidBar-On:
latest

* Issues fixed (if any): None

When receiving Params JSON size, verify that received exactly the `sizeof(<size variable>)` bytes.

In addition, added a check that the size received is reasonable - positive and is less than 8KB (which probably is much more that needed and may be reduced).

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2820,7 +2820,7 @@ JSON_read(int fd)
                             json = cJSON_Parse(str);
                         }
                         else {
-                            printf("WARNING:  JSON size of data read does not correspond to offered length\n");
+                            warning("JSON size of data read does not correspond to offered length");
                         }
 	            }
 	            free(str);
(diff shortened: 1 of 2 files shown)


src/iperf_api.c
```diff
@@ -2728,8 +2728,9 @@ JSON_read(int fd)
      * Then read the JSON into a buffer and parse it.  Return a parsed JSON
      * structure, NULL if there was an error.
      */
-    if (Nread(fd, (char*) &nsize, sizeof(nsize), Ptcp) >= 0) {
-	hsize = ntohl(nsize);
+    rc = Nread(fd, (char*) &nsize, sizeof(nsize), Ptcp);
+    hsize = ntohl(nsize);
+    if (rc == sizeof(nsize) && hsize > 0 && hsize <= MAX_PARAMS_JSON_STRING) {
```
swlars, line 2733: Hey David, thank you for the pull request. We like this change, but we would prefer it if the rc check was separate and before we use the value of nsize. If Nread fails, it would be better to just fall through. 

```
if (rc == sizeof(nsize)){
    hsize = ntohl(nsize);
    if (hsize > 0 && hsize <= MAX_PARAMS_JSON_STRING) {
        ...
    }
}
```
Also, I realize it's not there now, but adding a print statement else if it fails the size checks would be helpful. Let us know what you think!

swlars:
Perfect! Thanks again for the pull request!

davidBar-On:
@swlars, unfortunately this PR has a bug which was discovered by issue #1776.  Submitted PR #1779 with a fix.

PR #1714, opened 2024-06-08
fix -Wformat-y2k error

neheb:
%c potentially prints a 2 digit year. Just use a normal format.

all

* Issues fixed (if any):
* -Wformat-y2k error

%c potentially prints a 2 digit year. Just use a normal format.


bmah888:
The problem here isn't exactly that `%c` is used as a format string, but more that we allow the user to specify an argument to `--timestamp` that includes the user's specification of a format string (i.e. `--timestamp="%Y-%m-%d %H:%M:%S"`). Going to fix that instead.

bmah888 (approved):

bmah888:
Pushed amended pull request. Thanks!

PR #1717, opened 2024-06-16
Add SKIP-RX-COPY support - using MSG_TRUNC socket option

davidBar-On:
* Version of iperf3 (or development branch, such as `master` or
  `3.1-STABLE`) to which this pull request applies:
master

* Issues fixed (if any):
#1678 (partially)

* Brief description of code changes (suitable for use as a commit message):

Add SKIP-RX-COPY support - using MSG_TRUNC socket option to UDP and TCP.  This is a subset of the original PR #1690 that originally supported both skip-rx-copy and zerocopy for UDP.  Because of UDP related issues found by @MattCatz, I split the handling of the two solutions into two PRs.

The added option is `--skip-rx-copy`: for both TCP and UDP, `recv(..., MSG_TRUNC)` is used instead of `read()`.

After the [changes suggested by @MattCatz](https://github.com/esnet/iperf/pull/1690#issuecomment-2161519302) the PR code is mature enough.  
(Note that running bootstrap.sh; configure is required before make to support the new features.)

Note that whether the feature enhance performance seems to depend on the test environment.  However, th e support for the feature also allows to test its impact in the specific environment.

PR #1720, opened 2024-06-23
Add SO_ZEROCOPY and MSG_ZEROCOPY support for UDP and TCP

davidBar-On:
master

* Issues fixed (if any): #1678

Add support for zero-copy for both UDP and TCP, using SO_ZEROCOPY/MSG_ZEROCOPY.  Although it is not clear when and in what environments the added functionality improve performance, support was added to allow testing this functionality.  The assumption is that different environments may have different levels of support for this functionality.
(Note that running `bootstrap.sh; configure` is required before `make` to support the new features.)

The added support is for `MSG_ZEROCOPY`. When used, socket option `SO_ZEROCOPY` is set and `send(...., MSG_ZEROCOPY)` is used instead of `write()`.  `MSG_ZEROCOPY` is used in the following cases:
- UDP: when `-Z/--zerocopy` option is set.
- TCP: when `--zerocopy=z` is set.  Otherwise, `sendfile()` continue to be used for TCP zero copy.

This PR is a rework of the original suggested support in #1690, as it was too naive and did not include support for notifications.

The main implementation is based on the [msg_zerocopy.c example](https://github.com/torvalds/linux/blob/master/tools/testing/selftests/net/msg_zerocopy.c) for supporting `MSG_ZEROCOPY`.

Note that per the [Zero-copy networking discussion](https://lwn.net/Articles/726917/) about the [patch that added MSG_ZEROCOPY support](https://lwn.net/Articles/726353/), the expected performance improvement is expected to be less than 10% and only for large packets.  Therefor, it would be helpful if this PR will be tested on different environments to see whether and when it improves performance.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -337,6 +337,20 @@ if test "x$iperf3_cv_header_tcp_info_snd_wnd" = "xyes"; then
   AC_DEFINE([HAVE_TCP_INFO_SND_WND], [1], [Have tcpi_snd_wnd field in tcp_info.])
 fi
 
+# Check for MSG_ZEROCOPY (mostly on Linux)
+AC_CACHE_CHECK([MSG_ZEROCOPY send option],
+[iperf3_cv_header_msg_zerocopy],
+AC_COMPILE_IFELSE(
+  [AC_LANG_PROGRAM([[#include <sys/types.h>
+                     #include <sys/socket.h>
+                     #include <netinet/in.h>]],
+                   [[int foo = MSG_ZEROCOPY;]])],
+  iperf3_cv_header_msg_zerocopy=yes,
+  iperf3_cv_header_msg_zerocopy=no))
+if test "x$iperf3_cv_header_msg_zerocopy" = "xyes"; then
+    AC_DEFINE([HAVE_MSG_ZEROCOPY], [1], [Have MSG_ZEROCOPY send option.])
+fi
+
 # Check if we need -lrt for clock_gettime
 AC_SEARCH_LIBS(clock_gettime, [rt posix4])
 # Check for clock_gettime support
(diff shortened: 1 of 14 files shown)


davidBar-On:
Closing, as it is not clear that this functionality is valuable, as I wrote in the description. (Code is still available for testing or coming up with enhanced solution.)

marcosfsch:
MSG_ZEROCOPY does improves performance a lot for large TCP flows, specially when used with pacing. Brian Tiernery and me, made some tests and the results can be checker in [1, 2].
I'll run new UDP tests to provide evidence if there are improvements and I hope that this patch can be upstreamed.

[1] https://github.com/marcosfsch/INDIS-2024/blob/main/AmLight/ubuntu22-6.8/summary.txt
[2] https://github.com/marcosfsch/INDIS-2024/blob/main/ESnet/ubuntu22-6.8/summary.txt

marcosfsch:
Some preliminary UDP results show 18% better performance for UDP on the TX side. From 36G to 42G for single UDP flow.

bltierney:
Also, to get improvements with MSG_ZEROCOPY, you have to set this in /etc/sysctl.conf:

 net.core.optmem_max = 1048576

But we are finding MSG_ZEROCOPY helps a lot on 100G hosts.

davidBar-On:
OK.  Thanks for the input.  @bmah888, can you please re-open this PR?

bltierney:
I should also mention that we were using your original patch 1690, not patch 1720. Seems like this should not matter, but thought I should mention that.

davidBar-On:
> I should also mention that we were using your original patch 1690, not patch 1720. Seems like this should not matter, but thought I should mention that.

That is interesting to know.  As far as I remember that should not matter.  The reason I split #1960 to this PR and #1717 (SKIP-RX-COPY) is that per my internal tests I assumed there is more chance that SKIP-RX-COPY will be merged into the mainline.  Per this discussion I now understand this may not be the case, but still I assume it is easier to merge less changes per each PT than merge all.

Issue #1719, opened 2024-06-20
support for 64bit fq-rate


bltierney:
Starting with Linux 5.x, fq pacing is now supported for speeds > 32G.
see: https://patchwork.ozlabs.org/project/netdev/patch/20190228231728.242461-2-edumazet@google.com/
and https://kernelnewbies.org/Linux_5.1

This change is needed in iperf3 to support this:

sed -i -e 's/unsigned int fqrate/uint64_t fqrate/g' -e 's/Setting fair-queue socket pacing to %u/Setting fair-queue socket pacing to %"PRIu64"/g' src/iperf_tcp.c src/iperf_udp.c

PR #1728, opened 2024-07-10
Changed fqrate from uint to uint64 to allow pacing above 32G

marcosfsch:
master

* Issues fixed (if any):
https://github.com/esnet/iperf/issues/1719

Starting with Linux 5.x, fq pacing is now supported for speeds > 32G.
see: https://patchwork.ozlabs.org/project/netdev/patch/20190228231728.242461-2-edumazet@google.com/
and https://kernelnewbies.org/Linux_5.1
To fix this we only need to change fqrate variable from uint to uint64 on both perf_tcp.c and iperf_udp.c

### DIFF
--- a/src/iperf_tcp.c
+++ b/src/iperf_tcp.c
@@ -131,10 +131,10 @@ iperf_tcp_accept(struct iperf_test * test)
 
     if (test->settings->fqrate) {
 	/* Convert bits per second to bytes per second */
-	unsigned int fqrate = test->settings->fqrate / 8;
+	uint64_t fqrate = test->settings->fqrate / 8;
 	if (fqrate > 0) {
 	    if (test->debug) {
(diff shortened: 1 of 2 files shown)


swlars:
Hi, we like this change, but we're concerned about compatibility on supported devices with a Linux kernel version that's less than 5, like Rocky8. Do you know of a way to detect at run or compile time if 64-bit values are accepted?

marcosfsch:
I don't know a way to test the values at runtime, but I just compiled and runned iperf3 with this patch and it silently trims the value to 32-bits on rocky8, and mantains the behaviour without tha patch. i.e. if you set above 32G it overloads and start from zero again.

swlars:
We also tested it and it seems to work as expected. We have not tested on any big-endian systems, so we'll probably make a note of that somewhere. Thanks for the merge request!

Issue #1726, opened 2024-07-05
Include RTT information in SCTP tests `--json` as well


UN0910:
# Context

* Version of iperf3: 3.17.1

* Operating system (and distribution, if any): Linux

# Enhancement Request

* Current behavior
- Since version 3.10.1, iperf3 supports RTT information in TCP tests, but this information is missing in SCTP tests.

* Desired behavior
- After obtaining the `--json` output from SCTP tests, the information should also include RTT details.

PR #1731, opened 2024-07-14
Add SCTP information to --json output

davidBar-On:
master

* Issues fixed (if any): #1726

Add SCTP information to JSON output: RTT, PMTU, WND, CWND.

JSON creation is using the same code used to create the TCP output, so some fields have value 0, e.g. VARRTT.  The assumption is that the SCTP fields have the same meaning as the TCP fields.  Especially, WND for TCP is the "send windows" while the SCTP is taken from "receive window", but I hope that their meaning is the same.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -83,7 +83,6 @@ typedef atomic_uint_fast64_t atomic_iperf_size_t;
 typedef unsigned int uint
 #endif // __vxworks or __VXWORKS__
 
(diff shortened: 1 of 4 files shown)


bmah888 (approved):
Thanks for the PR! Tested on Linux (Ubuntu 20.04 LTS) and FreeBSD 13. +1 to merge this.

PR #1734, opened 2024-07-16
Add debug messges when State changes

davidBar-On:
Master

* Issues fixed (if any): None

Added DEBUG_LEVEL_INFO debug messages when state changes.  I found these debug messages useful in different error cases, especially when there is a problem with the client/server interface.

### DIFF
--- a/src/iperf_util.c
+++ b/src/iperf_util.c
@@ -595,3 +595,30 @@ getline(char **buf, size_t *bufsiz, FILE *fp)
 }
 
 #endif
+
+/* Translate numeric State to text - for debugging pupposes */
+char *
+state_to_text(signed char state)
+{
+    char *txt;
+
+    switch (state) {
+        case 0: txt = "Test reset"; break;
+        case TEST_START: txt = "TEST_START - starting a new test"; break;
+        case TEST_RUNNING: txt = "TEST_RUNNING"; break;
+        case TEST_END: txt = "TEST_END"; break;
+        case PARAM_EXCHANGE: txt = "PARAM_EXCHANGE - Client to Server Parameters Exchange"; break;
+        case CREATE_STREAMS: txt = "CREATE_STREAMS"; break;
+        case SERVER_TERMINATE: txt = "SERVER_TERMINATE"; break;
+        case CLIENT_TERMINATE: txt = "CLIENT_TERMINATE"; break;
+        case EXCHANGE_RESULTS: txt = "EXCHANGE_RESULTS"; break;
+        case DISPLAY_RESULTS: txt = "DISPLAY_RESULTS"; break;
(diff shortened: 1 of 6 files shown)


bmah888:
Thanks for the pull request! Just having glanced over it, this seems like a good thing to have, especially since the test state machine isn't super-well-documented and I personally regard it as a bit of magic.

davidBar-On:
> ... especially since the test state machine isn't super-well-documented ...

To better help with this I added a very short textual description of the states to the debug messages (and added some debug messages in additional places where the state is changed).

src/iperf_util.c
```diff
@@ -595,3 +595,29 @@ getline(char **buf, size_t *bufsiz, FILE *fp)
...
+        case 11: txt = "Server Terminated"; break;
+        case 12: txt = "Client Terminated"; break;
+        case 13: txt = "Exchanging Results"; break;
+        case 14: txt = "Displaying Results"; break;
+        case 15: txt = "Starting"; break;
+        case 16: txt = "Done"; break;
+        case -1: txt = "Error"; break;
+        case -2: txt = "Serevr Error"; break;
+        default: txt = "Unknown State";
+    }
```
bmah888, line 620: I think this will be really helpful in localizing issues related to the control channel or test control.

Instead of numeric constants in this switch statement, what about using the definitions such as `TEST_START`, `TEST_RUNNING`, etc. from `iperf_api.h`?

I'm wondering if the text strings should just be the names of those constants, e.g. `"TEST_START"` but I don't feel super-strongly about that.

Finally in `case -2`, the text string should be `"Server Error"`.

bmah888 (changes requested):
Thanks for the PR! I think this will be very helpful for developers.

src/iperf_util.c
davidBar-On: > Instead of numeric constants in this switch statement, what about using the definitions such as TEST_START, TEST_RUNNING, etc. from iperf_api.h?

Oops! Not sure why I used the numeric values...  Changed.

> I'm wondering if the text strings should just be the names of those constants, e.g. "TEST_START" but I don't feel super-strongly about that.

I was also not sure about the right approach.  I now changed that to a mixed text - using the names for all the values with added text where the name is not clear enough.  Also, added "Test reset" for value 0.

In addition, to reduce number of debug print statements, when possible I changed the `test->state = <new state>` to `iperf_set_test_state(test, <new state>)` function call and added the debug message in that function.  Also added debug messages before reading a new state.

bmah888: Thanks for the improvements, I liked the slight refactoring in `iperf_set_test_state()`.

bmah888 (approved):
+1 to merge this

PR #1744, opened 2024-08-13
Fix rcv-timeout issue because of Nread timeout

davidBar-On:
master

* Issues fixed (if any): None

Fix an issue with `--rcv-timeout` - did not work with value over 10 seconds (my first PR for a problem I found at work as iperf3 user :smile:).  This is because `Nread()` timeout after 10 seconds, but `iperf_recv_mt()` collected statistics even when receiving 0 bytes.  Therefore, it was assumed that data blocks where received and so receive did not timeout.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2000,8 +2000,14 @@ iperf_recv_mt(struct iperf_stream *sp)
 		i_errno = IESTREAMREAD;
 		return r;
 	    }
-	    test->bytes_received += r;
-	    ++test->blocks_received;
+            
+            /* Collect statistics only if receive did not timeout (e.g. `Nread()` may timeout).
+             * This is also important for `--rcv-timeout` to work properly.
+             */
+            if (r > 0) {
+	        test->bytes_received += r;
+	        ++test->blocks_received;
+            }
 
     return 0;
 }


davidBar-On:
@bmah888, as this is a bug fix (although not critical), just making sure it was not overlooked.

bmah888:
Thanks for the PR!

Documenting how I reproduced this bug. With a built from tip of the `master` branch:

`server% iperf3 --server --rcv-timeout=2000`
`client% iperf3 --client server.example.com --time=60`

Suspend the client process after a few seconds, observe the timeout on the server side after 2 seconds (approximately).

`server% iperf3 --server --rcv-timeout=15000`
`client% iperf3 --client server.example.com --time=60`

Suspend the client process after a few seconds, the server will never timeout and also will never terminate the test.

bmah888:
With this PR, testing the above scenario results in the server terminating successfully.

Also tested this scenario:

`server% iperf3 --server`
`client% iperf3 --client server.example.com --time=60 --rcv-timeout=15000`

After a few seconds, suspend the server side process. Without this PR, the client will not honor the `--rcv-timeout` parameter, and it runs to the end of the `--time` parameter and hangs.

With the PR, the `--rcv-timeout` parameter works correctly.

This looks good for a merge.

bmah888 (approved):

Issue #1750, opened 2024-08-23
 Segmentation fault under running kernel's netfilter concurrency testing


xTire:
# Context

* Version of iperf3:  3.17.1+ (cJSON 1.7.15)

* Hardware: x86_64&aarch64

* Operating system (and distribution, if any): openEuler & openSUSE leap 15.6

* Other relevant information :  

1. The number of processor cores exceeds 16. The more cores, the more likely the problem is to be triggered;
2. Physical machines are much easier to reproduce than virtual machines

# Bug Report

* Expected Behavior
   The kernel self-test script has completed running

* Actual Behavior
   When executing the concurrent test phase, a segmentation fault occurred in iperf3

```
TEST: concurrency
  net,port                                                      [ OK ]
  port,net                                                      [ OK ]
  net6,port                                                     ./nft_concat_range.sh: line 456: 799955 Segmentation fault      (core dumped) ip netns exec B "$@" > /dev/null 2>&1
./nft_concat_range.sh: line 456: 799958 Segmentation fault      (core dumped) ip netns exec B "$@" > /dev/null 2>&1
./nft_concat_range.sh: line 456: 799965 Segmentation fault      (core dumped) ip netns exec B "$@" > /dev/null 2>&1
[ OK ]
  net,port,net                                                  [ OK ]
  net6,port,net6,port                                           [ OK ]
TEST: timeout
  net,port                                                      [ OK ]
```

* Steps to Reproduce
1. Download Linux kernel self test script:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/plain/tools/testing/selftests/netfilter/nft_concat_range.sh?h=v5.10.224
3. Running the script multiple times in a row: 
`for i in $(seq 5); do ./nft_concat_range.sh; done`

* Call trace
```
warning: Can't open file /tmp/iperf3.74s4y2 (deleted) during file-backed mapping note processing
[New LWP 2683041]
[New LWP 2682985]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib64/libthread_db.so.1".
Core was generated by `iperf3 -s -DB 2001:db8::0000:000c -p 13'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000000392f1cb9 in ?? ()
[Current thread is 1 (Thread 0x7f38bd3f5640 (LWP 2683041))]
(gdb) bt
#0  0x00000000392f1cb9 in ?? ()
[...]

PR #1752, opened 2024-08-26
Avoid subthread signal handling

xTire:

```
```

* Version of iperf3:  master

* Issues fixed (if any): #1750

* Brief description of code changes:
The main and sub-threads responding to the signal processing function at the same time will cause the sub-thread to access the test instance that has been released by the main thread, vice versa.

### DIFF
--- a/src/iperf_api.h
+++ b/src/iperf_api.h
@@ -474,8 +474,8 @@ enum {
     IEPTHREADCANCEL=151,        // Unable to cancel thread (check perror)
     IEPTHREADJOIN=152,		// Unable to join thread (check perror)
     IEPTHREADATTRINIT=153,      // Unable to initialize thread attribute (check perror)
-    IEPTHREADSIGMASK=154,      // Unable to initialize sub thread signal mask (check perror)
-    IEPTHREADATTRDESTROY=155,      // Unable to destroy thread attribute (check perror)
+    IEPTHREADATTRDESTROY=154,      // Unable to destroy thread attribute (check perror)
+    IEPTHREADSIGMASK=155,      // Unable to initialize sub thread signal mask (check perror)
     /* Stream errors */
     IECREATESTREAM = 200,   // Unable to create a new stream (check herror/perror)
     IEINITSTREAM = 201,     // Unable to initialize stream (check herror/perror)
--- a/src/iperf_api.h
+++ b/src/iperf_api.h
@@ -474,7 +474,8 @@ enum {
     IEPTHREADCANCEL=151,        // Unable to cancel thread (check perror)
     IEPTHREADJOIN=152,		// Unable to join thread (check perror)
     IEPTHREADATTRINIT=153,      // Unable to initialize thread attribute (check perror)
-    IEPTHREADATTRDESTROY=154,      // Unable to destroy thread attribute (check perror)
+    IEPTHREADSIGMASK=154,      // Unable to initialize sub thread signal mask (check perror)
+    IEPTHREADATTRDESTROY=155,      // Unable to destroy thread attribute (check perror)
     /* Stream errors */
     IECREATESTREAM = 200,   // Unable to create a new stream (check herror/perror)
     IEINITSTREAM = 201,     // Unable to initialize stream (check herror/perror)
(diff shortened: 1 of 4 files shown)


xTire:
as mentioned in https://github.com/esnet/iperf/issues/1750#issuecomment-2313137172, this PR case client cannot exit in response to SIGTERM signal.

bmah888:
Thanks for the pull request! I'm looking this over. Also figuring out how to test this in my environment (the directions in #1750 were clear, I'm just not sure how to best do this with the resources I have available). Probably we'll want to squash and merge these two commits when we merge them to `master`.

xTire:
> as mentioned in [#1750 (comment)](https://github.com/esnet/iperf/issues/1750#issuecomment-2313137172), this PR cause client cannot exit in response to SIGTERM signal.

[33a4b96](https://github.com/esnet/iperf/pull/1752/commits/33a4b96271a476d34a893879ca98487c37d915ab) should fix it

davidBar-On:
> [33a4b96](https://github.com/esnet/iperf/pull/1752/commits/33a4b96271a476d34a893879ca98487c37d915ab) should fix it

I tested this version, and it seems to works o.k.😄 For both the server and the client (using -P4) I tried to kill the main thread or a "worker" thread and in all cases termination was as expected. I guess that anyway it will also be good to also add a `mutex()` to protect the `exit()` in `iperf_errexit()`.

Itis-hard2name:
The logic for setting SIG_BLOCK in the two code segments is identical. How can we consolidate this into a single function so that both the client and server sides can call this function, making the code look more elegant?
Would you like a suggestion on how to refactor the code to achieve this?

bmah888:
I apologize for the nitpicking to come: It would not have occurred to me to put in the conditional compilations in 33a4b96, because those signal definitions were at least as old as the POSIX.1-1990 standard. (Anecdotally I remember using them on BSD4 systems in the late 1980s.) I almost feel like they're not really needed, and removing them will make the source code a little easier to read and follow. Like if you try to build this on a system that for some reason doesn't have these signal definitions (such as `SIGTERM`, etc.) it's probably not very UNIX-like. Am I missing something here?

I'm trying to test the code, but I don't really have a suitable environment set up to reproduce the original problem. It seems to work OK in what limited testing I can do though, so that's good.

xTire:
The conditional compilation was added just because the [original signal capture function registration](https://github.com/esnet/iperf/blob/7679199ec99c5db194c554b7b11066d347891735/src/iperf_api.c#L4826) was defined like this.  This is done simply to not break existing facilities and maintain consistency.

For current Linux systems, I think [da21dc2](https://github.com/esnet/iperf/pull/1752/commits/da21dc205821d96187026563237a374e0fe7c260) is enough.

bmah888:
> The conditional compilation was added just because the [original signal capture function registration](https://github.com/esnet/iperf/blob/7679199ec99c5db194c554b7b11066d347891735/src/iperf_api.c#L4826) was defined like this. This is done simply to not break existing facilities and maintain consistency.
> 
> For current Linux systems, I think [da21dc2](https://github.com/esnet/iperf/pull/1752/commits/da21dc205821d96187026563237a374e0fe7c260) is enough.

Ah actually after a little research I see the other conditional compilation was originally added as cherry-pick of a part of PR #935. Let's keep 33a4b96 as a part of the PR. Thanks!

src/iperf_api.h
```diff
@@ -474,7 +474,8 @@ enum {
     IEPTHREADCANCEL=151,        // Unable to cancel thread (check perror)
     IEPTHREADJOIN=152,		// Unable to join thread (check perror)
     IEPTHREADATTRINIT=153,      // Unable to initialize thread attribute (check perror)
-    IEPTHREADATTRDESTROY=154,      // Unable to destroy thread attribute (check perror)
+    IEPTHREADSIGMASK=154,      // Unable to initialize sub thread signal mask (check perror)
+    IEPTHREADATTRDESTROY=155,      // Unable to destroy thread attribute (check perror)
```
bmah888, line 478: Could you leave `IEPTHREADATTRDESTROY` with its original numerical value of 154, and just add `IEPTHREADSIGMASK` as 155? It probably doesn't make a huge difference but it's probably better to not change existing definitions if there's not a reason to.

bmah888:
If you can fix the nitpick I just posted about `IEPTHREADSIGNALMASK`, I'm leaning towards doing a squash-and-merge on all these changes. Thanks!

src/iperf_api.h
xTire: Sure, there will be a push later after local compilation and testing.

bmah888:
Thanks! Squash and merged.

PR #1755, opened 2024-08-28
fix incorrect freeaddrinfo call

kulikjak:
`freeaddrinfo` call introduced with https://github.com/esnet/iperf/commit/111212bca2eaab948cff11fafddcb097e7615f52 is incorrect, because if `getaddrinfo` failed, there is nothing in there to free.

We encountered segfault there on Solaris, but I can reproduce that on Linux the pointer remains `NULL` as well when `getaddrinfo` fails.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -145,7 +145,6 @@ create_socket(int domain, int proto, const char *local, const char *bind_dev, in
     if ((gerror = getaddrinfo(server, portstr, &hints, &server_res)) != 0) {
 	if (local)
 	    freeaddrinfo(local_res);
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR!

The line of code in question was added in PR #1622.

bmah888 (approved):
This makes sense, we shouldn't be trying to do `freeaddrinfo(NULL)`.

neheb:
Without this, instant crash on musl with

```
iperf3 -c server -P 4 -t 30
```

Issue #1760, opened 2024-09-09
sigfault when multi thread repeated recycling in client


answer9030:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the Discussions section for this project at
https://github.com/esnet/iperf/discussions
or to the iperf3 mailing list at iperf-dev@googlegroups.com.
A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3: 3.16

* Hardware:

* Operating system (and distribution, if any): centos

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.): glibc-2.38-29

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate. Note that submissions of bug fixes, new
features, etc. should be done as a pull request at
https://github.com/esnet/iperf/pulls_

# Bug Report
Long duration streaming with multiple threads resulted in segment fault on the client side at the end of the test.

Use the following command for testing:
    server side: iperf -s
    client side:  iperf -c serverIP -t 2400 -P 8

At the end of the test, client side encountered segment fault.

* Expected Behavior
Test completed normally.

* Actual Behavior
1. The testing environment spans multiple network nodes.
2. iperf's management connection does not send packets for a long time, triggering a timeout for connectioin tracking on intermediate node.
3. At the end of the test, the client reclaimed the thread and sent an end message using the management connection, but the message was discarded by the intermediate link.
4. Resulting in management connection timeout, triggering thread recycling again.
5. When the number of threads reaches 5 or more, glibc has already unmapped the address pointed to by “sp->thr”, resulting in a segment fault during the second thread recycling.

* Steps to Reproduce
Use the following command for testing:
    server side: iperf -s
    client side:  iperf -c serverIP -t 2400 -P 8

* Possible Solution

# Enhancement Request

* Current behavior

* Desired behavior

* Implementation notes

bmah888:
First and foremost, iperf3 shouldn't crash. So there's clearly something wrong.

[...]

PR #1761, opened 2024-09-10
Avoid duplicate thread recycling.

answer9030:
At the end of the test, the traffic thread has been reclaimed. If there is an exception in the control connection, it will cause the thread to be reclaimed repeatedly. Use sp->done to avoid repeated thread recycling.

```
```

* Issues fixed (if any): https://github.com/esnet/iperf/issues/1760

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -807,6 +807,9 @@ iperf_run_client(struct iperf_test * test)
     /* Cancel all outstanding threads */
     i_errno_save = i_errno;
     SLIST_FOREACH(sp, &test->streams, streams) {
+        if (sp->done) {
+            continue;
+        }
         sp->done = 1;
         int rc;
         rc = pthread_cancel(sp->thr);


bmah888 (commented):
Thanks for the PR! I think the change makes sense...the only way that `sp->done` gets set to 1 would be if we're going to terminate the corresponding thread.

I'd probably just check for `sp->done` rather than a check for a specific value (`sp->done == 1`), this would be consistent with the other checks against `sp->done`.

There are two other places in `iperf_client_api.c` and one in `iperf_server_api.c` that do a similar iteration over all (or most) of the streams and try to `pthread_cancel()` their corresponding threads. Do you think that similar checks for `sp->done` would be appropriate in those cases as well?

answer9030:
I think there is no need to perform similar checks on `sp ->done` in other places, as these places will not trigger duplicate thread recycling.

bmah888:
The original scenario for #1760 is a little difficult for me to reproduce exactly, but thanks for the detailed explanation of what triggers the segfault. I settled for tested this by more-or-less forcing the error handling on the client to run. I replaced this code in `iperf_client_api.c` (function `iperf_run_client`):
```
		if (iperf_set_send_state(test, TEST_END) != 0)
                    goto cleanup_and_fail;
```
with:
```
iperf_set_test_state(test, TEST_END);
//		if (iperf_set_send_state(test, TEST_END) != 0)
                    goto cleanup_and_fail;
```
I was able to reproduce the original segfault and as well as verify that the client exits sanely.

bmah888 (approved):

Issue #1756, opened 2024-09-03
Iperf3: connection is getting closed from the client side unexpectedly


Ankita13-code:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the Discussions section for this project at
https://github.com/esnet/iperf/discussions
or to the iperf3 mailing list at iperf-dev@googlegroups.com.
A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context
iperf3 error: the client has unexpectedly closed the connection 
* Version of iperf3: 3.17.1

* Hardware:

* Operating system (and distribution, if any): Azure Linux/CBL-Mariner

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate. Note that submissions of bug fixes, new
features, etc. should be done as a pull request at
https://github.com/esnet/iperf/pulls_

# Bug Report

* Expected Behavior
The lisa test case [perf_udp_iperf_sriov](https://github.com/microsoft/lisa/blob/d97439f4afe688a4e410b3090ab0dac3b09c5da1/microsoft/testsuites/performance/networkperf.py#L281) uses iperf3 to test the network throughput between 2 Azure Linux VMs. This test should pass for Azure Linux VMs. 
* Actual Behavior
The test is intermittently failing with the following error - `error: the client has unexpectedly closed the connection`.  It fails on the following commands - 
Client side - `sudo iperf3 -t 10 -c <server_ip> -u  -J  -i1  -f g  -p 757  -P 64 -l 8192  -4` 
Server side - `sudo iperf3  -s -1  -J  -i10  -f g  -p 757` 
* Steps to Reproduce
-- Spin up 8 iperf servers at the same time in parallel (we are using port 750-757 here) - `sudo iperf3  -s -1  -J  -i10  -f g  -p <port>`
-- Now run 8 clients in parallel with the following parameters - `sudo iperf3 -t 10 -c <server_ip> -u  -J  -i1  -f g  -p <port> -P 64 -l 8192  -4` 
[...]

PR #1765, opened 2024-09-14
No Server error when connection closed while waiting for IPERF_DONE

davidBar-On:
master

* Issues fixed (if any): #1756

It seems that the problem in #1756 is that the server "received" the client's close of the control channel, before the `IPERF_DONE` arrived to the server.  Since this is not really an error, the change is that in this case the server just print a warning message , instead of issuing an error.


davidBar-On:
Apparently, the issue in #1756 is not related to this PR.  Therefore, closing the PR as it is not clear if it has a real use case.

PR #1773, opened 2024-09-19
Remove the usage of pacing_timer and simplify iperf_mt_send

davidBar-On:
master

* Issues fixed (if any): discussion #1736

**Removed the usage of `pacing_timer`**, and keep it as a parameter for backward compatibility for servers running older versions of iperf3.

In addition, **simplified `iperf_send_mt()`**, since it contained some leftovers of the multi-streams support (including some variables renaming).

Tested with and without nanosleep support.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1993,10 +1993,10 @@ iperf_check_total_rate(struct iperf_test *test, iperf_size_t last_interval_bytes
 int
 iperf_send_mt(struct iperf_stream *sp)
 {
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the PR! I like the idea of this change (haven't had a chance to really study it though).

bmah888 (approved):
Tested this on all our supported platforms, pacing (still) works for both forward and reverse direction tests. I will merge this shortly. Thanks for making iperf3 smaller and (slightly) less complicated!

Issue #1770, opened 2024-09-19
iperf3 -O (--omit) to get more than 60 seconds omit time


tblaszcz:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the Discussions section for this project at
https://github.com/esnet/iperf/discussions
or to the iperf3 mailing list at iperf-dev@googlegroups.com.
A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3: iperf3.12

* Hardware: NA

* Operating system (and distribution, if any): Linux

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate. Note that submissions of bug fixes, new
features, etc. should be done as a pull request at
https://github.com/esnet/iperf/pulls_

# Bug Report

* Expected Behavior

* Actual Behavior

* Steps to Reproduce

* Possible Solution

# Enhancement Request
For many testing purposes the omit option should allow to omit more than 60 seconds

* Current behavior
* Omit allows to put max 60 seconds 

* Desired behavior
* Omit allows to put any int value in seconds

* Implementation notes

tblaszcz:
Hello, Thank you for quick answer. The reason is to get tput only for some specific period of time or to get tput only from the end of iperf connection. Not related to slow start but some external features activity impacting the connection throughput. Due to some internal conditions the feature cannot be activated in first 60 seconds of the iperf. Because of that I need to omit more time to be sure that the tput is calculated for the time when feature is active. I can imagine more cases like this. Please let me know if you can consider extending the omit time. Thanks a lot.

davidBar-On:
Submitted PR #1774 which extends the `--omit` limit to the same limit as `--time` (86,400 seconds).  I don't know if and when this PR will be merged into the main line, but you may use its code to build a private version.

PR #1774, opened 2024-09-20
No limit to omit time period

davidBar-On:
master

* Issues fixed (if any): #1770

Change the `--omit` time period limit from 60 to the same limit as `--time` (86,400 seconds = one day).
My assumption is that the 60 seconds was just to be more user friendly, assuming that the omit period is only to omit possible performance fluctuations at the first few seconds of the tests.  Per #1770 there are other use cases that requires larger omit time period.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1493,7 +1493,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                 break;
             case 'O':
                 test->omit = atoi(optarg);
-                if (test->omit < 0 || test->omit > 60) {
+                if (test->omit < 0 || test->omit > MAX_TIME) {
                     i_errno = IEOMIT;
                     return -1;
                 }
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1493,7 +1493,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                 break;
             case 'O':
                 test->omit = atoi(optarg);
-                if (test->omit < 0 || test->omit > MAX_TIME) {
+                if (test->omit < 0 || test->omit > MAX_OMIT_TIME) {
                     i_errno = IEOMIT;
                     return -1;
                 }
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -447,6 +447,7 @@ struct iperf_test
 #define MIN_INTERVAL 0.1
 #define MAX_INTERVAL 60.0
 #define MAX_TIME 86400
+#define MAX_OMIT_TIME 600
 #define MAX_BURST 1000
 #define MAX_MSS (9 * 1024)
 #define MAX_STREAMS 128
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -213,7 +213,7 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "this OS does not support sendfile");
             break;
         case IEOMIT:
-            snprintf(errstr, len, "bogus value for --omit");
+            snprintf(errstr, len, "bogus value for --omit (maximum = %d seconds)", MAX_OMIT_TIME);
             break;
         case IEUNIMP:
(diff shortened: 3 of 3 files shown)


bmah888:
Thanks for the PR! Some of these hard limits exist to put sane boundaries on some of the test parameters, the idea being that if a user really needed to do say very long tests they should be at least knowlegable enough to modify the code and recompile it.

In this case, there (probably?) isn't any particular resource constraint caused by having longer `--omit` times, so it probably makes sense to change the bound as in this PR.

src/iperf_api.c
```diff
@@ -1493,7 +1493,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                 break;
             case 'O':
                 test->omit = atoi(optarg);
-                if (test->omit < 0 || test->omit > 60) {
+                if (test->omit < 0 || test->omit > MAX_TIME) {
```
bmah888, line 1496: I think I've changed my mind on this. I like the ideas of getting rid of the hard-coded `60` and allowing something longer. I'm not sure that allowing up to a day of idle time, followed by up to a day of actual test, really has a practical use, and it ties up the server for longer than might be desired.

What about having another constant, say `MAX_OMIT_TIME`, defined to something like 600 (that's 10 minutes, which is a pretty long time), do the check against that?

Again, if anyone feels a need for a longer omit time, they hopefully have the technical skill to check out the source code, make the change, and compile it.

What do you think?

bmah888 (commented):
Thanks again for the PR! I sorta changed my mind about how I feel about this, I'd like to do this in a slightly different way.

src/iperf_api.c
davidBar-On: The reason I used the `MAX_TIME` is because one of the use-cases for longer omit time mentioned in #1770 is "**get output only from the end of iperf connection**".  If this use-case is regarded as useful, then the maximum omit time should be the same as the maximum test time.

Another option for the above use-case, is to allow **negative** omit time, to mean that the omit time is for the test time **except** for the positive value of the "omit" time.  E.g. `-t 100 -O -20`  means omitting the first 80 seconds of the test.  In this case smaller (e.g. 600) absolute omit time is probably sufficient.

If the above use-case doesn't seem useful enough, or the negative value option is taken, then I agree that (absolute) 600 seconds should be sufficient.  In any case, I assume that defining `MAX_OMIT_TIME` can be useful for easier modifications by anyone that wants to change the limit.

bmah888: My understanding of how `--omit` works (which admittedly I didn't write myself) is a little different. If the user specifies `--omit`, iperf3 runs a "pre-test" for that number of seconds before it starts the actual test of however many seconds, blocks, or bytes. So, if I start up a client with `--omit 5 --time 20`, the client sends data for 5 seconds (which presumably is enough to overcome startup transients) and then does a 20-second test. The server knows to throw away those first 5 seconds and only report on the 20 seconds.

If someone wanted to only get data "from the end", they might do something like "--omit 15 --time 5`, which does a 15-second pre-test followed by 5 seconds of actual test. You could also think of this as saving only the last 5 seconds of a 20 second test.

My interpretation of #1770 is just that they needed an omit time longer than 60 seconds to set up some stuff for whatever they were doing. 60 seconds is a extremely long time by networking standards (in terms of RTT, buffer utilization, etc.) and I can't remember us at ESnet needing an omit time that long for anything we were doing. But I figure that supporting a slightly longer omit time can't hurt.

I understand that the semantics of `--omit` can be a little confusing. If I were designing the feature I might have done this a little differently.

davidBar-On: You are right.  I knew that the omit time is a pre-test period but somehow forgot that ...

OK. I changed the code as requested - limit omit time to `MAX_OMIT_TIME` which is set to 600.

bmah888: Thanks! Hoping to finish this off later this week.

Issue #1768, opened 2024-09-18
Lower throughtput numbers reported with v3.16


pb8o:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the Discussions section for this project at
https://github.com/esnet/iperf/discussions
or to the iperf3 mailing list at iperf-dev@googlegroups.com.
A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3: iperf3 3.16 and above

* Hardware: HP ProBook laptop

* Operating system (and distribution, if any): Ubuntu 22.04

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

# Bug Report

* Expected Behavior

Consistency between versions 3.15 and below and 3.16 and above

* Actual Behavior

iperf3 3.16 reports much smaller numbers 

* Steps to Reproduce

```
# ip netns exec netns-master-0 ./iperf3-amd64-3.15 -c 192.168.0.2 -n $(( 2**20 * 100 )) -N
Connecting to host 192.168.0.2, port 5201
[  5] local 192.168.0.1 port 40766 connected to 192.168.0.2 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-0.03   sec   100 MBytes  30.0 Gbits/sec  135   1.07 MBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-0.03   sec   100 MBytes  30.0 Gbits/sec  135             sender
[  5]   0.00-0.03   sec  97.6 MBytes  29.0 Gbits/sec                  receiver

iperf Done.
```

```
# ip netns exec netns-master-0 ./iperf3-amd64-3.16 -c 192.168.0.2 -n $(( 2**20 * 100 )) -N
Connecting to host 192.168.0.2, port 5201
[  5] local 192.168.0.1 port 40798 connected to 192.168.0.2 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[...]

PR #1775, opened 2024-09-22
End test immediately when all number of blocks/bytes sent/received

davidBar-On:
master

* Issues fixed (if any): #1768

Currently, after the change to multi-thread, when limiting the test by number of bytes (`-n`) or number of blocks (`-k`), the test ends only when the current interval (`-i`) ends and not when all the data was sent/received.  This is because the `select()` no longer monitor the steams data.  Therefore the throughput statistics are not properly calculated.  I.e. for 1 second interval, if sending was done in 2.5 seconds, the statistics will be as if sending was done in 3 seconds.  This is much worse if the interval period is larger.

This suggested fix is using the [self-pipe trick](https://cr.yp.to/docs/selfpipe.html), based on [this example](https://www.michaelkerrisk.com/tlpi/code/online/dist/altio/self_pipe.c.html).  When `-n` or `-k` are set, a self-pipe is added to the Client's select.  When the limit is reached, one byte is written to the pipe to wake-up the `select()` in the main thread.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -288,6 +288,7 @@ enum debug_level {
 struct iperf_test
 {
     pthread_mutex_t print_mutex;
+    pthread_mutex_t pipe_mutex;
 
     char      role;                             /* 'c' lient or 's' erver */
     enum iperf_mode mode;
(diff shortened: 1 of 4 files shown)


bmah888:
Thanks for the PR! Admittedlly this might take alittle time to study what this change does.

PR #1777, opened 2024-09-26
fix typo in iperf_util.c

iZarrios:
fix typo in `make_cookie` documentation.

### DIFF
--- a/src/iperf_util.c
+++ b/src/iperf_util.c
@@ -105,7 +105,7 @@ void fill_with_repeating_pattern(void *out, size_t outsize)
  * Generate and return a cookie string
  *
  * Iperf uses this function to create test "cookies" which
- * server as unique test identifiers. These cookies are also
+ * serve as unique test identifiers. These cookies are also
  * used for the authentication of stream connections.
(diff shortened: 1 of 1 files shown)


src/iperf_util.c
```diff
@@ -105,7 +105,7 @@ void fill_with_repeating_pattern(void *out, size_t outsize)
  * Generate and return a cookie string
  *
  * Iperf uses this function to create test "cookies" which
- * server as unique test identifiers. These cookies are also
+ * serves as unique test identifiers. These cookies are also
```
bmah888, line 108: If we're going to fix that word, it should be just "serve" because the word "cookies" is plural.

bmah888 (changes requested):
Thanks for the pull request! Definitely we want good documentation that isn't confusing. As long as we're fixing this word it'd be good to get the subject/verb agreement correct.

iZarrios:
Fixed it.

bmah888 (approved):

bmah888:
Thanks for the revision, merged!

Issue #1738, opened 2024-07-23
Support affinity/pinning of parallel flows to different CPUs


marcosfsch:
# Enhancement Request

* Current behavior
iperf3 supports setting affinity to a single CPU, which made sense when it was single threaded

* Desired behavior
Support setting a rage/list of CPUs to affinity of multiple flows: Ex: "iperf3 -c localhost -P 4 -A 1-4,1-4"

* Implementation notes
There are at least two way this can be implementated and I'd suggest the second approach:
1) Replicate numactl behavior, where "numactl -C 1-4" would bind each flow to the CPU range "1-4" and Linux would be responsible for the scheduling, allowing for dinamically rebalancing the flows but generating noise and perfromance drops.
2) Statically schedule each flow to a different CPU, i.e. in a round-robin fashion, so that when you have a maching set of parallel threads and CPUs you'd have 1 flow per CPU, minimizing reschedules during the transfer and optimizing performance

bmah888:
You raise a good point in that the current `-A` behavior doesn't work very well in a multi-threaded iperf3. So far our standard practice within ESnet is just to do `numactl`, as you suggested in the first approach. How bad are the downsides you mentioned...have you or others observed these problems? (I don't think we have, but multithreaded application performance analysis is not my forte.)

In considering these different implementations, we also want to keep in mind other OS platforms that support `-A`, such as FreeBSD (a supported platform), and Windows (while not officially supported, an environment that I'd like to avoid gratuitously breaking).

manedurphy:
I happened to be exploring an idea related to the `affinity` feature and stumbled upon this existing request. What I was looking into was adding the ability to pin specific streams/threads to target CPU cores. If I run an `iperf3` client on an 4-core machine, then I should be able to pin each stream/thread to each core of that machine: `iperf3 --client my.iperf3-server.com --parallel 4 --stream-affinity 0,1,2,3`. Implemented this with a comma-separated list in [iperf/#1939](https://github.com/esnet/iperf/pull/1939), but it's just a proof-of-concept, not the final design. I raised the PR as reference material for expanding the conversation on this enhancement request.

[...]

PR #1778, opened 2024-09-26
Add mutli-CPU Affinity option support for Linux (not for FreeBSD)

davidBar-On:
master

* Issues fixed (if any): #1738

Suggested support for multi-CPU Affinity option on Linux.  The PR is **only for Linux**, since I don't have FreeBSD machine.

The **format of a CPUs list** is using `/` as a separator ("n1/n1/...") since `,` is already used to separate between the client and server CPUs lists.

The fix is backward compatible, passing the CPUs list to the server in a new "server_cpu_set" JSON item and passing the first CPU defined in the server's list in the legacy "server_affinity" item.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -305,7 +305,12 @@ struct iperf_test
     int       omit;                             /* duration of omit period (-O flag) */
     int       duration;                         /* total duration of test (-t flag) */
     char     *diskfile_name;			/* -F option */
-    int       affinity, server_affinity;	/* -A option */
+#if defined(HAVE_SCHED_SETAFFINITY)
+    cpu_set_t cpu_set, server_cpu_set;          /* -A option */
+#else
+    int       affinity;	                        /* -A option */
+#endif /* HAVE_SCHED_SETAFFINITY */
+    int       server_affinity;	                /* -A option */
 #if defined(HAVE_CPUSET_SETAFFINITY)
     cpuset_t cpumask;
 #endif /* HAVE_CPUSET_SETAFFINITY */
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1166,6 +1166,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
     struct xbind_entry *xbe;
     double farg;
     int rcv_timeout_in = 0;
+    int i;
 
     blksize = 0;
     server_flag = client_flag = rate_flag = duration_flag = rcv_timeout_flag = snd_timeout_flag =0;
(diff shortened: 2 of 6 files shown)


junousi:
Hello! I'm trying to test this affinity patch on two Ubuntu servers with 100G NICs, and would like to verify the test set up is at least somewhat reasonable. Please instruct if PR is not the right place for this kind of discussion, e.g. if some user forum would be more suitable.

Current NIC bindings:

```
client:~$ numactl -H|grep cpus
node 0 cpus: 0 2 4 6 8 10 12 14
node 1 cpus: 1 3 5 7 9 11 13 15
client:~$ cat /sys/class/net/NIC/device/numa_node 
0
```

```
server:~$ numactl -H|grep cpus|column -t
node  0  cpus:  0  4  8   12  16  20  24  28
node  1  cpus:  1  5  9   13  17  21  25  29
node  2  cpus:  2  6  10  14  18  22  26  30
node  3  cpus:  3  7  11  15  19  23  27  31
server:~$ cat /sys/class/net/NIC/device/numa_node 
1
```

Server compiled from this branch and fired up:

```
server:~$ ./iperf3 --server --interval 0 -p 1234
```

Client compiled from this branch and ran with affinity sets that should match the NICs:

```
client:~$ ./iperf3 --parallel 8 --interval 0 --client 10.10.10.1 --port 1234 -A "0/2/4/6/8/10/12/14,1/5/9/13/17/21/25/29"
...
[SUM]   0.00-10.03  sec  13.6 GBytes  11.7 Gbits/sec    0             sender
[SUM]   0.00-10.03  sec  13.6 GBytes  11.6 Gbits/sec                  receiver
```

It seems we have still tweaking to do either on the iperf3 parameters, or the server configs, or both.

Could you please advise are there any best practises for testing this branch regarding:

- Recommended kernel versions
- Prepending the `iperf3` command with e.g. `numactl --cpunodebind=netdev:NIC` (not needed?)
- Irqbalance, RSS & friends
- Combined rings (currently `16` for both NICs per `ethtool -l`)
- Eventually parallelizing additional iperf3 instances over the NUMA nodes which are *not* bound to the NIC (don't do it?)
- Any other pointers or hints for testing?

davidBar-On:
Hello @junousi, it is usually better to handle such discussions in iperf3 "discussion" (see "Discussions" at the top line).  However, it this case I think that it is better to handle this discussion in other forums, probably a forum about NUMA (I am not familiar with this topic).

One suggestion about iperf3 is to make sure that the different 8 threads per stream are running on different CPU.  E.g. this can be done by running the iperf3 test for long enough time (using the `-t` option) and in parallel running `htop` command to see the load of each CPU.

PR #1787, opened 2024-10-18
No select() when reading stream data

davidBar-On:
master

* Issues fixed (if any): #1707

Suggested solution to make iperf3 UDP performance comparable to nuttcp (and probably other tools).  The change is to **not** use `select()` when reading stream data.  This by replacing in these case the calls to `Nread()` to calls to `Nread_no_select()` (new function).

The assumption is that the `select()` in `Nread()` was added to make sure that the **main thread** will not get stuck (when reading control/JSON data), and that it is not needed for the data streams.  E.g., the server will not get stuck even if one of its data streams threads is stuck, as the that thread will be canceled at the end of the test.

Although the reported issue was only about UDP, the change was done also for TCP, SCTP, as it seems reasonable to do.  If that may cause issues, the change to these protocols may be removed.

### DIFF
--- a/src/iperf_sctp.c
+++ b/src/iperf_sctp.c
@@ -56,7 +56,7 @@ iperf_sctp_recv(struct iperf_stream *sp)
 #if defined(HAVE_SCTP_H)
     int r;
 
-    r = Nread(sp->socket, sp->buffer, sp->settings->blksize, Psctp);
+    r = Nread_no_select(sp->socket, sp->buffer, sp->settings->blksize, Psctp);
     if (r < 0)
         return r;
 
--- a/src/iperf_tcp.c
+++ b/src/iperf_tcp.c
@@ -58,7 +58,7 @@ iperf_tcp_recv(struct iperf_stream *sp)
 {
     int r;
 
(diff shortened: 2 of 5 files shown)


bmah888:
I did some testing with our dev hosts (two older servers connected by a 100GE path, one router hop within the same rack) and saw much less packet loss and greater "goodput" with single-stream UDP performance. (I briefly tested some other scenarios as well, but this addressed the main reason for this code change.)

I was trying to see if there's an easy way to factor out the common code between `Nread()` and `Nread_no_select()` but haven't found one yet. (This is not a blocker for merging this change.)

Hoping to get this merged in the next few days after a little more testing.

davidBar-On:
> ... easy way to factor out the common code between `Nread()` and `Nread_no_select()` ...

The way I considered was to add a parameter to `Nread()` for whether to do the `select()`, but I didn't do that as it would require to change all the calls to `Nread()`.

bmah888 (approved):
Looks good...testing showed a significant improvement on our test hosts (reflected in less packet loss and higher goodput for UDP tests).

Issue #1019, opened 2020-06-13
Allow different control connection port number than data connection


t-lin:
# Context
UDP tests require a TCP control connection to be set up beforehand. This connection currently uses the same port number as the UDP port used for the test / data connection.

Consider an environment where the network between the client & server are comprised of one or more third-party network services (e.g. VPN, proxy, firewall, etc.) with very strict security policies. Only certain ports are opened, and no port number is opened for both TCP and UDP. It's very difficult to request a new port number be opened (when requesting a port, they return a randomized port number and only for one protocol).

I'd be willing to help if someone can point me to the files/functions to look at.

# Enhancement Request
Allow users to specify the port number used for the control connection, different than the port number specified by `-p` which is used for the actual test data.

* Current behavior
Control connection defaults to the same port number used for the test / data connection.

* Desired behavior
Control connection uses a different port number than the test / data connection.

* Implementation notes
Introduce a new flag to specify the control port? Ideally this flag should be applicable for both when iperf3 runs in server mode or client mode.

deliciouslytyped:
We ran into a case where this would have been useful.

davidBar-On:
Submitter PR #1790 that supports a UDP data streams port number that is different than the TCP control port number.

onetimesso:
I am very interested in this capability and would like to help with testing.  Where can I get the branch?

davidBar-On:
> I am very interested in this capability and would like to help with testing. Where can I get the branch?

You can get and build the PR #1790 branch code by (I have now rebased it to the latest iperf3 master branch):
```
git clone -b issue-1019-udp-data-port https://github.com/davidBar-On/iperf
cd iperf
./bootstrap
./configure
make
```

The built iperf3 is `./src/iperf3`.  You can make this version your computer iperf3 by `sudo make install`.

onetimesso:
Thanks for the code and tips on how to build it.  I was able to build the code successfully on the following three platforms, all running Debian GNU/Linux 12 (bookworm).

RaspberryPi 4B 8 GB RAM
Linux rpi4-bookworm 6.6.62+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.62-1+rpt1 (2024-11-25) aarch64 GNU/Linux

HP ProBook 640 G4
Linux HP-640G4 6.1.0-18-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.76-1 (2024-02-01) x86_64 GNU/Linux

[...]

PR #1790, opened 2024-10-23
Support for UDP Data-port that is different than the Control-port

davidBar-On:
master

* Issues fixed (if any): #1019

Add support for a UDP Data-port that is different than the control port, by adding an optional second value to `-p`.  E.g. `-u -p /5210` (or `-u -p 5201/5210`) will use port 5201 for the control channel and port 5210 for the data streams.

This is to allow running UDP tests in systems where the port number used for TCP is not open for UDP, so the UDP data streams should use a different port


bmah888:
This sounds useful, thanks for the PR!

Issue #1711, opened 2024-06-05
How to use the iperf library to obtain bandwidth information in real-time at intervals?


QingshuangZhu:
I compiled iperf into a library and then used it in my code. I want to obtain the bandwidth information once every interval. 
What should I do? Thank you.

davidBar-On:
As far as I am aware there is no such capability in iperf3.  One `iperf_run_client(()` is called, there is practically no interaction with the library, except for the possibility to view the output (stdout or the `iperf_set_test_logfile(()` file).

QingshuangZhu:
Thank you very much for your answer. I took a brief look at the iperf3 source code and it indeed does not support this feature.

Currently, I am doing it like this: first, I set the JSON streaming output with `iperf_set_test_json_output(test, 1)` and `iperf_set_test_json_stream(test, 1)`; And I set `logfile` and `outfile` through `iperf_set_test_logfile("./iperf3.log")`; then I execute `iperf_run_client()` in one thread, and another thread waits for 2 seconds before reading the JSON data from `iperf3.log` and parsing out the bandwidth information. 

I look forward to iperf3 supporting this feature through a callback function. Thank you again.

davidBar-On:
Thanks for the detailed explanation.  Not sure how easy/difficult it is to implement that, but Just to make sure.  Are you looking for a callback at the end of the test, so there will be no need to wait the 2 seconds, something like the function bellow that will set JSON file output and will define the callback at the end of the test, something like:
```c
void end_of_test_callback(struct iperf_test *test);
int iperf_set_test_json_end_of_test_callback(
    struct iperf_test,
    (*json_callback)(struct iperf_test *),
    char *logfile_name
);
```

Or, something else?

QingshuangZhu:
Thank you very much for your reply. I hope to obtain bandwidth information in real-time at each time interval, rather than at the end of the test. The two-second wait is to ensure that the test data is written into the logfile file. I think an **interval event callback function** should be added, but I feel it might not be easy to implement (I am not familiar with the iperf3 source code).

davidBar-On:
Submitted PR #1798 that supports a callback function that gets the output JSON strings, instead of writing them to the output file.   The callback function is set using `iperf_set_test_json_callback()`.

bmah888:
I'm reviewing PR #1798. It basically works as designed, although for it to fulfill the original request of getting interval statistics, the client program needs to have set the equivalent of `--json-stream`. From the API, it's basically something like:
```
[...]

PR #1798, opened 2024-11-17
Add callback function to get JSON Output strings

davidBar-On:
master

* Issues fixed (if any): #1711

Add `iperf_set_test_json_callback()` that allows a user app to set a callback function to get the JSON output strings, instead of printing them to the output file.

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -343,8 +343,8 @@ struct iperf_test
     int	      verbose;                          /* -V option - verbose mode */
     int	      json_output;                      /* -J option - JSON output */
     int	      json_stream;                      /* --json-stream */
-    void      (*json_callback) (char *);        /* used by user apps to receive the JSON strings,
-                                                   instead of writing them to the output file */
+    void      (*json_callback) (struct iperf_test *, char *); /* allow user apps to receive the
+                                                JSON strings,instead of writing them to the output file */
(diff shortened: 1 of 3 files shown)


swlars:
Thanks for the pull request! This looks useful!

bmah888 (commented):
Everything seems to work correctly. I have one question about the intended callback function...all of the other callbacks take a `test` argument. For those callbacks (such as `on_new_stream`, `on_test_start`, etc.), the `test` pointer gives all the information needed by the callback. The `json_callback` is different, in that it looks like it only takes a string containing a JSON rendered object. Do you think that having the `test` structure as well might be needed to give some context for working with the JSON? So then callback would take both a `(iperf_test *)` and a `(char *)`?

bmah888:
Also, the code changes in #1805 that deal with existing callback hooks look relevant here as well.

davidBar-On:
> ... I'm just wondering if having a pointer to the test structure might be useful, say to refer to test parameters, etc.

Completely agree.  Added the `test` parameter and changed per #1805.  Now the callback setting function definition is:
`void iperf_set_test_json_callback(struct iperf_test *ipt, void (*callback)(struct iperf_test *, char *));`

bmah888 (approved):
This looks good, merging!

Issue #596, opened 2017-06-07
bytes/blockcount with reverse mode


vgramkris:
iperf 3.1.2
CentOS 6.4 x86_64

While testing with iperf3 blockcount (-k) option (similar behavior with bytes (-n) option), I see the following observations
1) Normal mode
Client:  iperf3 -p 25000 -c <server_ip> -u -l 1370 -b 3m -k 2000
Server: iperf3 -p 25000 -s -i 1
Client sends ~2000 packets to the server irrespective of the packets lost in the network. Hence, even if server received only 1500 packets due to nw losses, the client stopped sending after transmitting 2000 packets.

2) Reverse mode
Client:  iperf3 -p 25000 -c <server_ip> -u -l 1370 -b 3m -k 2000 -R
Server: iperf3 -p 25000 -s -i 1
Server sends packets till the client receives ~2000 packets. Hence, if this is a lossy nw such that 25% packets are lost in transmission, then server effectively sends 2500 packets and client receives 2000 packets and only then the test ends.

Please clarify if this is the intended implementation.

bmah888:
The problem with this is similar to #594, where the server knows when the test is done but doesn't have an easy way to tell the client that (and the client is the side that controls stopping the test...this is regardless of whether `--reverse` is used or not).  Needs some thought to see whether there's a straightforward solution to this problem or not.

vgramkris:
Thanks for the clarifications

mruprich:
Hi,
I observed a little bit different behavior with iperf 3.9 on Fedora but it is related to the combination of -k/-n and -R. It seems that if I use -k or -n together with -R, the test actually runs indefinitely. This is how I used it:
# iperf3 --server --daemon --port 5201
# iperf3 -c 127.0.0.1 -p 5201 -i2 -VR --forceflush --debug -l 1M -N --parallel 3 --dscp 0 --fq-rate=0 [-n|-k] 10 -T'dope'

After this, no matter whether I use -k or -n, the test never ends. (Well I don't want to say never but it ran for an hour so I killed it). Any idea why this happens? Is this issue related to the issue here?

Thanks for any input on this.
Michal

davidBar-On:
In my local computer I see the same problem using iperf3 version 3.7, but it doesn't exist when I am using 3.10.

davidBar-On:
[...]

PR #1800, opened 2024-11-21
Change help to make current -n/-k behavior a Feature

davidBar-On:
master

* Issues fixed (if any): workaround for #596, #1768

Currently, there are two issues with the `--bytes` and `--blockcount` options:
- Test ends only at the end of the interval when at least the specified number of bytes/blocks were sent ( Issue #1768 - issue started with the multi-thread versions, and PR #1775 is a suggested fix).
- The bytes/blocks count is done only by the client, even in reverse mode (Issue #596).

The purpose of this PR to **make the current behavior a feature** by documenting it in the help message, assuming that these issues are **not** going to be fixed.


swlars:
Thanks for the pull request! This looks like a pretty straightforward request, so we should be getting to it soon.

bmah888 (approved):
This looks good, thanks! Merging shortly.

PR #1801, opened 2024-11-21
Do not try to cancel NULL thread (causing Segmentation fault)

davidBar-On:

* Issues fixed (if any): none

Suggested Bug fix: thread id 0 is not regarded as illegal id (at least in Ubuntu Linux), and `pthread_cancel()` **crashes** (Segmentation fault) when trying to cancel a thread with id 0.  That happens if streams creation was successful, but there was a failure before all threads are created.  The problem happened to me at work and I can reproduce it in my private PC.

The suggested fix is to add an indication `thread_created` per stream to indicate if the thread was created for the stream, and try to cancel the stream's thread **only if it was created**.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4498,6 +4498,7 @@ iperf_new_stream(struct iperf_test *test, int s, int sender)
 
     memset(sp, 0, sizeof(struct iperf_stream));
 
+    sp->thread_created = 0;
     sp->sender = sender;
     sp->test = test;
     sp->settings = test->settings;
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4498,7 +4498,6 @@ iperf_new_stream(struct iperf_test *test, int s, int sender)
 
     memset(sp, 0, sizeof(struct iperf_stream));
 
-    sp->thread_created = 0;
     sp->sender = sender;
     sp->test = test;
     sp->settings = test->settings;
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -193,6 +193,7 @@ struct iperf_stream
     struct iperf_test* test;
 
     pthread_t thr;
+    int thread_created;
     int       done;
 
     /* configurable members */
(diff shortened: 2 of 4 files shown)


src/iperf_api.c
```diff
@@ -4498,6 +4498,7 @@ iperf_new_stream(struct iperf_test *test, int s, int sender)
 
     memset(sp, 0, sizeof(struct iperf_stream));
 
+    sp->thread_created = 0;
```
bmah888, line 4501: Technically don't need this line because of the `memset()` above. It's not wrong to have it here though.

bmah888 (commented):
Thanks for the PR! This looks like a good change (set of changes). Just one comment, mostly FYI and not something that needs to be "fixed". Just need to do some testing on this...

src/iperf_api.c
davidBar-On: Line removed.

bmah888:
@davidBar-On I have just one more question...the code looks good according to my reading of it...I haven't been able to trigger the original bug to compare the original and fixed behavior. Any suggestion about how I can test this? Thanks!

davidBar-On:
@bmah888, the way I tested it was by causing the client to fail after the streams creation, just before the threads creation.  E.g., in `iperf_client_api.c`: adding `goto cleanup_and_fail;` just before:
```c
                SLIST_FOREACH(sp, &test->streams, streams) {
                    if (pthread_create(&(sp->thr), &attr, &iperf_client_worker_run, sp) != 0) {
                        ...         
```

bmah888:
> @bmah888, the way I tested it was by causing the client to fail after the streams creation, just before the threads creation. E.g., in `iperf_client_api.c`: adding `goto cleanup_and_fail;` just before:
> 
> ```c
>                 SLIST_FOREACH(sp, &test->streams, streams) {
>                     if (pthread_create(&(sp->thr), &attr, &iperf_client_worker_run, sp) != 0) {
>                         ...         
> ```

Thanks! I'm still not able to reproduce the segmentation fault, but with the help of `--debug` I can see that the code in this PR is preventing iperf3 from trying to cancel non-existent threads, which is probably close enough.

bmah888 (approved):

PR #1812, opened 2024-12-14
Reset the congestion protocol on the server between tests

hwti:
  master

* Issues fixed (if any):

Reset the congestion protocol between tests. Otherwise, the server keeps the last congestion protocol asked by a client (-C option), applying it to the following clients not asking for a specific protocol.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -3262,6 +3262,9 @@ iperf_reset_test(struct iperf_test *test)
 
     SLIST_INIT(&test->streams);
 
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the bug report and PR! I've been able to reproduce the problem.

bmah888 (approved):
Looks good and it solves the problem. Thanks for the PR!

bmah888:
Tested this by using the `-verbose` option to both iperf3 client and server, with various `-C` arguments.

PR #1814, opened 2024-12-20
Remove (by design) memory-leak by from the Intervals Results linked-list

### LINKED ISSUE #1806: Significant jitter increase from 3.6 to 3.17
# Context

Performing Network stress testing on 24 ports simultaneously with iperf3.6 (what was available with `sudo apt install iperf3`) yielded consistent jitter rates of approximately 50 µs and packet drops of under 0.0006 %. 
I upgraded iperf on the test equipment to 3.17 due to https://github.com/esnet/iperf/pull/1260. Since upgrading the rate of packet drops and jitter is significantly higher at 30% and 1 ms.

`iperf3 -s -A {1:13} -p 5112 -B 192.168.1.10 -i 0.1 --one-off --forceflush`
`iperf3 --client 192.168.1.10 -A {1:13} -u -i 0.1 -b 300m --pacing-timer 1 -O 10 --R --forceflush`

* Version of iperf3:

Version 3.6 and `iperf 3.17.1 (cJSON 1.7.15)`

* Hardware:

Same hardware between the two tests, so unlikely to be a factor

`4.19.0-26-amd64 #1 SMP Debian 4.19.304-1 (2024-01-09) x86_64 GNU/Linux`

# Bug Report

* Expected Behavior

No change in performance between the two versions

* Actual Behavior

Significant performance hit

1. Performance on 3.6  

 image

1.  Performance on 3.17 

image

* Steps to Reproduce

Was able to move between 3.6 and 3.17 and the same behaviour was demonstrated.

davidBar-On:
master

* Issues fixed (if any): may fix #1806

Intervals Results history is saved in a linked list, but is seems that this history is not used (history is used only as JSON or printed output).  Therefore, keeping this linked list cause unnecessary memory usage and practically this is a **memory leak**.  As each entry i the list is about 400 bytes, with 0.1 sec interval, about 15MB are allocated per one stream per hour, which is about 350MB for one day test per stream.

I suspect that #1806 [problem after running for 3 hours ](https://github.com/esnet/iperf/issues/1806#issuecomment-2533562802)with 24 streams was caused by this memory leak (although this is not confirmed yet).

**This PR suggest the removal of the Interval Results history**, by removing the last entry when allocating a new one.  Practically making 1 the maximum list length.

The suggested change is not optimal, as it removes the old entry and then insert the new one.  A better approach may be to replace the last entry, just reuse the last entry, or even don't use a linked list at all.  However, since the operation is relatively not frequent, the chosen approach seems to be good enough with lower risk.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2844,6 +2844,14 @@ add_to_interval_list(struct iperf_stream_result * rp, struct iperf_interval_resu
 {
     struct iperf_interval_results *irp;
 
+    /* Only the last interval result is needed, so removing last old entry to reduce memory consupmtion */
+    if (!TAILQ_EMPTY(&rp->interval_results) &&
+        (irp = TAILQ_LAST(&rp->interval_results, irlisthead)) != NULL
+    ) {
+        TAILQ_REMOVE(&rp->interval_results, irp, irlistentries);
+	free(irp);
+    }
+
     irp = (struct iperf_interval_results *) malloc(sizeof(struct iperf_interval_results));
     memcpy(irp, new, sizeof(struct iperf_interval_results));
     TAILQ_INSERT_TAIL(&rp->interval_results, irp, irlistentries);


swlars:
Thanks for the pull request! This looks like a good change. We'll test it and look at it in more detail soon.

swlars (approved):
Looks good! Thanks!

Issue #1771, opened 2024-09-19
-M has no effect on packet size during data packets initiated from iperf3 client


avijeet-pattanayak:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the Discussions section for this project at
https://github.com/esnet/iperf/discussions
or to the iperf3 mailing list at iperf-dev@googlegroups.com.
A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context
In iperf3 application though the MSS is negotiated based on -M option during TCP SYN, but the data packets generated by iperf3 client are much bigger in size. Looks like it is not honoring the -M option to set MSS 

* Version of iperf3:
iperf3 version 3.5

* Hardware:
General purpose server
* Operating system (and distribution, if any):
Linux 
_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate. Note that submissions of bug fixes, new
features, etc. should be done as a pull request at
https://github.com/esnet/iperf/pulls_

# Bug Report

* Expected Behavior

* Actual Behavior

* Steps to Reproduce

* Possible Solution

# Enhancement Request

* Current behavior

* Desired behavior

* Implementation notes

davidBar-On:
> iperf3 version 3.5

iperf3.5 is a very old version. Suggest to use a new version (although per issue #779 probably it will not help this specific issue).

> In iperf3 application though the MSS is negotiated based on -M option during TCP SYN, but the data packets generated by iperf3 client are much bigger in size. Looks like it is not honoring the -M option to set MSS

Just to make sure, are you referring to the **packet** size or **segment** size?   iperf3 packet (message) size is set by the `-l` option.

In any case, more information is needed (see #779 for example): which Linux OS and version the server and the client machines using, are you looking on the client or server size to see the segment size?  Which tool are you using for that (Wireshark, tcpdump, etc.), etc.

avijeet-pattanayak:
Can you please recommend the latest version of iperf3 which can be used in this case.
[...]

PR #1816, opened 2024-12-21
Increase MAX_MSS to 32K

nshopik:

```
```

Today hardware capable 16K MTU already, this raise MAX_MSS to 32K which can be set via -M option. 

I would increased to max allowed for ipv4 datagram size 65536 - 20 - 20, but its not possible to set value above 32767 manually as it return error. 
`setsockopt(3, SOL_TCP, TCP_MAXSEG, [32768], 4) = -1 EINVAL (Invalid argument)`

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -452,7 +452,7 @@ struct iperf_test
 #define MAX_TIME 86400
 #define MAX_OMIT_TIME 600
 #define MAX_BURST 1000
-#define MAX_MSS (9 * 1024)
+#define MAX_MSS (32 * 1024 - 1)
 #define MAX_STREAMS 128
 
 #define TIMESTAMP_FORMAT "%c "


nshopik:
32767 limit come from here ` * Never offer a window over 32767 without using window scaling. Some poor stacks do signed 16bit maths!`
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/net/tcp.h?h=v6.12.6#n68

bmah888:
Thanks for the PR, we'll take a look at it! Note there is a conceptual difference between the window size and the MSS, and there isn't necessarily a relation between the two.

bmah888:
I thought this would be a no-brainer. But while trying to test this code change I ran into a problem where trying to set the MSS has absolutely no effect. This was pointed out in #1771 (maybe other places as well). I presume that PMTU is figuring out the MSS, although testing that will require a more complicated network path than what I currently have. In any case, `-M` seems to do nothing.

Ubuntu 22.04.5 on x86_64, tip of `master` (after 3.20 tag), testing over `lo` interface, which has an MTU of 65536.

@nshopik did the `-M` option actually do anything at all in your test setup?

nshopik:
I actually don't remember exact test setup. Might as well rebuild it from scratch to actually see results.

nshopik:
Yeah quick testing confirms #1771 behavior

bmah888:
I was earlier thinking this might be platform-dependent (and maybe it is) but I'm not able to make `iperf -M` work on macOS 26, Ubuntu 22.04.05, or FreeBSD 14.3-RELEASE.

nshopik:
> Note there is a conceptual difference between the window size and the MSS, and there isn't necessarily a relation between the two.

Right, but there is actually code reference which calls that define from header

https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/net/ipv4/tcp.c?h=v6.12.62#n3831

```
	case TCP_MAXSEG:
		/* Values greater than interface MTU won't take effect. However
		 * at the point when this call is done we typically don't yet
		 * know which interface is going to be used
		 */
		if (val && (val < TCP_MIN_MSS || val > MAX_TCP_WINDOW)) {
			err = -EINVAL;
			break;
		}
		tp->rx_opt.user_mss = val;
		break;
```
So error comes this part of code and header which I mention before. Thus my limit to 32K - 1.

Issue #1808, opened 2024-12-12
iperf3: error - unable to receive parameters from client: Bad file descriptor


coinhu1995:
# Context
Got error when measuring bandwidth between macos (local) and ubuntu (AWS EC2 with public IP)

* Version of iperf3:
Client and server are the same: iperf 3.17.1+

* Hardware:
- Client: Apple macmini M1 - macos Sonoma 14.2.1
* Operating system (and distribution, if any):

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate. Note that submissions of bug fixes, new
features, etc. should be done as a pull request at
https://github.com/esnet/iperf/pulls_

# Bug Report

* Expected Behavior

* Actual Behavior

* Steps to Reproduce

* Possible Solution

# Enhancement Request

* Current behavior

* Desired behavior

* Implementation notes

davidBar-On:
> warning: JSON size of data read does not correspond to offered length

To better understand this issue, I created a [branch in my private iperf3 fork](https://github.com/davidBar-On/iperf/tree/issue-1808-json-parameters-length-mismatch) which print a debug message when this problem happens.  The message starts with "_*** TEMP DEBUG - Parameters JSON: ..._" and shows the expected Parameters JSON length and the actual length received.

Can you **rebuild** iperf3 for the **server** from this branch, run it and send the debug message output?

coinhu1995:
@davidBar-On  Here is the log
```
iperf-issue-1808-json-parameters-length-mismatch % iperf3 -c 52.77.179.242 -V -d
iperf 3.17.1+
Darwin CPP00114445A 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:53:34 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T8103 x86_64
Control connection MSS 1388
Reading new State from the Server - current state is 0-Test reset
State change: client received and changed State to 9-PARAM_EXCHANGE - Client to Server Parameters Exchange
send_parameters:
{
	"tcp":	true,
	"omit":	0,
	"time":	10,
	"num":	0,
	"blockcount":	0,
[...]

PR #1817, opened 2025-01-05
JSON_Read() warning messages enhancements

davidBar-On:
master

* Issues fixed (if any): none

Enhancement to the `JSON_Read()` `warning()` messages done while evaluating issue #1808.  I believe that they can help in other situations.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2789,12 +2789,12 @@ JSON_read(int fd, int max_size)
                             json = cJSON_Parse(str);
                         }
                         else {
-                            sprintf(msg_buf, "JSON size of data read does not correspond to offered length - expected %d bytes but received %d; errno=%d", hsize, rc, errno);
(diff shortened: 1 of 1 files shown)


bmah888 (changes requested):
Thanks for the PR! Just one quick comment so far: it'd probably be better to use `snprintf()` rather than `sprintf()` in this patch to limit the number of bytes that get written into `msg_buf`. So maybe something like `snprintf(msg_buf, sizeof(msg_buf), ...`

The chance of overflowing `msg_buf` with the current code is probably zero, but the use of `snprintf(3)` is generally discouraged and I think some compilers will warn about its usage.

bmah888 (approved):

bmah888:
Thanks again! Tested this with normal iperf3 operation, as well as some scripts that simulate abnormal operation. Going to do a squash and merge shortly.

Issue #1820, opened 2025-01-14
Iperf3.18 -  packet count for burst mode seems broken when sending UDP traffic (on Windows)


fitchtravis:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the Discussions section for this project at
https://github.com/esnet/iperf/discussions
or to the iperf3 mailing list at iperf-dev@googlegroups.com.
A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3:

Using a build from - https://github.com/ar51an/iperf3-win-builds
```
iperf 3.18 (cJSON 1.7.15)
CYGWIN_NT-10.0-20348 MELPRTG01PRD 3.5.4-1.x86_64 2024-08-25 16:52 UTC x86_64
Optional features available: CPU affinity setting, support IPv4 don't fragment, POSIX threads
```
* Hardware:
System Manufacturer:       VMware, Inc.
System Model:              VMware Virtual Platform
Processor(s):       1 Processor(s) Installed.
                            [01]: Intel64 Family 6 Model 85 Stepping 4 GenuineIntel ~2095 Mhz
Total Physical Memory:     32,767 MB

* Operating system (and distribution, if any):

OS Name:                   Microsoft Windows Server 2022 Datacenter
OS Version:                10.0.20348 N/A Build 20348

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate. Note that submissions of bug fixes, new
features, etc. should be done as a pull request at
https://github.com/esnet/iperf/pulls_

# Bug Report

* Expected Behavior

When using -b 100M/100, we expect to see iperf3 sending at 10Mbps with a burst of 100 packets.

Identically issue to - https://github.com/esnet/iperf/issues/1192

Without the bust mode option, tests run normal. 

With burst mode iperf seems to be ignoring the bandwidth throttle limit

* Actual Behavior

iPerf3 is sending as much data as possible.

```
[...]

PR #1821, opened 2025-01-16
Fix issue that bitrate is ignored when burst is set (with Nanosleep)

davidBar-On:
master

* Issues fixed (if any): #1820

Fix a bug that was introduced in PR #1773 (my PR ...) that bit bitrate is ignored when burst is set (and nanosleep is supported).  The bug was that once there was a green light for sending, throttling was never evaluated.


bmah888:
Thanks for the PR, we'll take a look! 

Do you think this problem is specific to Windows (only asking because that was mentioned in the original bug report in #1820)? We'll get this fixed in the next release either way.

davidBar-On:
The problem general for all systems and is **not** specific to Windows.

bmah888 (approved):
Thanks! This looks good (as far as I can tell). I reproduced the bug that this change fixes, and I confirmed that the PR fixes the bug.

PR #1835, opened 2025-02-15
Add missing code for SERVER_ERRROR handling under HAVE_SSL

davidBar-On:
master

* Issues fixed (if any): none

When `test_is_authorized(test)` fails (under `HAVE_SSL`) and `SERVER_ERROR` state is set, the code for sending `errno` to the client is missing.


bmah888 (approved):
Thanks for the PR, this is definitely an error/omission. I'm not sure why the bug didn't result in the client blowing up somehow, but maybe the client was going to error-exit in this situation anyway.

Issue #1830, opened 2025-02-11
Feature Request: Pure UDP Mode (No TCP)


gryphon2411:
Hi iperf3 team,

First of all, thanks for the amazing iperf3.

A quick request: could we get an option for UDP tests to disable the TCP control channel (e.g. `--pure-udp`)?

This is needed for environments where *any* TCP is blocked (firewalls, TCP-restricted connections etc.), _unlike_ iperf2. 

This would be incredibly helpful, since there is no other solution but to use iperf2.

Thanks a lot for considering.

davidBar-On:
@gryphon2411, I submitted PR #1843 that adds support for a UDP control socket, so no UDP should used.  It will be useful if you can test this PR and see that it does work in your system that blocks TCP.

As I commented in the PR, I am not sure about the reliability of the control interface between the server and the client, as packets over UDP socket may be lost.  It will also be useful if you can comment about that (assuming that you can run the PR code successfully).

gryphon2411:
Hi @davidBar-On,

Thank you for submitting PR #1843 and for the work on this feature! 

Following your request, I've put together a Docker Compose environment to test the new `--udp-control` flag.

The included `run-tests.sh` script successfully demonstrates the intended functionality:
1.  Baseline iperf2 UDP test succeeds.
2.  The test simulating standard iperf3 (client without `--udp-control`) correctly fails against the PR server (using `--udp-control`).
3.  The test using the PR build with `--udp-control` on both client and server succeeds.

This confirms the basic operation of the UDP control channel feature. 

While noting @swlars' [comment](https://github.com/esnet/iperf/pull/1843#issuecomment-2749068982) regarding potential test divergence and prioritization, this test confirms the `--udp-control` flag functions as expected to enable UDP tests in environments where TCP control is not feasible.

Details and files for reproduction are below:

**`Dockerfile`**
```dockerfile
# Dockerfile
FROM alpine:latest

# Install build tools, runtime libs, iperf2, and iproute2
RUN apk update && apk add --no-cache \
    build-base git autoconf automake libtool libstdc++ iperf iproute2

# Clone iperf3 repo and checkout PR #1843 branch
RUN git clone https://github.com/esnet/iperf.git /opt/iperf3-src && \
    cd /opt/iperf3-src && \
    git fetch origin pull/1843/head:pr1843 && git checkout pr1843

WORKDIR /opt/iperf3-src

# Configure and Compile iperf3 from PR source
RUN ./configure && make

[...]

PR #1843, opened 2025-02-22
Add support for UDP Control Sockt for UDP tests (no use of TCP)

davidBar-On:
master

* Issues fixed (if any): #1830

Add support for using UDP Control Socket, using a new `--udp-control` option, so there is **no use of TCP** and iperf3 can work over firewalls, etc. that block TCP.  A server using `--udp-control` can only work with clients that use this option, and visa versa.  (Implementing a server that is able to receive connect requests over either TCP or UDP seems to be too complex.)

The changes should not impact any functionality if the new option is not used.   The only changes that are related to TCP control tests are change of few `read()` to `Nread()` to make sure read timeout mechanisms are supported.  (It may be that these changes should have been done anyway.)

No retry mechanisms are added for the control messages, so it may be that a test will fail because a control message was lost over the UDP socket.  However, it seems that the available timeout mechanisms make sure that the server does not get stuck in such case.  Also, once all test sockets are created not control message is sent until the final phase of the test.


swlars:
Hi David, 

Thanks for the pull request! We've definitely gotten feedback that this could be a useful feature, but we're hesitant seeing that the TCP/UDP tests would diverge even more than they do already. We might consider this in the future, but it's not a high priority. 

Thanks again for the pull request, though, I know a lot of people will find this useful.

PR #1846, opened 2025-02-25
json: add UNIX timestamp in milliseconds

phip1611:
json: add UNIX timestamp in milliseconds

This extends the "timestamp" field with a UNIX timestamp in milliseconds
representing the current local wall clock time.

The "timestamp" field now has a new subkey:

    "timestamp": {
      "time": "Tue, 25 Feb 2025 11:49:59 GMT",
      "timesecs": 1740484199,
      // new
      "timemillisecs": 1740484199926
    },

This is helpful when one does fine-grained network tests with iperf,
where iperf events need to be aligned with events coming from other
resources. A time resolution based on seconds is too coarse-grained
for that.

It is enough to use milliseconds here, as networking always comes
with small timing offsets and delays. Using microseconds wouldn't 
add a benefit. 

It is sufficient to only change the code at this place, as the rel time
offset for all further events already includes milliseconds. Currently,
iperf3 prints this time offset in seconds encoded as a double with
a microsecond granularity.

Signed-off-by: Philipp Schuster <philipp.schuster@cyberus-technology.de>

---

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -72,8 +72,9 @@
 #include "net.h"
 #include "iperf.h"
 #include "iperf_api.h"
-#include "iperf_udp.h"
 #include "iperf_tcp.h"
+#include "iperf_time.h"
+#include "iperf_udp.h"
 #if defined(HAVE_SCTP_H)
 #include "iperf_sctp.h"
 #endif /* HAVE_SCTP_H */
@@ -944,7 +945,6 @@ mapped_v4_to_regular_v4(char *str)
 void
 iperf_on_connect(struct iperf_test *test)
 {
-    time_t now_secs;
     const char* rfc1123_fmt = "%a, %d %b %Y %H:%M:%S %Z";
     char now_str[100];
     char ipr[INET6_ADDRSTRLEN];
@@ -953,11 +953,17 @@ iperf_on_connect(struct iperf_test *test)
     struct sockaddr_in *sa_inP;
     struct sockaddr_in6 *sa_in6P;
     socklen_t len;
+    struct iperf_time now;
+    time_t now_secs;
+    unsigned long long now_millisecs;
 
-    now_secs = time((time_t*) 0);
+    iperf_time_now_wallclock(&now);
+    now_millisecs = (time_t) iperf_time_in_usecs(&now) / 1000;
+    now_secs = (time_t) (now_millisecs / 1000);
+
     (void) strftime(now_str, sizeof(now_str), rfc1123_fmt, gmtime(&now_secs));
     if (test->json_output)
-	cJSON_AddItemToObject(test->json_start, "timestamp", iperf_json_printf("time: %s  timesecs: %d", now_str, (int64_t) now_secs));
+	cJSON_AddItemToObject(test->json_start, "timestamp", iperf_json_printf("time: %s  timesecs: %d  timemillisecs: %d", now_str, (int64_t) now_secs, now_millisecs));
     else if (test->verbose)
 	iperf_printf(test, report_time, now_str);
 
--- a/src/iperf_time.c
+++ b/src/iperf_time.c
@@ -36,16 +36,28 @@
 #include <time.h>
 
 int
-iperf_time_now(struct iperf_time *time1)
+clock_gettime_helper(struct iperf_time *now, clockid_t clk_id)
 {
     struct timespec ts;
     int result;
-    result = clock_gettime(CLOCK_MONOTONIC, &ts);
+    result = clock_gettime(clk_id, &ts);
     if (result == 0) {
-        time1->secs = (uint32_t) ts.tv_sec;
-        time1->usecs = (uint32_t) ts.tv_nsec / 1000;
+        now->secs = (uint32_t) ts.tv_sec;
+        now->usecs = (uint32_t) ts.tv_nsec / 1000;
     }
     return result;
+}
+
+int
+iperf_time_now(struct iperf_time *now)
+{
+    return clock_gettime_helper(now, CLOCK_MONOTONIC);
+}
+
+int
+iperf_time_now_wallclock(struct iperf_time *now)
+{
+    return clock_gettime_helper(now, CLOCK_REALTIME);
 }
 
 #else
@@ -53,14 +65,21 @@ iperf_time_now(struct iperf_time *time1)
 #include <sys/time.h>
 
 int
-iperf_time_now(struct iperf_time *time1)
+iperf_time_now_wallclock(struct iperf_time *now)
 {
     struct timeval tv;
     int result;
+    // Returns the non monotonic local wallclock time.
     result = gettimeofday(&tv, NULL);
-    time1->secs = tv.tv_sec;
-    time1->usecs = tv.tv_usec;
+    now->secs = (uint32_t) tv.tv_sec;
+    now->usecs = (uint32_t) tv.tv_usec;
     return result;
+}
+
+int
+iperf_time_now(struct iperf_time *now)
+{
+    return iperf_time_now_wallclock(now);
 }
 
 #endif
--- a/src/iperf_time.h
+++ b/src/iperf_time.h
@@ -34,6 +34,24 @@ struct iperf_time {
     uint32_t usecs;
 };
 
+/**
+ * Retrieves the current wallclock time.
+ *
+ * When during operation the system-wide clock changes, users will see
+ * this and results might become inconsistent.
+ */
+int iperf_time_now_wallclock(struct iperf_time *time1);
+
+/**
+ * Retrieves the current time using a monotonic clock.
+ *
+ * When during operation the system-wide clock changes, users still will see
+ * a strictly monotonic increasing clock.
+ *
+ * Please note that a monotonic clock is only available on systems that have the
+ * `clock_gettime()` system call. On other systems, this falls back to
+ * `iperf_time_now_wallclock()`.
+ */
 int iperf_time_now(struct iperf_time *time1);
 
 void iperf_time_add_usecs(struct iperf_time *time1, uint64_t usecs);
@@ -44,6 +62,9 @@ int iperf_time_diff(struct iperf_time *time1, struct iperf_time *time2, struct i
 
 uint64_t iperf_time_in_usecs(struct iperf_time *time);
 
+/**
+ * Returns the time in seconds as double type with a microsecond granularity.
+ */
 double iperf_time_in_secs(struct iperf_time *time);
 
 #endif


src/iperf_api.c
```diff
@@ -953,11 +953,15 @@ iperf_on_connect(struct iperf_test *test)
...
-    now_secs = time((time_t*) 0);
-    (void) strftime(now_str, sizeof(now_str), rfc1123_fmt, gmtime(&now_secs));
+    clock_gettime(CLOCK_REALTIME, &time_spec);
+    // Unix timestamp in milliseconds (local wall clock time)
+    now_millisecs = time_spec.tv_sec * 1000000 + time_spec.tv_nsec / 1000000;
+
+    (void) strftime(now_str, sizeof(now_str), rfc1123_fmt, gmtime(&time_spec.tv_sec));
     if (test->json_output)
-	cJSON_AddItemToObject(test->json_start, "timestamp", iperf_json_printf("time: %s  timesecs: %d", now_str, (int64_t) now_secs));
+	cJSON_AddItemToObject(test->json_start, "timestamp", iperf_json_printf("time: %s  timesecs: %d  timestamp_ms: %d", now_str, (int64_t) time_spec.tv_sec, now_millisecs));
```
phip1611, line 964: hint for reviewers: The `%d` is correct as `iperf_json_printf()` only knows `%d` for json numbers, no `%lld` or similar

src/iperf_api.c
```diff
@@ -953,11 +953,15 @@ iperf_on_connect(struct iperf_test *test)
...
-    now_secs = time((time_t*) 0);
-    (void) strftime(now_str, sizeof(now_str), rfc1123_fmt, gmtime(&now_secs));
+    clock_gettime(CLOCK_REALTIME, &time_spec);
+    // Unix timestamp in milliseconds (local wall clock time)
+    now_millisecs = time_spec.tv_sec * 1000000 + time_spec.tv_nsec / 1000000;
+
+    (void) strftime(now_str, sizeof(now_str), rfc1123_fmt, gmtime(&time_spec.tv_sec));
     if (test->json_output)
-	cJSON_AddItemToObject(test->json_start, "timestamp", iperf_json_printf("time: %s  timesecs: %d", now_str, (int64_t) now_secs));
+	cJSON_AddItemToObject(test->json_start, "timestamp", iperf_json_printf("time: %s  timesecs: %d  timemillisecs: %d", now_str, (int64_t) time_spec.tv_sec, now_millisecs));
```
snue, line 964: nit: I won't claim I like that new key name. I see it fits with the existing one. Shorter alternatives would be `time_ms` or `timemsecs` ... I'm fine with using the proposed one here.

snue (changes requested):
I see the need for a more accurate timestamp than second granularity occasionally. I like that this adds a separate new key instead of increasing the precision of existing timestamps. This should cause less churn for tools working with the existing schema.

I found an issue in the timestamp computation though.

src/iperf_api.c
```diff
@@ -953,11 +953,15 @@ iperf_on_connect(struct iperf_test *test)
     struct sockaddr_in *sa_inP;
     struct sockaddr_in6 *sa_in6P;
     socklen_t len;
+    unsigned long long now_millisecs;
 
-    now_secs = time((time_t*) 0);
-    (void) strftime(now_str, sizeof(now_str), rfc1123_fmt, gmtime(&now_secs));
+    clock_gettime(CLOCK_REALTIME, &time_spec);
+    // Unix timestamp in milliseconds (local wall clock time)
+    now_millisecs = time_spec.tv_sec * 1000000 + time_spec.tv_nsec / 1000000;
```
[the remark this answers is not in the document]
phip1611: Yikes. Copy and paste, thanks!

src/iperf_api.c
phip1611: Any preference by the code owners? @bmah888

snue (approved):
LGTM. I'd be happy to get millisecond granularity for the start timestamp.

swlars:
Thanks for the pull request! This seems like a nice, straightforward addition, so we'll try to get to it soon.

phip1611:
Any update here? @swlars

src/iperf_api.c
```diff
@@ -944,7 +944,7 @@ mapped_v4_to_regular_v4(char *str)
 void
 iperf_on_connect(struct iperf_test *test)
 {
-    time_t now_secs;
+    struct timespec time_spec;
```
bmah888, line 947: Thanks for the PR and apologies it's taken so long to actually look at it!

First, I agree with @swlars that this is good functionality to have. As I'm reviewing the change it's not quite as easy as we thought originally (for reasons that are entirely due to existing iperf3 code). We might want to implement this slightly differently to use some abstractions for some of the time library calls. Basically iperf3 started off not using `clock_gettime()` in any way (using `gettimeofday()` instead), but it was added somewhere around iperf-3.7 (see PR #738). At the time it wasn't clear if just switching over to `clock_gettime()` would break anything, so the aforementioned PR added some abstraction and conditional compilation to make sure that iperf3 would use `clock_gettime()` on hosts that have it, and `gettimeofday()` on hosts that didn't. That's all in `iperf_time.[ch]`.

Trying to use those abstractions would mean doing something like this (completely untested):
```
struct iperf_time now;
uint64_t now_millisecs;
time_t now_secs;

iperf_time_now(&now);
now_millisecs = iperf_time_in_usecs(&now) / 1000L;
now_secs = (time_t) iperf_time_in_secs(&now); // <-- yes this is awful

(void) strftime(now_str, sizeof(now_str), rfc1123_fmt, gmtime(&now_secs));

if (test->json_output)
  cJSON_AddItemToObject(test->json_start, "timestamp", iperf_json_printf("time: %s timesecs: %d timemillisecs: %d", now_str, (int64_t) now_secs, now_millisecs))
```

I'm not super-happy with what I just wrote, maybe you can come up with a better way to do this.

I note that so far I haven't seen an example of a host that runs iperf3 that doesn't also support `clock_gettime()` so it might be possible to just assume we always have it and get rid of a bunch of the backward compatibility stuff. That's kind of a bigger task though.

src/iperf_api.c
bmah888: I have a weak preference for `timemsecs`

src/iperf_api.c
phip1611: Hey! Reusing `iperf_time.h` makes total sense, no need to reinvent the weel. I adapted the code with one change. No need for 

`now_secs = (time_t) iperf_time_in_secs(&now); // <-- yes this is awful`

when we also can use 

`now_secs = (time_t) (now_millisecs / 1000);`

bmah888: I'm leaning towards a new function that is like `iperf_time_now()` but uses `CLOCK_REALTIME` instead. The existing use of `CLOCK_MONOTONIC` was there for a specific reason (according to the person who did PR #738), although I totally understand the reasoning for `CLOCK_REALTIME` for _this_ use case.

Note we need some way to do the same thing for the case where the host doesn't support `clock_gettime()`, which I'm not sure is even applicable to modern UNIX-like systems.

I have a couple build fixes that I'll comment on separately. Thanks for working with me on this!

src/iperf_api.c
```diff
@@ -953,11 +954,19 @@ iperf_on_connect(struct iperf_test *test)
...
     socklen_t len;
+    time_t now_secs;
+    unsigned long long now_millisecs;
+
+    iperf_time_now(&now);
+    now_millisecs = (time_t) iperf_time_in_usecs(&now) / 1000;
+    now_secs = (time_t) (now_millisecs / 1000);
+#include <stdio.h>
+    printf("millisecs= %lld ms\n", now_millisecs);
+    printf("secs= %ld s\n", now_secs);
```
bmah888, line 965: Is this debugging code?

src/iperf_api.c
```diff
@@ -953,11 +954,19 @@ iperf_on_connect(struct iperf_test *test)
     struct sockaddr_in *sa_inP;
     struct sockaddr_in6 *sa_in6P;
     socklen_t len;
+    time_t now_secs;
```
bmah888, line 957: This duplicates the declaration of `now_secs` in line 948. However we do need a declaration of `struct iperf_time now;` somewhere in this function.

src/iperf_api.c
phip1611: Yikes, sorry! This wasn't supposed to end up in my commit

phip1611:
I rebased the PR and refactored the code so that the iperf_time module/subsystem can report monotonic and real/wallclock time.

src/iperf_api.c
```diff
@@ -953,11 +954,17 @@ iperf_on_connect(struct iperf_test *test)
     struct sockaddr_in *sa_inP;
     struct sockaddr_in6 *sa_in6P;
     socklen_t len;
+    struct iperf_time now;
+    time_t now_secs;
```
snue, line 958: This local variable is already declared above.

src/iperf_time.h
```diff
@@ -34,6 +34,28 @@ struct iperf_time {
     uint32_t usecs;
 };
 
+/**
+ * Returns the current wallclock time.
```
snue, line 38:
```suggestion
/**
 * Retrieve the current wallclock time.
```

It returns the success code of the operation. The time is in the out-parameter. Same for the monotonic function.

src/iperf_time.c
```diff
@@ -31,35 +31,62 @@
...
+iperf_time_now_wallclock(struct iperf_time *now)
 {
     struct timeval tv;
     int result;
+    // Returns the non monotonic local wallclock time.
     result = gettimeofday(&tv, NULL);
-    time1->secs = tv.tv_sec;
-    time1->usecs = tv.tv_usec;
+    now->secs = tv.tv_sec;
+    now->usecs = tv.tv_usec;
```
snue, line 89: You don't cast here, but do it above. The `struct timeval` uses `time_t` and `suseconds_t`. The `struct timespec` from the `clock_gettime` call uses `time_t` and `long`. The `iperf_time` uses plain `int` for both. Let's cast here as well.

src/iperf_time.c
```diff
@@ -31,35 +31,62 @@
...
 #else
 
 #include <sys/time.h>
 
 int
-iperf_time_now(struct iperf_time *time1)
+iperf_time_now_monotonic(struct iperf_time *now)
+{
+    // Fall back to wallclock, we can't use monotonic clock.
+    return iperf_time_now_wallclock(now);
```
snue, line 78: I don't fancy that explicitly calling the `monotonic` variant falls back to `wallclock`. I'd rather we define `iperf_time_now()` here, and make it monotonic if possible, wallclock otherwise. That is basically the same behavior as you have now, but without the intermediate call to the "non-monotonic" monotonic function.

src/iperf_api.c
phip1611: yikes, forgot to add that change to the commit; fixed now

src/iperf_time.c
phip1611: good catch

src/iperf_time.c
phip1611: Fixed; i think the new version is a little more streamlined

src/iperf_time.h
phip1611: good idea, thanks!

src/iperf_time.c
snue: I like the new version and the added comments :+1:

bmah888 (approved):
Looking good now! Compiles and runs on our main supported platforms.

bmah888:
@phip1611 Merged. Thanks for the PR and revisions!

@snue Thanks for help reviewing, much appreciated.

Issue #1847, opened 2025-02-28
regression introduced with #1821: bitrate drop to zero


O-ring:
Hello everyone,

After commit #1821 on february 7, 2025, a regression was introduced: after a few seconds, the bitrate will drop to zero.

This is the result I get when I run an udp test:

iperf3  -V -c 10.9.99.125 -R -u -b 314M  -t 10

2025-02-28 13:26:47 CET iperf 3.18+
Control connection MSS 1448
Setting UDP block size to 1448
2025-02-28 13:26:47 CET Time: Fri, 28 Feb 2025 12:26:47 GMT
2025-02-28 13:26:47 CET Connecting to host 10.9.99.125, port 5201
2025-02-28 13:26:47 CET Reverse mode, remote host 10.9.99.125 is sending
2025-02-28 13:26:47 CET       Cookie: wu3byqxtng2f57jvq5zd5nyvxqbssjqyscm7
2025-02-28 13:26:47 CET       Target Bitrate: 314000000
2025-02-28 13:26:47 CET [  5] local 10.9.99.126 port 56848 connected to 10.9.99.125 port 5201
2025-02-28 13:26:47 CET Starting Test: protocol: UDP, 1 streams, 1448 byte blocks, omitting 0 seconds, 10 second test, tos 0
2025-02-28 13:26:48 CET [ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
2025-02-28 13:26:48 CET [  5]   0.00-1.00   sec  37.5 MBytes   314 Mbits/sec  0.049 ms  0/27124 (0%)  
2025-02-28 13:26:49 CET [  5]   1.00-2.00   sec  37.4 MBytes   314 Mbits/sec  0.048 ms  0/27114 (0%)  
2025-02-28 13:26:50 CET [  5]   2.00-3.00   sec  37.4 MBytes   314 Mbits/sec  0.051 ms  0/27106 (0%)  
2025-02-28 13:26:51 CET [  5]   3.00-4.00   sec  31.6 MBytes   265 Mbits/sec  0.059 ms  0/22904 (0%)  
2025-02-28 13:26:52 CET [  5]   4.00-5.00   sec  0.00 Bytes  0.00 bits/sec  0.059 ms  0/0 (0%)  
2025-02-28 13:26:53 CET [  5]   5.00-6.00   sec  0.00 Bytes  0.00 bits/sec  0.059 ms  0/0 (0%)  
[...]

PR #1848, opened 2025-02-28
Extending PR #1821 fix to PR #1773 for sending when bitrate is set

davidBar-On:
master

* Issues fixed (if any): #1847

Further fix to #1773 that was partially fixed by #1821.  This fix here is making sure that after sending, the conditions for green light will be evaluated if no message was send.  The issue was actually caused by wrong copy/paste (did not copy the `|| message_sent == 0`) 😞...

@bmah888, not that this seems to be a **quite urgent issue**.  It happens with **short packets much more frequently** (this is why I didn't find it in my tests).  My understanding is that this is because the long enough delays between packets, the micro-sleep mechanism resolution is good enough, and the delays between packets are bigger for large packets.


swlars:
Looks good! Thanks for the pull request.

PR #1861, opened 2025-03-21
Print an error message when a Send/Receive Worker Thread fail

davidBar-On:
master

* Issues fixed (if any): none

Add error message if a Send/Receive Worker Thread fail.  Currently such failure is "silent".


bmah888:
Thanks for the pull request! I agree this will give some better visibility into thread creation errors, and that'll be helpful.

PR #1863, opened 2025-03-25
replace strncat (unsafe) with strlcat (safe)

warrenburch:
Some code bases will not allow strncat anymore for safety concerns. Replace with strlcat, note when using strlcat you simply pass the full size of the buffer (you do not account for the NULL terminator, this is handled for you).

I tested this by forcing inclusion of features using ...

CFLAGS='-DHAVE_CPU_AFFINITY -DHAVE_FLOWLABEL  -DHAVE_TCP_CONGESTION -DHAVE_SENDFILE -DHAVE_SO_MAX_PACING_RATE -DHAVE_SO_BINDTODEVICE -DHAVE_DONT_FRAGMENT -DHAVE_PTHREAD' ./configure

NOTE: I tried to include SCTP and SSL but both failed to compile for missing headers.

I tested the 

TAILQ_FOREACH(t, &(test->server_output_list), textlineentries) {
		    strlcat(output, t->line, buflen+1);		 
            printf("otuput line: %s\n", t->line);
		}

section by forcing

        struct iperf_textline *line1 = malloc(sizeof(*line1));
        struct iperf_textline *line2 = malloc(sizeof(*line2));
        
        line1->line = strdup("First line\n");
        line2->line = strdup("Second line\n");
        
        TAILQ_INSERT_TAIL(&(test->server_output_list), line1, textlineentries);
        TAILQ_INSERT_TAIL(&(test->server_output_list), line2, textlineentries);

beforehand. Tested error path by forcing error in server, force the IEDAEMON error with errno=999. Error string was as expected.

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2555,8 +2555,7 @@ send_results(struct iperf_test *test)
 		/* Allocate and build it up from the component lines */
 		char *output = calloc(buflen + 1, 1);
 		TAILQ_FOREACH(t, &(test->server_output_list), textlineentries) {
-		    strncat(output, t->line, buflen);
-		    buflen -= strlen(t->line);
+		    strlcat(output, t->line, buflen+1);		    
 		}
 
 		cJSON_AddStringToObject(j, "server_output_text", output);
--- a/src/iperf_error.c
+++ b/src/iperf_error.c
@@ -556,11 +556,11 @@ iperf_strerror(int int_errno)
 
     /* Append the result of strerror() or gai_strerror() if appropriate */
     if (herr || perr)
-        strncat(errstr, ": ", len - strlen(errstr) - 1);
+        strlcat(errstr, ": ", len);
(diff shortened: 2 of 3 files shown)


bmah888:
Thanks for the pull request! We'll take a look.

Just noting for reference that the `strncat(3)` manual page has this to say about `strlcat(3)`:

```strlcat() is not present in glibc and is not standardized by POSIX, but is available on Linux via the libbsd library.```

warrenburch:
Thanks Bruce, good catch on glibc

warrenburch:
Is there any update on this? will this be adopted?

bmah888:
> Is there any update on this? will this be adopted?

To be honest I'm thinking not, because it's pretty platform-specific. We can't get rid of the strncat-using code entirely because there are platforms that just don't have strlcat. However I appreciate your sending the patch!

Issue #1865, opened 2025-03-26
Iperf3 3.18 version inconsistent results


FlamurMustafa:
I have this iperf3 test that is done between a server and client, but the results are not consistent between the two entities regarding total packet sent (6276 and 6277 on total).

*The session is being run in reverse mode.
*Client is running 3.18, and server is running version 3.7 of iperf3.
*Using 3.10 client and 3.7 server doesn't have this issue. Using 3.18 client with 3.18 server does have this issue.

Client Side:
*Arguments --bitrate 10.0M, --ulength 2000 (fragmentation enabled with intent).

![Image](https://github.com/user-attachments/assets/68f5e1ee-3086-4534-92fd-b6c2b401e877)

Server side:

![Image](https://github.com/user-attachments/assets/74e97f3a-d042-45cf-b0ac-e56d88df0aca)

I took a look at this [discussion](https://github.com/esnet/iperf/discussions/1541), but the env that I'm using doesn't buffer anything.

FlamurMustafa:
Shouldn't this buffering be present on older versions as well then, since it started from version 3.18? Also, tinkering around with ulength (2200 for example, )the said issue doesn't appear so often.

Is this an expected behavior of iperf?

davidBar-On:
Its difficult to tell what exactly is going on without traffic logs (e.g. using Wireshark).  However, since version 3.16 iperf3 is multi-threaded, and the control messages are sent between different threads than the threads for the data messages (until version 15 there was one thread for both).  Therefore probably it may happen that the "test end" control messages are handled by the control thread be for the data thread handles the last message(s).

FlamurMustafa:
![Image](https://github.com/user-attachments/assets/a47e45f7-a7ed-4850-be79-9bd93dda7a06)

So I found this at the end of the test, where in this interval 0.001563 there is one packet send after the client requests to end the communication, shouldn't there be no packets send after the TEST_END has been received.

This is not present on earlier versions.

davidBar-On:
> ... shouldn't there be no packets send after the TEST_END has been received.
This is not present on earlier versions.

[...]

PR #1868, opened 2025-03-29
UDP continute receiving in TEST_END state to allow receing last packets

davidBar-On:
master (3.18+)

* Issues fixed (if any): #1865

In UDP test, continue to receive packets during `TEST_END` state (until `PARAMETERS_EXCHANGE` state).  In reverse or bi-directional mode, that allows the Client statistics to properly take into account the packets that were sent from the Server between the time the Client switched to `TEST_END` and the time the server received and switched to this state.


bmah888:
Thanks for the pull request! We'll study the code change (and the original bug report).

PR #1887, opened 2025-05-17
RELNOTES: fix Multipath-TCP flag

matttbe:
Thank you for the recent v3.19 release! 

  `3.1-STABLE`) to which this pull request applies: 3.19

* Issues fixed (if any): /

* Brief description of code changes (suitable for use as a commit message): The option for Multipath-TCP support is apparently `--mptcp`, not `--multipath`. See: 

https://github.com/esnet/iperf/blob/57396df3c19227741b8cd40c24b105d28672b6f7/src/iperf_api.c#L1165

While at it, mention `-m` which is the short option.

Note that the text in the release section also needs to be updated: https://github.com/esnet/iperf/releases/tag/3.19

### DIFF
--- a/RELNOTES.md
+++ b/RELNOTES.md
@@ -7,7 +7,7 @@ iperf-3.19 2025-05-16
 * Notable user-visible changes
 
     * iperf3 now supports the use of Multi-Path TCP (MPTCPv1) on Linux
-      with the use of the `--multipath` flag. (PR #1661)
+      with the use of the `-m` or `--mptcp` flag. (PR #1661)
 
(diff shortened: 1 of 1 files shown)


bmah888 (approved):
Oops! Thanks for the patch, I'll get this committed shortly.

bmah888:
> Does this mean there will be a 3.19.1 (or similar) release soon?

Good question. At this point, we're not planning on doing a patch release just to fix the release notes. But if we need to do a patch release for some other reason, we'd probably pull the release notes fix into that.

Issue #1871, opened 2025-04-04
iperf3: use incompressible data for streaming test


rscheff:
Various access technologies, as well as transparent WAN accellerators, TCP proxies, VPNs and Encryption protocols willl try to compress (gz) the data stream and/or the packets. This is particularly notable by sending zeroed data payload, or when the data buffer which is used to send, is relatively small (few kB) - this can lead to results of throuhgput which are not attainable with normally compressible data, or incompressible data (e.g. media streaming data).

iperf3 should optionally have the capability to allocate a larger chunk of main memory, and fill this with random data or pre-compressed data (alternatively, consume a file / stdin with user-provided data to send, which is sent repeatedly when the data runs out before the test is over).

davidBar-On:
Per iprf3 code, the iperf3 packets data is already randomized using "_/dev/urandom_".  However, each message (packet) includes the same contents.  I wrote the data into a file using iperf3 `-F` option and then compressed in using zip.  The compression results:
1. **One packet** (1KB / 31KB, 32KB 128KB): 0% compression.
2. **4,000 packets of 1KB or 31KB**: about 99% compression.
3. **4,000 packets of of 32KB**: 0% compression.

I am not familiar of the zip compression algorithm but from the above results it seems that it is effective for **repeated random values** only up to length of 32KB.  **Do you know what is that length for the network components you mentioned?**  I.e. what should be the size of the iperf3 "larger chunk of main memory"?

davidBar-On:
Submitted PR #1907 with proposed solution - allow to send blocks with different randomized data in each (up to certain limit and then repeating the blocks).

PR #1907, opened 2025-07-12
Add option to send different randomized blocks

davidBar-On:
master

* Issues fixed (if any): #1871

Add a second option to `-l` to specify number of different (randomized) blocks to send.  The purpose is that not all sent blocks will be the same, and as the blocks contents is randomized by default, network compression will not have an effect.

Test, that stored the received data for 1KB blocks into a file, shows that when sending the same block contents, zip of the file achieves more than 99% compression.   On the other hand, when the block contents repeats only every 100 blocks, zip achieves almost no compression.

The change includes also initializing a stream buffer only when it is a sending stream.


bmah888:
Thanks for the PR, we'll take a look!

PR #1914, opened 2025-07-25
server_timer_proc - better handle server test time expiration

davidBar-On:
make

* Issues fixed (if any): none

Enhancements to `server_timer_proc()` for better handling of server test time expiration


bmah888:
Thanks for the PR! Will review.

bmah888:
I tried fixing up a couple merge conflicts so that we can test this a bit with `master`-branch based code. Let's see how this goes.

davidBar-On:
> I'm trying to figure out how the thread associated with the stream gets cancelled / ended. From what I can tell, this does happen but I can't tell how or where. Can you help me understand this part?

Regarding the question, because the sockets are closed and `test->done` is set, the `worker_run()` will terminate either per the `test->done` or because read/write will fail.  That terminates the threads.

Regarding the reason for the PR, I don't remember what was the exact problem, and for some reason I didn't described it, but I believe it was:
1. `server_timer_proc()` close the control socket.
2. `select()` fail and `cleanup_server(test)` or client's `goto cleanup_and_fail` are executed.
3. The threads are terminated using `pthread_cancel()`.
4. If a thread is reading/writing when it is canceled than that may cause core dump.  The purpose of the PR is to prevent this core dump.

If I remember correctly, I added closing the sockets in `server_timer_proc()` and did not add threads canceling, since I was not sure if canceling will not cause problems in the code executed after `select()` fails.

PR #1927, opened 2025-08-13
docs: clarify --get-server-output needs the test to complete

rmelotte:

* Issues fixed (if any): Documentation clarification.

When using '--get-output-server', it's not immediately clear that the client will get the server output only if the test runs to completion.

To avoid people spending time debugging why they sometimes don't get the server output, mention it explicitly in the doc as well as in the man page.


bmah888 (approved):
Thanks for the PR! Will merge this shortly.

Note that you don't need to change `invoking.rst`, approximately once per release we'll regenerate that part of the documentation from the manual page sources.

PR #1932, opened 2025-08-27
Send Server's get_Parameters() errors to the client

davidBar-On:
master

* Issues fixed (if any): none

In case `get_parameters()` failure, send `SERVER_ERROR` state to the client, following by the error code (`i_errno`) set inside `get_parameters` (e.g. the `IEUDPFILETRANSFER` error in #1909).

This is actually a generalization of already existing mechanism for specific error codes.  E.g., with this PR changes, PR #1684 can just set the `IEMAXSERVERTESTDURATIONEXCEEDED` error in `get_parameters()`, without adding sending `SERVER_ERROR`, etc. specifically for that error.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2247,8 +2247,21 @@ iperf_exchange_parameters(struct iperf_test *test)
 
     } else {
 
-        if (get_parameters(test) < 0)
+        if (get_parameters(test) < 0) {
+            if (iperf_set_send_state(test, SERVER_ERROR) != 0)
+                return -1;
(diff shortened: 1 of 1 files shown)


manedurphy:
I think I'd like to leverage this change for [iperf/#1931](https://github.com/esnet/iperf/pull/1931) as well. I can validate the requested bit rate within `get_parameters` similar to how you validated the use of `--file` and `--udp` used together.

bmah888:
Thanks for the PR! I think this is the right way to solve the problem.

Intending to merge this after some testing.

PR #1946, opened 2025-09-24
server: ignore rcv-timeout in BIDIRECTIONAL mode

### LINKED ISSUE #1766: iperf3: error - control socket has closed unexpectedly (iperf 3.17.1 ) on Debian 12.7
* Version of iperf3:
Server: iperf 3.17.1 (cJSON 1.7.15)
Client: iperf 3.17.1 (cJSON 1.7.15)

* Hardware:
Server: ARMv7 VM Running Debian
Client VM AMD64

Server: Debian 12.7
Client: Debian 12.7

# Bug Report
Client 15.137.211.151
Server 15.137.211.152

Both systems are connected via LAN on the same switch.

When I run the test for the first time, I get an error message:

root@gj-inst-w10-amd64:~# iperf3 --client 15.137.211.152 --get-server-output --bitrate 1M --interval 1 --time 10 --len 0 --parallel 4 --bidir
Connecting to host 15.137.211.152, port 5201
[  5] local 15.137.211.151 port 45090 connected to 15.137.211.152 port 5201
[  7] local 15.137.211.151 port 45100 connected to 15.137.211.152 port 5201
[  9] local 15.137.211.151 port 45116 connected to 15.137.211.152 port 5201
[ 11] local 15.137.211.151 port 45128 connected to 15.137.211.152 port 5201
[ 13] local 15.137.211.151 port 45144 connected to 15.137.211.152 port 5201
[ 15] local 15.137.211.151 port 45148 connected to 15.137.211.152 port 5201
[ 17] local 15.137.211.151 port 45164 connected to 15.137.211.152 port 5201
[ 19] local 15.137.211.151 port 45172 connected to 15.137.211.152 port 5201
root@gj-inst-w10-amd64:~#

If I run it right after again, all is working fine, here the last lines:
[ 13][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec    0             sender
[ 13][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec                  receiver
[ 15][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec    0             sender
[ 15][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec                  receiver
[ 17][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec    0             sender
[ 17][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec                  receiver
[ 19][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec    0             sender
[ 19][RX-C]   0.00-10.00  sec  1.25 MBytes  1.05 Mbits/sec                  receiver
[SUM][RX-C]   0.00-10.00  sec  5.00 MBytes  4.19 Mbits/sec    0             sender
[SUM][RX-C]   0.00-10.00  sec  5.00 MBytes  4.19 Mbits/sec                  receiver

iperf Done.

I have attached a trace of the server side. In the first failing test one side is resetting in the middle of the connection.

I can reproduce it every time. If there is more time between 2 tests ( more than around 3 minutes) then the first test is failing again. If I run many test after the previous test ended all is fine, if there is less time between tests

[iperf.zip](https://github.com/user-attachments/files/17006738/iperf.zip)

scarpein:
### What & Why
In `--bidir` tests, the server could prematurely abort due to `rcv-timeout`.  
However, when running in **BIDIRECTIONAL mode**, the receive timeout should not apply, since the send side is governed by `--snd-timeout`.

This addresses [#1766](https://github.com/esnet/iperf/issues/1766):  
`iperf3: error - control socket has closed unexpectedly (iperf 3.17.1) on Debian 12.7`.

### Fix
Exclude BIDIRECTIONAL in the rcv-timeout condition:

```diff
- else if (test->mode != SENDER && t_usecs > rcv_timeout_us) {
+ else if (test->mode != SENDER && test->mode != BIDIRECTIONAL && t_usecs > rcv_timeout_us) {
```

### Behavior Change

* **Unchanged:** Pure receiver mode (`client → server` one-way) still uses `rcv-timeout`.
* **Changed:** In `--bidir`, the server no longer aborts on `rcv-timeout`; termination follows normal end conditions (or `--snd-timeout` if specified).

### Reproduction / Verification

**Server:**

```bash
iperf3 -s --rcv-timeout 5000 --snd-timeout 15000 -1 -J
```

**Client:**

```bash
iperf3 -c <server> --bidir -t 10 -J
```

* **Before (unpatched):** Server may terminate early with an `rcv-timeout`.
* **After (patched):** No premature termination under `--bidir`.

### Notes

* Built and tested locally using:

  ```bash
  ./bootstrap.sh   # only if ./configure fails
  ./configure
  make
  ```
* Minimal one-line change, consistent with intended semantics of `--bidir`.
* Closes #1766

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -672,7 +672,7 @@ iperf_run_server(struct iperf_test *test)
                  * Running a test. If we're receiving, be sure we're making
                  * progress (sender hasn't died/crashed).
                  */
-                else if (test->mode != SENDER && t_usecs > rcv_timeout_us) {
+                else if (test->mode != SENDER && test->mode != BIDIRECTIONAL && t_usecs > rcv_timeout_us) {
                     /* Idle timeout if no new blocks received */
                     if (test->blocks_received == last_receive_blocks) {
                         test->server_forced_no_msg_restarts_count += 1;


bmah888:
Thanks for the pull request! We'll take a look.

davidBar-On:
> In **BIDIRECTIONAL mode (`--bidir`)**, the server may abort prematurely due to the **receive timeout (`rcv-timeout`)**.
> However, in bidirectional mode the receive timeout should not apply, since the send side is governed by **`--snd-timeout`**.
> ...
> Fix ... Exclude BIDIRECTIONAL mode from the `rcv-timeout` condition:

`rcv-timeout` is used to **protect the server from being stuck** because of a problem in the test, **any type** of test.  On the other hand, `snd-timeout` is applicable only to TCP tests and is using the TCP `TCP_USER_TIMEOUT` option.  Also, `snd-timeout` is optional, which `rcv-timeout` has a default of 2 minutes.  Therefore, `rcv-timeout` cannot be just ignored, probably even for TCP tests when `snd-timeout` is set.

A simple solution for the case that `rcv-timeout` is smaller then `snd-timeout` may be to not let such test to run - on the test initialization on the server issue an error in such case.

>     * Related issue: [iperf3: error - control socket has closed unexpectedly (iperf 3.17.1 ) on Debian 12.7 #1766](https://github.com/esnet/iperf/issues/1766) (`iperf3: error - control socket has closed unexpectedly`).
>     * **Additional scenario:**
>       * When the server has been started and the `rcv-timeout` period has elapsed, if a client initiates a `--bidir` test, the server incorrectly interprets this as a receive-timeout event and calls `cleanup_server(test)`, closing the control socket.
>       * If the client immediately runs a second test, it proceeds normally.
>       * However, if another test is started only after another `rcv-timeout` interval has passed, the same issue reoccurs (`rcv-timeout` triggers → `cleanup_server(test)` executes → control socket closes).

**Can you give more information for how you got to this conclusion?**  E.g. by providing the relevant server outputs when `--timestamps` is used (to see the timings of the events).

The reason I am asking is that `rcv-timeout` is only used by the `select()` in the main thread during a test.  Therefore, unless there is a bug in the implementation, there should not be any relation between consecutive tests.

scarpein:
@davidBar-On Thank you for your response. After reviewing your comments, I took a closer look at the issue.
I acknowledge that my initial approach was too simplistic.

Here is my detailed view on the root cause and potential solution:

1. The server sets `last_receive_time` using `iperf_time_now(&last_receive_time)` after executing `iperf_server_listen(..)`.
2. The client connects and starts the bidirectional measurement.
3. If, within the first second after establishing send/receive ports, the client sends data but no data is received, the `last_receive_time` remains as the server start time (set in step 1).
4. If the `rcv_timeout` elapses after the server starts, the following condition will inevitably be met:

   ```c
   else if (test->mode != SENDER && t_usecs > rcv_timeout_us) {
       ...
       if (test->blocks_received == last_receive_blocks)
   }
   ```
5. Under poor network conditions, when the client performs a bidirectional test at high bandwidth, this symptom occurs quite frequently.
6. I agree that my earlier workaround for BIDIRECTIONAL mode was only a temporary and unsafe solution.
7. I believe the correct logic is that once the client-server connection is established, the `rcv_timeout` check should be guaranteed for a certain period (instead of disconnecting immediately as it does now).
8. Therefore, I see two possible ways to address the issue caused by `cleanup_server(..)` after the condition check:

   * **8.1** Only check when `last_receive_blocks > 0` after connection. → But this risks infinite waiting if no packets are ever received from the client, so it is not acceptable.
   * **8.2** Set `last_receive_time` to `now` at the moment the server receives the measurement request from the client. → This way, the `rcv_timeout` is properly guaranteed after connection.

---

**Here is the server log output observed during the issue:**

```
Mon Sep 29 07:36:33 2025 -----------------------------------------------------------
Mon Sep 29 07:36:33 2025 Server listening on 50000 (test #3)
Mon Sep 29 07:36:33 2025 -----------------------------------------------------------
Mon Sep 29 07:36:39 2025 Time: Mon, 29 Sep 2025 07:36:39 GMT
Mon Sep 29 07:36:39 2025 Accepted connection from 175.113.80.207, port 55266
Mon Sep 29 07:36:39 2025       Cookie: uwgm4bowwhxsriojmpivypjccksshd46fzwj
Mon Sep 29 07:36:39 2025       TCP MSS: 0 (default)
Mon Sep 29 07:36:39 2025       Target Bitrate: 100000000
Mon Sep 29 07:36:39 2025 [  5] local 211.239.165.14 port 50000 connected to 175.113.80.207 port 55280
Mon Sep 29 07:36:39 2025 [  8] local 211.239.165.14 port 50000 connected to 175.113.80.207 port 55282
Mon Sep 29 07:36:39 2025 Starting Test: protocol: TCP, 1 streams, 131072 byte blocks, omitting 0 seconds, 3 second test, tos 0
Mon Sep 29 07:36:40 2025 iperf3: Server restart (#1) during active test due to idle timeout for receiving data
Mon Sep 29 07:36:40 2025 iperf3: error - idle timeout for receiving data
Mon Sep 29 07:36:40 2025 iperf 3.19.1+
Mon Sep 29 07:36:40 2025 Mon Sep 29 07:36:40 2025 Linux sti-iperf-server 6.8.0-57-generic #59-Ubuntu SMP PREEMPT_DYNAMIC Sat Mar 15 17:40:59 UTC 2025
Mon Sep 29 07:36:40 2025 -----------------------------------------------------------
Mon Sep 29 07:36:40 2025 Server listening on 50000 (test #4)
Mon Sep 29 07:36:40 2025 -----------------------------------------------------------
```

**Here is the JSON result with the timestamp option added.**
[server_log.json](https://github.com/user-attachments/files/22591518/server_log.json)


**And below is the log with `tickcount` added for further verification.**

```
Server listening on 50000 (test #6)
[2467345738] before select... test->mode=0, timeout=0.0
[2467361763] after select... res:1, test->mode=0, timeout=0.0
Accepted connection from 175.113.80.207, port 52890
[2467361765] before select... test->mode=-1, timeout=5.0
[2467361768] after select... res:1, test->mode=-1, timeout=4.997162
[  5] local 211.239.165.14 port 50000 connected to 175.113.80.207 port 52900
[2467361769] before select... test->mode=-1, timeout=5.0
[2467361769] after select... res:1, test->mode=-1, timeout=4.999996
[  8] local 211.239.165.14 port 50000 connected to 175.113.80.207 port 52904
[2467361771] before select... test->mode=-1, timeout=0.999885
[2467362772] after select... res:0, test->mode=-1, timeout=0.0
[2467362772] test->blocks_received:0, last_receive_blocks:0, t_usecs=17034235, rcv_timeout_us=5000000
iperf3: Server restart (#1) during active test due to idle timeout for receiving data
iperf3: error - idle timeout for receiving data
Server listening on 50000 (test #7)
```


**The server and client commands are as follows:**
* **Server:**
  ```bash
  iperf3 -s -p 50000 --rcv-timeout 5000 --timestamp -V
  ```
Setting rcv_timeout to 5000 was only to shorten the default value of 120 seconds in order to reproduce the issue more easily.


* **Client:**
  ```bash
  iperf3 -c xxx.xxx.xxx.xxx -p 50000 -t 3 -i 1 --bidir -V -b 100M
  ```


Thank you.

davidBar-On:
> 8.2 Set last_receive_time to now at the moment the server receives the measurement request from the client. → This way, the rcv_timeout is properly guaranteed after connection.

Great catch!  I fully agree with the analysis and the suggested fix approach (note that I am not from the iperf3 maintenance team).

The fix you suggest may be done by:
1. Move the [`last_receive_time` and `last_receive_blocks` initialization](https://github.com/scarpein/iperf/blob/a3706f0d37751eb5368058f579c902ce1113efdc/src/iperf_server_api.c#L587-L588) **after** the [test threads creation](https://github.com/scarpein/iperf/blob/a3706f0d37751eb5368058f579c902ce1113efdc/src/iperf_server_api.c#L915-L925).
2. Make the [`if (test->blocks_received == last_receive_blocks)`](https://github.com/scarpein/iperf/blob/a3706f0d37751eb5368058f579c902ce1113efdc/src/iperf_server_api.c#L677) test only when `test->state == TEST_RUNNING` (or at least when `test->state != IPERF_START`).

scarpein:
Updated the PR description:
- Removed unnecessary and off-scope content.
- Expanded the root-cause analysis to better isolate the core issue.

bmah888:
Thanks @scarpein and @davidBar-On for working on this and for your extremely well-written analysis and discussions! I'm trying to reproduce the original problem but I'm failing miserably, using tip-of-`master` iperf3 (3.19.1+).

Server side:
`iperf3 -s --rcv-timeout 5000 --timestamp -V
`

Client side:
`iperf3 --client 192.168.128.2 -t 3 -i 1 --bidir -V -b 100M
`

The tests seem to run fine, without any errors. These hosts are connected via a 100Gbps path.

Do I need to do anything else to perturb the test setup to make this failure happen?

davidBar-On:
> Do I need to do anything else to perturb the test setup to make this failure happen?

(Updated) I created a version where the client does **not** send anything, by adding  `sleep(100);` in `iperf_client_api.c` after:
https://github.com/esnet/iperf/blob/37b1d89a481c656a8e516a25f9b0656f9efb2fa2/src/iperf_client_api.c#L81-L82

When running this client several times one after the other (using larger `--rcv-timeout` and `-t` values can make it easier to test these scenarios):
1. If the client starts more then 5 secs (the server's `--rcv-timeout` value) after the last run (or after the server started) the failure happens.
2.  If the client starts less then 2 secs (5 minus 3, where 3 is the `-t` value) after the last run it will run successfully (but of course no data will be sent).
3. If the client starts 3 to 5 secs after the last run the failure will happen during the test.

scarpein:
> Do I need to do anything else to perturb the test setup to make this failure happen?

Hi, sorry for the late reply.
I actively use iperf3 in many real-world measurement scenarios, and it has been incredibly valuable for accurate network quality analysis. I truly appreciate the effort and dedication you put into maintaining this project and addressing even subtle issues—it has helped me a lot.

To share some additional context on why this issue is easily reproducible in my environment:

I typically run the iperf3 server in a cloud environment, and connect to it from clients over LTE or 5G mobile networks to measure bidirectional performance. In these scenarios, I frequently encounter the failure described in this PR.

From my testing, I found that two independent factors can contribute to the problem, either individually or together:

First, although bidirectional transmission starts almost simultaneously, depending on the path characteristics (latency, buffering/queuing, QoS/AQM, etc.) and traffic patterns, a heavy flow in one direction can temporarily delay or even starve the control/data flow in the opposite direction (client → server), even if the client is still active.
This behavior appears to be influenced more by traffic asymmetry, timing, and how the network path handles congestion than by the sheer amount of available bandwidth itself.

Second, mobile networks such as LTE/5G are inherently asymmetric and often have higher uplink latency or lower uplink capacity, which makes the receive side even more susceptible to delay or congestion.

Interestingly, I confirmed that:

Even with a cloud server and a wired (Ethernet) client, the issue can still occur in some paths.

And even when the server is not in the cloud, if the client connects over LTE/5G, the same failure can appear.

In other words, this is not limited to one specific type of network.
Instead, it seems to arise from the combination of:

the traffic pattern in bidirectional mode (server sending first / asymmetric flow), and

the characteristics of the network path (latency, buffering, or uplink/downlink imbalance).

Either factor alone can surface the issue, and when both are present, it becomes much easier to reproduce.

(Just my opinion / hypothesis based on real-world usage.)

Thank you again for your time and for maintaining such an essential tool.

bmah888:
Thanks @scarpein and @davidBar-On for the additional testing information, I'm now able to reproduce this in our local test environment, and I've confirmed that the fix addresses the problem (also, the logic looks correct from reading the diff).

PR #1949, opened 2025-09-27
Zerocopy Flag Does Not Send Expected Data

manedurphy:

```
```

* Issues fixed (if any):

Without the `--zerocopy` flag, the data sent on the wire is either random from `/dev/urandom` or a repeating pattern as implemented with the `--repeating-payload` flag. Since the `mmap` buffer is configured to `MAP_PRIVATE`, the data written to the buffer is not written to disk, so the `--zerocopy` send of data is 0 bytes.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -4727,7 +4727,7 @@ iperf_new_stream(struct iperf_test *test, int s, int sender)
         free(sp);
         return NULL;
     }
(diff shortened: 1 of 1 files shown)


swlars:
Thanks for the pull request! This looks fairly straightforward so we should get to it pretty quickly. How did test this pull request, with tcpdump?

swlars:
We tested using: 

`server% iperf3 --server

client% iperf3 --client server -n 1

client% sudo tcpdump -i en0 -X tcp port 5201`

And it seems to work! Thanks!

bmah888 (approved):
Thanks for the PR! This looks right and it appears (by packet inspection) to solve the problem.

Issue #1954, opened 2025-10-06
Get Client Output from server


ottbytel:
There is a way to get the client output on the server?

example: 
sudo iperf3 -s -p XXXX -J  --get-client-output

The goal is to retrieve the test results seen by the client on the server. (reverse the --get-server-output)

PR #1955, opened 2025-10-07
Support Server params to Client and get Server full-version info

davidBar-On:
master

* Issues fixed (if any): none specific

Add support for **sending server parameters to the client**.  This is done by the client informing the server that it supports getting its parameters, as part of sending the client's parameters (`"cap_server_params"`).  A server that support this feature will send to the client a new temporary state `SERVER_PARAM_EXCHANGE`, followed by the servers parameters.  After receiving the server parameters, the client set its state back to one before receiving `SERVER_PARAM_EXCHANGE`.

I considered to send the client capabilities (instead of `"cap_server_params"`) and future server capabilities as a 32 bits map parameter `"client_cababilities"` and `"client_cababilities"`, but I am not sure if bits map works o.k. between big/little endian platforms.

The trigger for adding this feature is issue #1954 that asks for a server option to receive that the client output.  If this PR will be merged it will be possible to suggest such enhancement.

To demonstrate the use this feature, the following two additional enhancements were added (may be split into separate PRs if needed):
1. The server **always send its version number** (`"server_version"`) to the client.  Client display the version in verbose mode (aslo add that the server display the client version in verbose mode).  Note that as I see it, server version number is just for information, and not for deciding which feature is supported by the server.  My approach is that future server capabilities will be added to the server parameters, and the client will be able to check which capabilities the server supports.
2. Send the **server version information**, same as displayed by `-v`, when requested by the client (`"get_server_version"`).  Client is asking for the version info per new `--get-server-version` option, display the version info and terminates.  If the option is not supported by the server, then the client notifies that.


davidBar-On:
@bmah888, @swlars, was it considered whether to merge this PR into the master branch?  (I have just rebased it with the latest master.)

The feature is fully backward compatible, and I believe that it is useful, even just to get server's version number, but it also opens the way for other enhancement.

swlars:
Hi David! Thanks for this pull request! This is something we've considered in the past, especially with some versions no longer being fully backwards compatible, so we are interested in this change. Thanks for bringing it to our attention, we'll hopefully take a closer look soon.

PR #1956, opened 2025-10-08
Resolve issue with openssl 3.5.3 encrypt

mdlowman:

```
```

* Issues fixed (if any):
1.  Setting the output buffer size fixes authentication on openssl 3.5.3.
2. While I was looking at the code, also fixed a small out-of-bounds access on base64 decoding.
3. Renamed keysize to output_buffer_len to be clearer.
4. Replaced some allocations of larger buffers; the larger allocations did not affect correctness, but made the code harder to follow.
5. Added a warning when data was silently truncated. Makes no difference to the caller, but makes it more obvious that there is no chunking of input material and that the plaintext is capped at the number of bits of the encryption.

Only the first commit is necessary to resolve the issue with openssl 3.5.3.
If you prefer to accept only a subset, or to consider them in separate pull requests, I can edit this/open others.

Set output buffer size prior to encrypt operation
When calling EVP_PKEY_encrypt with a non-null output buffer,
the output buffer length must be provided. Attempts to write
beyond this length will fail.

Rename keysize to output_buffer_len
This more accurately represents the meaning; it is the minimum
buffer allocation necessary for an encrypt or decrypt operation
to succeed. This is the same size for both ciphertext and
cleartext, as padding is applied.

Avoid out-of-bounds access when base64 decoding short strings
Check the length before reading memory.

Don't over-allocate followed by partial reads
We know how much we expect to read; the input buffer
has a defined size. Allocate the exact buffer expected
instead of a larger one with a read expected to return
only partial data. This makes it simpler to follow the
logic and to avoid off-by-one errors.

Add warnings on silent truncation
Input should not be this long, but makes the expectations
of the code clearer.

### DIFF
--- a/src/iperf_auth.c
+++ b/src/iperf_auth.c
@@ -251,6 +251,9 @@ int encrypt_rsa_message(const char *plaintext, EVP_PKEY *public_key, unsigned ch
     output_buffer_len = RSA_size(rsa);
 #endif
     plaintext_len = strlen(plaintext);
+    if (plaintext_len > output_buffer_len) {
+        fprintf(stderr, "Plaintext of size %zd truncated to %d; data is lost.\n", plaintext_len, output_buffer_len);
+    }
     rsa_buffer  = OPENSSL_malloc(output_buffer_len);
     *encryptedtext = (unsigned char*)OPENSSL_malloc(output_buffer_len);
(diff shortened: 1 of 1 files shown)


swlars:
Looks good! Thank you for the pull request!

PR #1960, opened 2025-10-30
Do not modify --bind and --client arguments inplace

klemensn:
Do as iperf_parse_hostname()'s comment says already:  pass a copy of
getopt(3)'s `optarg` aka. to avoid strtok(3) scribbling over `argv[]`.

Otherwise arguments like "fe80::1%vport0" appear as "fe80::1" in the
process list and cause exact matching of process name and arguments
(against what was used) to fail.

OpenBSD's net/iperf3 package ships a rc.subr(8) script and the service
framework uses pgrep(1) to check for running processes, where this
bug causes a mismatch due to the scope identifier being stripped:

```
	$ rcctl get iperf3 flags
	-6 --bind fe80::1%vport0
	$ rcctl check iperf3
	iperf3(failed)
	$ pgrep -fl iperf3
	33091 /usr/local/bin/iperf3 -s -D -6 -B fe80::1
```

Pass a copy to avoid modification, thus fixing rcctl(8) reporting:
```
	$ rcctl check iperf3
	iperf3(ok)
	$ pgrep -fl iperf3
	98863 /usr/local/bin/iperf3 -s -D -6 -B fe80::1%vport0
```

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1289,7 +1289,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                 }
 		iperf_set_test_role(test, 's');
                 break;
-            case 'c':
+            case 'c': {
                 if (test->role == 's') {
                     i_errno = IESERVCLIENT;
                     return -1;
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the pull request! I see and understand the problem you're trying to fix, iperf3 shouldn't go mucking around with argv[] like that. We'll take a look.

klemensn:
> Thanks for the pull request! I see and understand the problem you're trying to fix, iperf3 shouldn't go mucking around with argv[] like that. We'll take a look.

For archives, this has since been applied to OpenBSD's `iperf3` package via https://github.com/openbsd/ports/commit/bd56c81e6031956f7279aa68d3ddec2f624c4a1f.

src/iperf_api.c
```diff
@@ -1308,7 +1309,9 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
                     return -1;
```
bmah888: Do we need a `free(arg)` somewhere around here to prevent leaking the strdup-ed string in the case where we don't `HAVE_SO_BINDTODEVICE`?

src/iperf_api.c
```diff
@@ -1455,7 +1459,9 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
...
```
bmah888: Same question about `free(arg)` here?

PR #1987, opened 2025-12-19
Add warning about loose Ethernet connections causing iperf3 failures

bandejiyar:
Added a warning section to README.md explaining that iperf3 tests may fail if ethernet cables or intermediate connections are not fully seated. Loose or partially connected cables introduce impedance mismatches, causing signal reflections, bit errors, link instability, and throughput degradation during high‑speed testing.

  iperf-3.20

* Issues fixed (if any): 
added a warning for user to know that if iperf intermittently fail then it is possible that physical connections between iperf server and client may be loosened. This will save user's lot of time.

Only added a warning in readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Building iperf3
 
 ### Prerequisites: ###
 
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR! While you have a valid point, that loose cabling, etc. can make testing problematic, there's no way we could account for all possible conditions that would create packet loss, etc.

bandejiyar:
Sure. I realized this insight could have saved my nine months of debugging iperf, so just wanted to share it for the benefit of anyone who might use it in the future, but nvm.

Issue #1984, opened 2025-12-15
iperf3 client summary reports 0% UDP loss despite non-zero loss in server output


kindacute:
Whenever I introduce 30% UDP packet drop on my router, iperf3 always report 0% losses. I don't understand why it happens. I run command with --debug and --verbose parameters and see that losses about 30%:

`    "server_output_text": "-----------------------------------------------------------\nServer listening on 5201\n-----------------------------------------------------------\nAccepted connection from 194.37.101.163, port 49961\n[129] local 51.158.1.21 port 5201 connected to 194.37.101.163 port 50509\n[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams\n[129]   0.00-1.00   sec  86.3 KBytes   707 Kbits/sec  0.236 ms  22/83 (27%)  \n[129]   1.00-2.00   sec  89.1 KBytes   730 Kbits/sec  0.212 ms  23/86 (27%)  \n[129]   2.00-3.00   sec  89.1 KBytes   730 Kbits/sec  0.213 ms  24/87 (28%)  \n[129]   3.00-3.05   sec  1.41 KBytes   256 Kbits/sec  0.200 ms  0/1 (0%)  \n- - - - - - - - - - - - - - - - - - - - - - - - -\n[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams\n[129]   0.00-3.05   sec   266 KBytes   715 Kbits/sec  0.200 ms  69/257 (27%)  receiver\n",`

So why in the end (and without --verbose, --debug) I see 0 packet losses:
```

Reading new State from the Server - current state is 13-EXCHANGE_RESULTS
State change: client received and changed State to 14-DISPLAY_RESULTS
interval_len 0.997376 bytes_transferred 124528
interval forces keep
[  7]   2.01-3.00   sec   122 KBytes   999 Kbits/sec  86
- - - - - - - - - - - - - - - - - - - - - - - - -
Test Complete. Summary Results:
[ ID] Interval           Transfer     Bitrate         Jitter    Lost/Total Datagrams
[  7]   0.00-3.00   sec   366 KBytes   999 Kbits/sec  0.000 ms  0/259 (0%)  sender
[  7]   0.00-3.05   sec   266 KBytes   715 Kbits/sec  0.200 ms  0/257 (0%)  receiver
CPU Utilization: local/sender 0.9% (0.1%u/0.8%s), remote/receiver 0.1% (0.0%u/0.1%s)
[...]

PR #1988, opened 2025-12-20
Fix backward compatibility issue of client incorrectly print 0 UDP packets loss

davidBar-On:
master

* Issues fixed (if any): #1984

Fix backward compatibility issue of client showing UDP 0 packets loss, although packets were lost.  The issue is with older iperf3 servers that did not send omitted packets info.

The main issue was that for some reason, the omitted error count variable was not auto initialized to zero.  However, when I added a `printf()` of its value, it was initialized to 0.  Maybe a compiler issue, as the later initialization of the variable is done under `if` statement with no `else`.

For robustness, also ignored the omitted values if omitting time is not specified.  That allows showing the real packets loss when server does not support sending the omitting info.

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -2762,7 +2762,9 @@ get_results(struct iperf_test *test)
     cJSON *j_server_output;
     cJSON *j_start_time, *j_end_time;
     int sid;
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR! We'll review and comment.

Issue #1986, opened 2025-12-17
Server hangs endlessly (0.00 Mbits/sec) on sudden client disconnect – Works in 3.14, broken in 3.16+


Vili-B:
### iperf3 Version(s)
Working: 3.14

Intermittent: 3.15

Broken: 3.16 and newer

### Operating System / Architecture
Alma Linux 10

### Defect Description
I am encountering a regression in how iperf3 handles sudden client disconnections (specifically with mobile devices losing signal).

When a client device (mobile) abruptly loses connection during a test (e.g., 4G/5G drops or is toggled off), the server instance should detect the loss, report "Connection reset by peer," and return to the listening state.

In versions 3.16 and newer, the server fails to detect the disconnect. Instead of resetting, it hangs indefinitely, printing 0.00 Mbits/sec lines until manually killed.

### Regression / Bisect Info
I have tested multiple versions to isolate the behavior:

v3.14: Works reliably. The server detects the drop immediately, prints iperf3: error - unable to read from stream socket: Connection reset by peer, and waits for a new client.

v3.15: Unreliable. It works sometimes, but occasionally exhibits the hanging behavior.

v3.16+ (up to current): consistently broken. The server never resets and hangs on 0.00 Mbits/sec until the process is killed.

### Steps to Reproduce
Start an iperf3 server: iperf3 -s

Connect a client (a mobile device over 4G or 5G): iperf3 -c <server_ip> -t 100

Midway through the test, simulate a signal loss on the client (e.g., disable 4G/5G data or toggle airplane mode).

Observe Server Output:

Expected (v3.14): Server errors out with "Connection reset" and restarts the listener.

Actual (v3.16+): Server continues running, printing 0.00 Mbits/sec indefinitely.

davidBar-On:
@Vili-B, can you try running the server with `--rcv-timeout` option to see if it helps.  E.g. setting `--rcv-timeout 10` should stop the server about 10 seconds after the "signal loss".  Note that for backward compatibility the default is no receive timeout.

(Even if `--rcv-timeout` solve the problem, it is probably only a workaround, as I guess the server's receiving thread is terminated.  However, currently the main thread of iperf does not detect termination of sending/receiving threads.  E.g. see PR #1654.)

WatcherOfTheSkies:
Do you mean `--rcv-timeout`?

It's default is 120 sec, and range 100sec to 1 day!

davidBar-On:
[...]

PR #1990, opened 2025-12-27
Remove data sockets closing from server_timer_proc to avoid crash

davidBar-On:
master

* Issues fixed (if any): none

While evaluation Issue #1986 and simulating disconnect between the client and server, the **server crashed** after the call to `server_timer_proc()`.  Using gdb I found that the problem was that the `sp->test` address in the data thread was garbled.  I believe that this was because of race condition between the main and data threads:
1. `server_timer_proc()` closed the data stream and freed the `sp` buffers.
2. The `sp` buffer was used for something else, garbling the `test` address` (if this step happens later, depending on the threads race conditions, the crash will not happen).
3. Since the data stream was closed, the data thread `read()/recv()` returned 0 - end of file.
4. `iperf_tcp_recv()` in the data thread performs the `if (sp->test->state == TEST_RUNNING)` test that cause the server crash because the `sp->test` is garbled.

The suggested fix is that `server_timer_proc()` will only close the control socket, but not the data streams.  This causes the `select()` to fail and then the threads/sockets are closed in the usual way after failure.

Another approach was that `server_timer_proc()` will close the threads before closing the streams, but I think just closing the control channel is better.

Also, I noted that `server_timer_proc()` does not issue any error message.  I didn't add such message as I am not sure if this is not intentional.

### DIFF
--- a/src/iperf_server_api.c
+++ b/src/iperf_server_api.c
@@ -319,19 +319,11 @@ static void
 server_timer_proc(TimerClientData client_data, struct iperf_time *nowP)
 {
     struct iperf_test *test = client_data.p;
-    struct iperf_stream *sp;
 
     test->timer = NULL;
     if (test->done)
         return;
     test->done = 1;
-    /* Free streams */
-    while (!SLIST_EMPTY(&test->streams)) {
-        sp = SLIST_FIRST(&test->streams);
-        SLIST_REMOVE_HEAD(&test->streams, streams);
-        close(sp->socket);
-        iperf_free_stream(sp);
-    }
     close(test->ctrl_sck);
     test->ctrl_sck = -1;
 }


swlars:
Hi David! Thanks for the pull request. This looks like an important fix. How were you testing the original code to see the crash?

davidBar-On:
> How were you testing the original code to see the crash?

Hi, while trying to recreate the test, I found a more straightforward issue that is caused by the same problem and does not depend on race conditions. **Before the following lines add `sleep(45)`**:
https://github.com/esnet/iperf/blob/0ae94b62629a0e08cf641c0e2797b6adc6df9657/src/iperf_client_api.c#L789-L790
When running TCP test, after 40 seconds at that point (the test end), the server timeout and starts a new test, but **without terminating the data threads**.  After the client's sleep ends it terminates, which cause the data thread and the server to crash at:
https://github.com/esnet/iperf/blob/0ae94b62629a0e08cf641c0e2797b6adc6df9657/src/iperf_tcp.c#L75-L78
Again, because the `test` pointer is garbled.

This PR also fixes this problem (as it is caused by the same issue described in this PR).

swlars:
Thanks! That worked for me! Confirmed, good catch!

PR #2002, opened 2026-02-09
Fix UBSan with `-i0`: avoid division by zero

jackyyf:

```
```

* Issues fixed (if any):

```
Fix UBSan with `-i0`: avoid division by zero

The interval is explicity allowed to be zero, but we are trying to
divide by the interval without checking if it's zero, which will cause
an UBSan failure.

This commit fixed the issue by verify the interval is not zero before
division.
```

* Actual UBSan report:
```
iperf3/src/iperf_api.c:1962:6: runtime error: inf is outside the range of representable values of type 'int'
    #0 0x555b386f6125 in iperf_parse_arguments iperf3/src/iperf_api.c:1962:6
    #1 0x555b386ef30a in main iperf3/src/main.c:117:9
<snip>

SUMMARY: UndefinedBehaviorSanitizer: float-cast-overflow iperf3/src/iperf_api.c:1962:6
```

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1967,7 +1967,8 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
     /* Set Total-rate average interval to multiplicity of State interval */
     if (test->settings->bitrate_limit_interval != 0) {
 	test->settings->bitrate_limit_stats_per_interval =
-	    (test->settings->bitrate_limit_interval <= test->stats_interval ?
+	    (test->settings->bitrate_limit_interval <= test->stats_interval ||
+	     test->stats_interval == 0 ?
 	    1 : round(test->settings->bitrate_limit_interval/test->stats_interval) );
     }
 


jackyyf:
Sorry, forgot to include the command line used:

`iperf3 -clocalhost -i0`

and iperf3 will fail when compiled with `-fsanitize=float-cast-overflow`

bmah888:
Thanks for the PR!

Quick test note: Configure iperf3 and then do something like this to get the C compiler flags (looks like this on macOS):
```
bmah@dock2 iperf % grep CFLAGS Makefile
CFLAGS = -g -O2 -Wall -pthread
PTHREAD_CFLAGS = -pthread
```

Then build with an extra argument in CFLAGS like this:
```
bmah@dock2 iperf % make CFLAGS="-g -O2 -Wall -pthread -fsanitize=float-cast-overflow"
```

THen run iperf3 like so:
```
src/iperf3 -c localhost -i0
```

jackyyf:
Oops, super sorry for missing the comment, as the email slipped through my inbox queue. Please fix them in a later commit, sorry again for the incident

PR #2025, opened 2026-04-15
No delay of Client final interval report

davidBar-On:
master 3.21+

* Issues fixed (if any): #2019

Print final interval report from Client on `TEST_END` (like the Server) instead of `DISPLAY_RESULTS`.  This is to prevent the problem that the last Client's interval report is delayed on high-latency links (like cellular systems).


bmah888:
Thanks for the PR! We'll take a look at it.

PR #2028, opened 2026-04-20
Remove a stray carriage return from the service file

rpigott:


bmah888:
Thanks for the PR!

Just for my education (since I'm more of a FreeBSD guy) this change is cosmetic, with no functional impact, right?

rpigott:
`\r\n` is treated just like `\n`, so it's a cosmetic change, yeah. Just noticed the extra ^M when I opened this in my editor and thought it probably shouldn't be there..

bmah888 (approved):

Issue #2029, opened 2026-04-27
`--bidir` TCP test deadlocks with "idle timeout for receiving data" under load


AlexMoshkov:
## Summary

When running `iperf3` in bidirectional mode (`--bidir`) on a loaded host (e.g. many concurrent tests, virtualized environment), the test immediately deadlocks: both receiver streams report `0 bytes` from the first second, and the server eventually exits with:

```
error - idle timeout for receiving data
```

## Environment

- iperf3 version: 3.15
- OS: Linux 6.6 Ubuntu 20.04

### Test configuration

The bug was observed in the following setup:

- **48 virtual machines** (KVM/QEMU), each running **two concurrent bidirectional iperf3 tests** (96 total iperf3 processes)
- Each VM simultaneously acts as **both server and client**: every VM runs an iperf3 server and connects as a client to another VM, so all VMs are testing against each other.
- Each iperf3 is pinned to a single physical CPU core on VM (`-A 0` / `-A 1`)
- IPv6 connectivity between VMs
- Test duration: 120 seconds

```bash
# Each VM runs a server (accepting connections from other VMs)
iperf3 -s -1 --idle-timeout 10 -J -A 0 -p 10000

# Each VM also runs as a client connecting to another VM
iperf3 --bidir -6 -t 120 -c <peer_vm> -J -A 1 -p 10000
```

Reproduces reliably under load. On an idle host the race window is small and the bug is rare.

## Expected behavior

Each `iperf3 --bidir` test completes successfully, reporting non-zero throughput in both directions for the full test duration.

## Actual behavior

Under load, some tests deadlock immediately: both receiver streams report `0 bytes` transferred from the very first interval. The server eventually exits with:

```
error - idle timeout for receiving data
```

The client-side receiver stream also shows `0 bytes` throughout. No data is transferred in either direction for the entire test duration.

Here logs that show this behavior:

Server intervals
```Json
{
        "streams":	[{
                "socket":	5,
                "start":	0,
                "end":	1.003693,
                "seconds":	1.0036929845809937,
                "bytes":	0,
                "bits_per_second":	0,
                "omitted":	false,
                "sender":	false
            }, {
                "socket":	8,
                "start":	0,
                "end":	1.003701,
                "seconds":	1.0037009716033936,
                "bytes":	12615680,
                "bits_per_second":	100553295.1101696,
                "retransmits":	1,
                "snd_cwnd":	449820,
                "snd_wnd":	0,
                "rtt":	5106,
[...]

PR #2030, opened 2026-04-28
iperf_tcp: replace connect() with timeout_connect()

AlexMoshkov:
  `3.1-STABLE`) to which this pull request applies: **master**

* Issues fixed (if any): https://github.com/esnet/iperf/issues/2029

In --bidir mode, iperf_tcp_connect() was called twice: first for the sender stream, then for the receiver. The old connect() call returned immediately on EINPROGRESS, so both TCP handshakes completed in parallel.
Under load with RSS, the server's accept() could return them out of order, causing both sides to assign the same role (both sender or both receiver) and deadlock.

Replace connect() with timeout_connect() so each handshake fully completes before the next connect() is issued, ensuring deterministic accept() order.

### DIFF
--- a/src/iperf_tcp.c
+++ b/src/iperf_tcp.c
@@ -589,7 +589,8 @@ iperf_tcp_connect(struct iperf_test *test)
     /* Set common socket options */
     iperf_common_sockopts(test, s);
 
-    if (connect(s, (struct sockaddr *) server_res->ai_addr, server_res->ai_addrlen) < 0 && errno != EINPROGRESS) {
+    if (timeout_connect(s, (struct sockaddr *) server_res->ai_addr, server_res->ai_addrlen,
+                        DEFAULT_NO_MSG_RCVD_TIMEOUT) < 0 && errno != EINPROGRESS) {
 	saved_errno = errno;
 	close(s);
 	freeaddrinfo(server_res);


bmah888:
Thanks for the PR! We'll take a look at this...the behavior in #2029 is not something we've seen although we don't do a lot of `--bidir` tests in our part of ESnet.

bmah888 (commented):
The impact of this change is that a given iperf3 client process won't have more than a single TCP connection outstanding, so the iperf3 server process will need to `accept()` them one at a time in the same order as the `connect()` were done. This is because `timeout_connect()` with either return with a fully connected socket or an error, it will never return with a connection still in progress. This makes it impossible for them to mix up the order of connections getting accepted, thus won't mix up the roles in a `--bidir` test. Is that right?

(I haven't tried running with the change yet, and I'm not even sure how I would reproduce this in my test environment. Thanks for giving a lot of detail in your initial bug report...that's very helpful although I'm not sure how to build that setup. That is very very different from the intended environment where iperf3 was designed to run!)

AlexMoshkov:
Thanks for reply! Sorry for my long answer.

> Is that right?

Yes, it is

bmah888 (approved):
I don't quite have the means to easily test that this fixes the problem (which I've never seen so far first-hand) but I did compile and run this code with and without `--bidir`, and with various `--parallel` values and didn't observe any regression.

PR #2033, opened 2026-05-02
Limit UDP GSO number of segments in a message to the allowed value

### LINKED ISSUE #2032: GSO broken for a length value lower than 508
Iperf 3.21 cannot establish a connection with very short packets. It works properly when the length parameter is greater than 508.
~~~
abajkowski@abajkowski:~$ iperf3 -c 192.168.1.1 -R -t 3600 -u -b 3G -l 500 --gsro
Connecting to host 192.168.1.1, port 5201
Reverse mode, remote host 192.168.1.1 is sending
[  5] local 192.168.1.10 port 55486 connected to 192.168.1.1 port 5201
[  5]   0.00-1.00   sec  0.00 Bytes  0.00 bits/sec  0.000 ms  0/0 (0%)  
^C[  5]   1.00-1.77   sec  0.00 Bytes  0.00 bits/sec  0.000 ms  0/0 (0%)  
[  5]   0.00-1.77   sec  0.00 Bytes  0.00 bits/sec  0.000 ms  0/0 (0%)  sender
[  5]   0.00-1.77   sec  0.00 Bytes  0.00 bits/sec  0.000 ms  0/0 (0%)  receiver
iperf3: interrupt - the client has terminated by signal Interrupt(2)
~~~

davidBar-On:
master 3.21+

* Issues fixed (if any): #2032

Suggested fix to PR #1925 to limit the number of segments in UDP GSO message to 128, which is the maximum allowed.  The problem was with messages length 507 or less, as 507 is the maximum length where `MAX_UDP_BLOCKSIZE/length >= 129`.

The limit `GSO_MAX_DG_IN_BF` is defined as a constant as I didn't find a reliable way to determine it dynamically.  From what I found, the value is defined in Linux as `UDP_MAX_SEGMENTS`.  Although it seems that in some (newer?) Linux distributions it is defined in `udp.h`, as in [here](https://github.com/torvalds/linux/blob/f1a5e78a55ebf2b05777fd5eb738038ddae609d6/include/linux/udp.h#L127), at least in WSL Linux that I use it is not defined in a header file, and probably it is defined in `udpgso.c` like in [here](https://github.com/torvalds/linux/blob/6fe0be6dc7faf984599b1e356ead1c49b64ed3ca/tools/testing/selftests/net/udpgso.c#L36-L38).

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -494,6 +494,7 @@ extern int gerror; /* error value from getaddrinfo(3), for use in internal error
 /* In Reverse mode, maximum number of packets to wait for "accept" response - to handle out of order packets */
 #define MAX_REVERSE_OUT_OF_ORDER_PACKETS 2
 
+#define GSO_MAX_DG_IN_BF (1 << 7UL) // 128 - the Linux Kernel limit hardcoded as `UDP_MAX_SEGMENTS (1 << 7UL)` in `udpgso.c`
 #define GSO_BF_MAX_SIZE MAX_UDP_BLOCKSIZE
 #define GRO_BF_MAX_SIZE MAX_UDP_BLOCKSIZE
 
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -420,7 +420,7 @@ iperf_handle_message_client(struct iperf_test *test)
 int
 iperf_connect(struct iperf_test *test)
 {
-    int opt;
+    int opt, n;
     socklen_t len;
 
     if (NULL == test)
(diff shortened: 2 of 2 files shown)


abajk:
Now it works fine even with the minimum UDP packet size :D

bmah888:
Thanks for the PR! This looks pretty reasonable (I haven't done any testing).

bmah888 (approved):
I was able to successfully reproduce the original (broken) behavior of #2032 without the patch, and verified that data actually gets sent and recorded with the patch. Looks good to me!

Issue #2037, opened 2026-05-20
Separate UDP GSO and GRO controls for Linux (--gso and --gro options)


roaoliver:
# [Enhancement] Add independent control for UDP GSO and GRO (--gso, --gro)

## Problem Statement
Currently, `iperf3` supports UDP Generic Segmentation Offload (GSO) and Generic Receive Offload (GRO) via a single combined flag: `--gsro`. While this is convenient, it forces both offloads to be active simultaneously. 

Testing shows that while GSO (sender-side) is almost universally beneficial, GRO (receiver-side) can cause significant CPU saturation on the receiver depending on the hardware and kernel version. Currently, there is no way to benefit from GSO without incurring the receiver-side overhead of GRO.

## Proposed Solution
I propose splitting the offload controls into two independent flags while maintaining backward compatibility:

1.  `--gso`: Enables only `UDP_SEGMENT` on the sender.
2.  `--gro`: Enables only `UDP_GRO` on the receiver.
3.  `--gsro`: Remains as a legacy/convenience flag that enables both.

## Technical Motivation & Test Evidence
Testing across different Network Interface Cards (NICs) reveals that a "one-size-fits-all" approach to offloading is suboptimal.

### Test Environments
*   **Env 1:** Intel X552 (10GbE) | Ubuntu 22.04 (Kernel 6.8)
*   **Env 2:** Broadcom BCM57800 (10GbE) | Ubuntu 22.04 (Kernel 5.15)
*   **Env 3:** Mellanox ConnectX-5 (25GbE+) | Ubuntu 24.04 (Kernel 6.17)

### Comparative Data
| Setup | Mode | Sender CPU | Receiver CPU | Notes |
| :--- | :--- | :--- | :--- | :--- |
| **Intel** | Baseline | 78.3% | 43.1% | |
| **Intel** | `--gsro` | 32.1% | **100.0%** | Receiver saturated by GRO |
| **Intel** | **`--gso`** | 32.0% | **41.4%** | Optimal efficiency |
| **Broadcom**| Baseline | 99.9% | 100.0% | 9.92 Gbps |
| **Broadcom**| `--gsro` | 70.6% | 99.5% | 9.92 Gbps |
| **Broadcom**| **`--gso`** | 72.1% | 98.6% | 9.92 Gbps (Stable) |
| **Mellanox**| Baseline | 99.8% | 76.8% | Bitrate stuck at 18.7G |
[...]

PR #2038, opened 2026-05-20
Separate UDP GSO and GRO controls for Linux (--gso and --gro options) #2037

roaoliver:

```
```

3.21
* Issues fixed (if any):
#2037 
Separate UDP GSO and GRO controls for Linux (--gso and --gro options)


src/iperf_api.c
```diff
@@ -1803,6 +1807,20 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 		test->settings->gso = 1;
 		test->settings->gro = 1;
                 break;
+      case OPT_GSO:
+    /* Enable GSO which is disabled by default */
+    /* Flag is available regardless of local support to allow client to request server to use it */
+    gso_flag = 1;
+    test->settings->gso = 1;
+    test->settings->gro = 0;
```
davidBar-On: Suggest to set only `gso`, so `--gso --gro` will be equivalent to `--gsro`.

src/iperf_api.c
```diff
@@ -1803,6 +1807,20 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
...
+    break;
+      case OPT_GRO:
+    /* Enable GRO which is disabled by default */
+    /* Flag is available regardless of local support to allow client to request server to use it */
+    gro_flag = 1;
+    test->settings->gso = 0;
+    test->settings->gro = 1;
```
davidBar-On: Suggest to set only `gro`, so `--gso --gro` will be equivalent to `--gsro`.

davidBar-On:
To reduce number of iperf3 options, instead of adding `--gso` and `--gro` it may be better to add a flag to `--gsro`. E.g `--gsro/s` will set only gso and `--gsro/r` will set only gro.

roaoliver:
Suggestion implemented in the code.

image

biergaizi:
For your information, the high CPU usage was caused by busy-polling via `MSG_DONTWAIT`, this has been root-caused by @kvasana in #2069, once #2069 is merged this workaround is no longer necessary in your use case (high CPU usage at receiver).

That being said, I still support supporting separate GSO and GRO options. In some test setups the server and client processing speeds are not symmetric, so it's still useful to isolate a sender or receiver side bottleneck.

PR #2040, opened 2026-05-22
iperf_auth: check BIO_read return value in Base64Decode

dxbjavid:
On malformed base64 input (e.g. the 2-char string "AB"), BIO_read in
Base64Decode returns -1, which is assigned to the size_t out-parameter
*length and silently becomes SIZE_MAX while the function still returns
0. Callers in iperf_auth.c (decode_auth_setting, load_pubkey_from_base64)
then pass that huge length downstream. Check the BIO_read result, free
the buffer and report failure if it is negative.


bmah888:
Thanks for the PR! We'll take a look at it.

bmah888 (approved):
This looks good, approved. I want to do a little follow-up work because the callers of `Base64Decode()` aren't real good at checking their return values, but this change by itself looks good to me. Thanks!

PR #2041, opened 2026-05-24
iperf_client_api: align default TCP block size to MSS

nshopik:
When the user has not specified -l, round the default 128 KiB block size down to the largest multiple of the MSS so each per-write chunk decomposes into integer-many full-sized wire segments after TSO/GSO segmentation, eliminating the short trailing segment per chunk.  When -M is set, align to settings->mss (the requested value); otherwise align to the control-socket MSS already read in iperf_connect().  Mirrors the pre-existing UDP block-size auto-sizing in the same function.

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf.h
+++ b/src/iperf.h
@@ -162,6 +162,7 @@ struct iperf_settings
     int       domain;               /* AF_INET or AF_INET6 */
     int       socket_bufsize;       /* window size for TCP */
     int       blksize;              /* size of read/writes (-l) */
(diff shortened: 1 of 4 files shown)


nshopik:
Tested manually on Debian with kernel `6.12.48+deb13-amd64` to confirm correct behavior and there no short tail packets.

bmah888:
Thanks for the pull request! I see the benefits of this behavior although it's a definite change from the prior default. We'll take a look.

Issue #1815, opened 2024-12-21
slow sctp performance on arm64 


nshopik:
# Context

* Version of iperf3: 3.12

* Hardware: arm ampere 2 cores with 4gb memory

* Operating system (and distribution, if any): `Linux 6.1.0-27-arm64 #1 SMP Debian 6.1.115-1 (2024-11-01) aarch64`

# Bug Report

* Expected Behavior

somewhat similar what we see on x86

* Actual Behavior
sctp test
```
iperf 3.12
Linux goro 6.1.0-27-arm64 #1 SMP Debian 6.1.115-1 (2024-11-01) aarch64
Control connection MSS 32768
Time: Sat, 21 Dec 2024 20:44:37 GMT
Connecting to host 127.0.0.1, port 5201
      Cookie: rp56igphb6uclgiy73ijfa2ymnn6a465xotj
[  5] local 127.0.0.1 port 32984 connected to 127.0.0.1 port 5201
Starting Test: protocol: SCTP, 1 streams, 65536 byte blocks, omitting 0 seconds, 10 second test, tos 0
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  6.62 MBytes  55.6 Mbits/sec
[  5]   1.00-2.00   sec   448 KBytes  3.67 Mbits/sec
[  5]   2.00-3.00   sec   640 KBytes  5.24 Mbits/sec
[  5]   3.00-4.00   sec   576 KBytes  4.72 Mbits/sec
[  5]   4.00-5.00   sec   768 KBytes  6.29 Mbits/sec
[  5]   5.00-6.00   sec   384 KBytes  3.15 Mbits/sec
[  5]   6.00-7.00   sec   576 KBytes  4.72 Mbits/sec
[  5]   7.00-8.00   sec   448 KBytes  3.67 Mbits/sec
[  5]   8.00-9.00   sec  2.00 MBytes  16.8 Mbits/sec
[  5]   9.00-10.00  sec  1.56 MBytes  13.1 Mbits/sec
- - - - - - - - - - - - - - - - - - - - - - - - -
Test Complete. Summary Results:
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-10.00  sec  13.9 MBytes  11.7 Mbits/sec                  sender
[  5]   0.00-10.00  sec  13.9 MBytes  11.6 Mbits/sec                  receiver
[...]

PR #2042, opened 2026-05-24
iperf_sctp: disable delayed-SACK by default to fix slow single-flow SCTP throughput

nshopik:
Single-flow SCTP throughput collapses to a few Mbits/sec on Linux (~300× slower than TCP on the same loopback) because  the receiver's delayed-SACK timer (sack_delay=200 ms, sack_freq=2) waits for a second DATA chunk that the blocked sender will not produce — each iperf3 64 KB block lands as one DATA chunk and stalls 200 ms per I/O cycle.  Reproducible on both arm64 and x86_64; not architecture-specific.

iperf3 is a throughput benchmark, so disable delayed-SACK by setting SCTP_DELAYED_SACK with sack_delay=0, sack_freq=1
  on all three SCTP socket entry points:

  - iperf_sctp_listen — endpoint default (FreeBSD-style SCTP_FUTURE_ASSOC)
  - iperf_sctp_accept — required on Linux; the listener-side setting does not propagate to accepted associations
  - iperf_sctp_connect — covers --reverse where the client becomes the receiver

Also propagate SCTP_NODELAY to both the server's listen and accepted sockets when -N is given, closing a pre-existing  asymmetry with iperf_tcp_listen's handling of TCP_NODELAY.

configure gains an AC_CHECK_TYPES probe for struct sctp_sack_info; new code is guarded by HAVE_STRUCT_SCTP_SACK_INFO and ENOPROTOOPT is treated as non-fatal, so platforms without support are silently unaffected.

master and fix performance
```
SCTP iperf3 -c localhost --sctp -t 10 │ 2.67 Mbits/sec │ 1.42 Gbits/sec
```

```
```

* Issues fixed (if any): #1815 

* Brief description of code changes (suitable for use as a commit message): disable delayed-SACK by default to fix slow single-flow SCTP throughput (#1815)

### DIFF
--- a/configure
+++ b/configure
@@ -16897,6 +16897,16 @@ printf '%s\n' "#define HAVE_STRUCT_SCTP_ASSOC_VALUE 1" >>confdefs.h
 
 fi
 
+ac_fn_c_check_type "$LINENO" "struct sctp_sack_info" "ac_cv_type_struct_sctp_sack_info" "#include <netinet/sctp.h>
+"
+if test "x$ac_cv_type_struct_sctp_sack_info" = xyes
+then :
+
+printf '%s\n' "#define HAVE_STRUCT_SCTP_SACK_INFO 1" >>confdefs.h
+
+
+fi
+
 fi
 
 done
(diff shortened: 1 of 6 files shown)


bmah888:
Thanks for the PR!

PR #2043, opened 2026-05-24
iperf_sctp: apply default socket buffer to fix slow single-CPU throughput Linux SCTP

nshopik:
Unlike TCP, it does not auto-tune SO_RCVBUF/SO_SNDBUF (there is no equivalent of tcp_moderate_rcvbuf).  Without an explicit setting the per-association receive window collapses to ~40 KB, which on single-CPU systems gates single-stream loopback throughput to roughly 1 Gbps even after the delayed-SACK fix from #1815: sender and softirq compete for the same core, so the small rwnd is never reopened in time.

Apply DEFAULT_SCTP_SOCKBUF (512 KB) to SO_RCVBUF and SO_SNDBUF on the client connect socket, the server listen socket, and the per- association accept socket whenever --window is not given.  The kernel silently caps at net.core.{r,w}mem_max where smaller, and the explicit --window path is unchanged.  Failures on the default path are non-fatal (best-effort, not user-requested).

Measured on Debian 13 (kernel 6.12.48), loopback, single stream, 30 s runs, median of 3:

    1-vCPU box
      TCP default          : 25.5 Gbps (unchanged)
      SCTP default before  :  0.94 Gbps
      SCTP default after   : 25.5 Gbps  (~27x)

    2-vCPU box
      SCTP default before/after: ~22 Gbps (within run-to-run variance)

The gap only manifests when sender and softirq share a CPU; on multi- core systems the kernel already works around the small rwnd and the patch is a no-op there.

```
```

* Issues fixed (if any):

### DIFF
--- a/src/iperf_api.h
+++ b/src/iperf_api.h
@@ -66,6 +66,7 @@ typedef atomic_uint_fast64_t atomic_iperf_size_t;
 #define DEFAULT_UDP_BLKSIZE 1460 /* default is dynamically set, else this */
 #define DEFAULT_TCP_BLKSIZE (128 * 1024)  /* default read/write block size */
 #define DEFAULT_SCTP_BLKSIZE (64 * 1024)
+#define DEFAULT_SCTP_SOCKBUF (512 * 1024) /* Linux SCTP has no rcvbuf auto-tune */
 #define DEFAULT_PACING_TIMER 1000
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the PR! Will test this.

PR #2045, opened 2026-06-03
Allow --zerocopy and --skip-rx-copy when only server supports them

### LINKED ISSUE #2044: zerocopy os check should wait for sever response when in reverse mode
According to davidBar-On the zerocopy option is forwarded to the server in reverse mode, but on windows, iperf permaturely says "this os does not support sendfile" even if the remote end does support send file 

```
iperf3 -c aura -R -Z
iperf3: parameter error - this OS does not support sendfile
```

this is wrong because the server is linux and is fully capable of using sendfile

> > Is the `--zerocopy` option forwarded to the server
> 
> `--zerocopy` option **is** forwarded to the server- see PR [#1204](https://github.com/esnet/iperf/pull/1204).

davidBar-On:
master 3.21+

* Issues fixed (if any): #2044

Allow `--zerocopy` and `--skip-rx-copy` even when one of the sides OS, client or server, not support them, as they may be supported by the other side.  When not supported, allow `zerocopy` only when not sending and `skip-rx-copy` only when not receiving.

(UPDATE note: deleted the `-P` related changes and created a separate PR for them.)

### DIFF
--- a/src/iperf_api.c
+++ b/src/iperf_api.c
@@ -1170,9 +1170,7 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 #if defined(HAVE_DONT_FRAGMENT)
 	{"dont-fragment", no_argument, NULL, OPT_DONT_FRAGMENT},
 #endif /* HAVE_DONT_FRAGMENT */
-#if defined(HAVE_MSG_TRUNC)
 	{"skip-rx-copy", no_argument, NULL, OPT_SKIP_RX_COPY},
-#endif /* HAVE_MSG_TRUNC */
 #if defined(HAVE_SSL)
     {"username", required_argument, NULL, OPT_CLIENT_USERNAME},
     {"rsa-public-key-path", required_argument, NULL, OPT_CLIENT_RSA_PUBLIC_KEY},
@@ -1557,10 +1555,6 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
 		TAILQ_INSERT_TAIL(&test->xbind_addrs, xbe, link);
                 break;
             case 'Z':
-                if (!has_sendfile()) {
-                    i_errno = IENOSENDFILE;
-                    return -1;
-                }
                 test->zerocopy = 1;
 		client_flag = 1;
                 break;
@@ -1776,12 +1770,10 @@ iperf_parse_arguments(struct iperf_test *test, int argc, char **argv)
             server_flag = 1;
 	    break;
 #endif /* HAVE_SSL */
-#if defined(HAVE_MSG_TRUNC)
-            case OPT_SKIP_RX_COPY:
+            case OPT_SKIP_RX_COPY: /* Allow the option even when `HAVE_MSG_TRUNC` is not defined, as server may support it */
                 test->settings->skip_rx_copy = 1;
                 client_flag = 1;
                 break;
-#endif /* HAVE_MSG_TRUNC */
 	    case OPT_PACING_TIMER:
 		test->settings->pacing_timer = unit_atoi(optarg);
 		if (i_errno != 0) {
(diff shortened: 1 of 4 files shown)


bmah888:
Thanks for the pull request. I'm trying to test it, but it looks like this is a fix (with respect to `--zerocopy` and `--skip-rx-copy` for a problem that does _not_ occur on Linux, FreeBSD, or macOS (the three main development problems). If this is only an issues (a platfom that we don't officially support) this is going to be a little hard for our team to test. Haven't quiet figured out how to deal with this yet.

davidBar-On:
> Haven't quiet figured out how to deal with this yet.

I tested it by creating a version with `HAVE_MSG_TRUNC` set to zero, and then run it once as a client and once as a server (when the other side is the version with `HAVE_MSG_TRUNC` set to 1).  I believe this is sufficient as each of these options affect either sending or sides but not both.

src/iperf_api.c
```diff
@@ -2842,7 +2842,23 @@ get_parameters(struct iperf_test *test)
             i_errno = IETOTALRATE;
             r = -1;
         }
-    }
+
+        /* Ensure a sending server supports zerocpy */
+        if (test->zerocopy == 1 && test->mode != RECEIVER && !has_sendfile()) {
+            i_errno = IENOSENDFILE; // Zerocopy required for the Client but is not supported
```
bmah888:
```suggestion
            i_errno = IENOSENDFILE; // Zerocopy required for the Server but is not supported
```

src/iperf_error.c
```diff
@@ -563,7 +563,10 @@ iperf_strerror(int int_errno)
             snprintf(errstr, len, "server test duration expired");
             perr = 1;
             break;
-	    default:
+        case IERVRSONLYSKIPRXCOPY:
+            snprintf(errstr, len, "this OS does not support Skip RX Copy");
```
bmah888:
```suggestion
            snprintf(errstr, len, "this OS does not support --skip-rx-copy");
```

src/iperf_locale.c
```diff
@@ -212,14 +212,17 @@ const char usage_longstr[] = "Usage: iperf3 [-s|-c host] [options]\n"
 #endif /* HAVE_FLOWLABEL */
                            "  -Z, --zerocopy            use a 'zero copy' method of sending data\n"
 #if defined(HAVE_MSG_TRUNC)
-                           "  --skip-rx-copy            ignore received messages using MSG_TRUNC option\n"
+#define IPERF_HELP_MSG_TRUNK_SUPPORTED_BY_THIS_OS ""
+#else /* !HAVE_MSG_TRUNC */
+#define IPERF_HELP_MSG_TRUNK_SUPPORTED_BY_THIS_OS " (this OS does not support but server OS may)"
 #endif /* HAVE_MSG_TRUNC */
+                           "  --skip-rx-copy            ignore received messages using MSG_TRUNC option" IPERF_HELP_MSG_TRUNK_SUPPORTED_BY_THIS_OS "\n"
```
bmah888: I'm opposed to making the help text more complicated in this way. I'd almost rather just emit the existing `--skip-rx-copy` text all the time.

bmah888 (commented):
Thanks for the PR. I have a concern that this adds argument parsing complexity basically to cater to the case of a Windows client connecting to a Linux (or other UNIX-like) host, and Windows isn't even an officially supported platform. This isn't even applicable to any of the supported platforms (Linux, FreeBSD, macOS), but we have to pay the cost of more complicated code that's harder to test and maintain.

src/iperf_api.c
davidBar-On: Changed

src/iperf_error.c
davidBar-On: Changed

src/iperf_locale.c
davidBar-On: Modified the change - removed the `#define IPERF_HELP_MSG_TRUNK_SUPPORTED_BY_THIS_OS` but still added a specific `--skip-rx-copy` help version for when not supported by the local iperf3 but may be supported by the remote end.

bmah888:
Thanks for the revisions! Taking another look.

bmah888 (approved):
Thanks for working on this! I was able to test this as you suggested by forcing some of the definitions in `src/iperf_config.h`. All seems to be working as intended. Merging shortly.

PR #2046, opened 2026-06-05
Explicitly check the return values from Base64Decode()

bmah888:
Follow-on to PR #2040.


swlars (approved):
Looks good! Thanks for cleaning that up!

PR #1985, opened 2025-12-17
Add option to set the ToS of the control connection.

liam-mcb:
Following on from the discussion in https://github.com/esnet/iperf/discussions/1982.

Added options to set the ToS/DSCP of the control connection.
This helps prevent the control connection from dropping when testing low-capacity networks.
It was suggested in the discussion thread that this option should be added to the existing `--tos` option with a slash separator.
I decided against that because the control ToS option is available to both the client and server.

Issue #830, opened 2018-12-20
iperf_udp.c issue of tos not set for the first test packet


velamen2009:
# Context

* Version of iperf3: 3.6+

* Operating system (and distribution, if any): tested on Ubuntu, MAC

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report
At iperf_udp.c, line 431, client sends the first test packet, the tos is always 0x00. 

* Expected Behavior
The first test packet shall use the same tos as user's input.

Reason:  Many vendors' scheduler only check the DSCP marking once when receive the first packet. If the user wants to send high QoS service such as 0xc0,  since the first packet is the test packet with tos 0x00, scheduler will treat the service as 0x00, best effort, QoS will not work in this case.

* Possible Solution
Send the first test packet with the same tos as user's setting

bmah888:
Is this still a problem?

davidBar-On:
> Is this still a problem?

DSCP is still not set in the first UDP streams response from the server.  Submitted PR #2047 with a suggested resolution.

Issue #510, opened 2017-02-08
DSCP bit appearing in different frame in iperf2 nad iperf3-current


purumishra:
Hi ,
   I am using DSCP parameter  of iperf for udp packets. I used both iperf2 and iperf3 latset version ie.iperf 3-CURRENT.
   
  Iperf2 sets it n 3rd frame while latest iperf(iperf 3-CURRENT) sets the same in 5th frame.

a. wireshark snippet for  iperf 3-CURRENT:

**_Frame 5:_** 342 bytes on wire (2736 bits), 342 bytes captured (2736 bits)
Ethernet II, Src: fa:16:3e:2c:7c:9a (fa:16:3e:2c:7c:9a), Dst: AristaNe_00:00:a0 (00:1c:73:00:00:a0)
Internet Protocol Version 4, Src: 96.118.224.85, Dst: 96.116.32.162
    0100 .... = Version: 4
    .... 0101 = Header Length: 20 bytes
    **_Differentiated Services Field: 0xb8 (DSCP: EF_** PHB, ECN: Not-ECT)
        1011 10.. = Differentiated Services Codepoint: Expedited Forwarding (46)
        .... ..00 = Explicit Congestion Notification: Not ECN-Capable Transport (0)

b. wireshark trace snippet for iperf2:

**_## Frame 3:_** 342 bytes on wire (2736 bits), 342 bytes captured (2736 bits)
Ethernet II, Src: fa:16:3e:1d:36:fd (fa:16:3e:1d:36:fd), Dst: AristaNe_10:64:4c (00:1c:73:10:64:4c)
Internet Protocol Version 4, Src: 96.116.32.162, Dst: 96.118.224.85
    0100 .... = Version: 4
    .... 0101 = Header Length: 20 bytes
    ## Differentiated Services Field: 0xb8 (DSCP: EF PHB, ECN: Not-ECT)
        1011 10.. = Differentiated Services Codepoint: Expedited Forwarding (46)
        .... ..00 = Explicit Congestion Notification: Not ECN-Capable Transport (0)

So cant we get DSCP bit in the third frame of iperf data in new iperf3 -Current.

Thanks
Puru

bmah888:
Hi, is this still a problem?

davidBar-On:
> Hi, is this still a problem?

DSCP is still not set in the first streams packets from the server: first UDP response and the first two TCP messages (ACKs).  Submitted PR #2047 with a suggested resolution.

PR #2047, opened 2026-06-10
Ensure DSCP/ToS is set in the first UDP/TCP streams packets

davidBar-On:
master 3.21+

* Issues fixed (if any): #830, #510

Suggested fix to ensure that DSCP/ToS is set in the first streams packets (with DSCP option help message enhancement).
Currently, DSCP is not set in the first stream packets from the server: first UDP response and the first two TCP messages (ACKs).

(Note this PR is not related to the control channel - PR #1985 suggests adding an option to set its DSCP/ToS.)


bmah888:
Thanks for the PR! We'll take a look.

bmah888 (approved):
Thanks, this looks good and works well, approved. I'll note this changes also that improves some error handling and adds a more specific error message for a bad DSCP value.

Issue #2051, opened 2026-06-24
JSON output (-J) emits two top-level "error" keys on server-relayed errors


therealevanhenry:
# Context

* Version of iperf3:
3.21 (cJSON 1.7.15), git `d39cf41526626b4e5a130f115d931cd6cbdffc19`. The relevant code paths are unchanged on current `master`, so earlier 3.x releases are very likely affected as well.

* Hardware:
x86_64 (immaterial — this is in the JSON-assembly logic, not platform-specific).

* Operating system (and distribution, if any):
Linux. Platform-independent.

# Bug Report

* Expected Behavior

With `-J`, an error that is relayed from the server should produce a single top-level `"error"` member in the JSON document.

* Actual Behavior

When the client receives a `SERVER_ERROR` control message, the `-J` document contains **two** `"error"` members in the top-level object — one with a `SERVER ERROR - ` prefix, one without:

```json
	"intervals":	[],
	"end":	{
	},
	"error":	"SERVER ERROR - client's requested duration exceeds the server's maximum permitted limit",
	"error":	"client's requested duration exceeds the server's maximum permitted limit"
```

RFC 8259 §4 permits duplicate names but states that in that case "the behavior of software that receives such an object is unpredictable." In practice the result is parser-dependent: last-wins parsers (Python `json`, JS `JSON.parse`, Go, Rust `serde_json`) surface only the bare message and silently drop the `SERVER ERROR - ` prefix; a first-wins reader (including cJSON's own `cJSON_GetObjectItem`) keeps the prefixed one. A consumer can't reliably tell which `"error"` string it will get.

* Steps to Reproduce

```sh
# terminal 1 — server that refuses tests longer than 2s
iperf3 -s -1 --server-max-duration 2 -p 5301

# terminal 2 — client requests a 10s test, which exceeds the limit
iperf3 -c 127.0.0.1 -p 5301 -t 10 -J
```

The server relays `SERVER_ERROR` (`IEMAXSERVERTESTDURATIONEXCEEDED`) and the client's `-J` document ends with the two `"error"` members shown above.

* Root Cause

Two independent code paths each add `"error"` to the same `test->json_top`, and `cJSON_AddStringToObject()` *appends* a node rather than replacing an existing key:

[...]

PR #2053, opened 2026-07-02
Only one Client error log about Server error

davidBar-On:
master 3.21+

* Issues fixed (if any): #2051

Suggested fix to the issue that the client logs two error messages when receiving an error from the server, which may be a problem in JSON output.

The suggested solution is that the `SERVER ERROR` prefix to the error text is added by `iperf_strerror()`, instead of logging a special error message (which causes the duplicate log).  This is by **negating** the error code received from the server, which indicates to `iperf_strerror()` that this is a server error.  In addition, handling of the `errno` received from the server was also added to `iperf_strerror()`

### DIFF
--- a/src/iperf_client_api.c
+++ b/src/iperf_client_api.c
@@ -394,17 +394,12 @@ iperf_handle_message_client(struct iperf_test *test)
                 i_errno = IECTRLREAD;
                 return -1;
             }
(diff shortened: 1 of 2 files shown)


bmah888:
Thanks for the PR, acknowledged!

Issue #836, opened 2019-01-10
Upstream receiver end intervals reporting wrong seconds value


FuzzyStatic:
_NOTE: The iperf3 issue tracker is for registering bugs, enhancement
requests, or submissions of code.  It is not a means for asking
questions about building or using iperf3.  Those are best directed
towards the iperf3 mailing list at iperf-dev@google-groups.com or
question sites such as Stack Overflow
(http://www.stackoverflow.com/).  A list of frequently-asked questions
regarding iperf3 can be found at http://software.es.net/iperf/faq.html._

# Context

* Version of iperf3:
3.6

* Hardware:
MacBookPro

* Operating system (and distribution, if any):
macOS

_Please note: iperf3 is supported on Linux, FreeBSD, and macOS.
Support may be provided on a best-effort basis to other UNIX-like
platforms.  We cannot provide support for building and/or running
iperf3 on Windows, iOS, or Android._

* Other relevant information (for example, non-default compilers,
  libraries, cross-compiling, etc.):

_Please fill out one of the "Bug Report" or "Enhancement Request"
sections, as appropriate._

# Bug Report

* Expected Behavior

The end streams receiver seconds should be receiver end time - receiver start time.

* Actual Behavior

It looks like the receiver seconds is reporting the sender's seconds. This bug does not look like it impacts results.

* Steps to Reproduce

Run an upstream test and look at the end stream JSON.

```
"end":  {
                "streams":      [{
                                "sender":       {
                                        "socket":       8,
                                        "start":        0,
                                        "end":  8.00232195854187,
                                        "seconds":      **8.00232195854187**,
                                        "bytes":        17896104,
                                        "bits_per_second":      17890911.255723491
                                },
                                "receiver":     {
                                        "socket":       8,
                                        "start":        0,
                                        "end":  8.113577127456665, // <-- this should be the seconds for receiver
                                        "seconds":      **8.00232195854187**,
                                        "bytes":        17413648,
                                        "bits_per_second":      17169884.726746753
                                }
                        }, {
                                "sender":       {
                                        "socket":       18,
                                        "start":        0,
                                        "end":  8.00232195854187, 
                                        "seconds":      **8.00232195854187**,
                                        "bytes":        24251592,
                                        "bits_per_second":      24244555.143511336
                                },
                                "receiver":     {
                                        "socket":       18,
                                        "start":        0,
                                        "end":  8.113577127456665, // <-- this should be the seconds for receiver
[...]

PR #2058, opened 2026-07-08
Fix JSON end interval receiver seconds value

davidBar-On:
master 3.21+

* Issues fixed (if any): #836

`end_time` was used for the "_end:streams:receiver:seconds_" value instead of `receiver_time`.


bmah888:
Thanks for the PR! I haven't tested this yet but it'd be super-cool if this was the simple fix to a long-standing bug.

PR #2066, opened 2026-08-13
Keep server listener open between tests

ross:
I ran into this while working on a personal project in the network/wifi measurement space. I was originally just using `--server` and everything was working fine, but the process was updated to use `--json` on the server side (client always was.) and I started seeing intermittent, but reliable failures. The process ran around 12 iperf tests and I only had all 12 run successfully once. Every other time it died 2-6 tests in. This was happening on iperf 3.20 and HEAD of master (as of yesterday.) 

I was able to work around the problem by putting an 0.025s sleep in between iperf runs. After that I've run 100s of invocations without a single issue. 

## Problem

A persistent iperf3 server closes and recreates its listening socket after every test. If the next client connects while the previous test is finishing—an especially visible window with `--json` on slower hardware—the connection can already be queued on that listener when it is closed. The kernel then resets the queued control connection.

## Fix

Keep the control listener open between successful tests in persistent-server mode and reuse it on the next server iteration. Error cleanup and `--one-off` retain their existing close behavior.

## Validation

The project builds cleanly and all five `make check` tests pass. Targeted loopback stress testing completed 140 immediate sequential connections without failure: 100 TCP tests, 20 reverse two-stream TCP tests with a custom socket window (exercising listener replacement), and 20 UDP tests. A separate `--one-off` test completed and exited normally.

- [x] Repeat the before/after validation on the slower/older NAS or Raspberry Pi-class hardware where the reset was originally observed.


ross:
> * [x]  Repeat the before/after validation on the slower/older NAS or Raspberry Pi-class hardware where the reset was originally observed.

Just ran for 15m straight with zero problems on my NAS (where I originally hit this.)  Went back to main and hit it between the first and second runs in trail 1. Third and fourth in trail 2.

bmah888:
Thanks for the PR!

PR #2071, opened 2026-08-19
fix zerocopy on Solaris

kulikjak:
Currently, `-Z`/`--zerocopy` argument doesn't work on Solaris. `HAVE_SENDFILE` is defined, but then there is no `sendfile` implementation to be used, which results in nothing ever being sent:
```
Accepted connection from 127.0.0.1, port 59470
[  5] local 127.0.0.1 port 5201 connected to 127.0.0.1 port 55152
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   1.00-2.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   2.00-3.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   3.00-4.00   sec  0.00 Bytes  0.00 bits/sec
[  5]   4.00-5.00   sec  0.00 Bytes  0.00 bits/sec
.....
```
This PR send Solaris the Linux way, which fixes the issue.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -43,7 +43,7 @@
 #endif
 
 #ifdef HAVE_SENDFILE
-#ifdef linux
+#if defined(linux) || defined(__sun)
 #include <sys/sendfile.h>
 #else
 #ifdef __FreeBSD__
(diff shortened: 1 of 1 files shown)


bmah888:
Thanks for the PR! We'll take a look, even though Solaris isn't an officially supported OS we've taken compatibility fixes such as this one for Solaris and other UNIX-y platforms in the past.

PR #2077, opened 2026-09-13
udp: fix server reply address for multi-address interfaces

### LINKED ISSUE #2059: iperf 3.21: UDP server with multiple IP aliases may send replies from unexpected addresses
Hi iperf3 Team,

It was noticed that when a UDP server has multiple IP aliases bound to the same interface, replies to the client may be sent from an unexpected alias, causing the client to refuse them.

For example, this can be replicated locally with the following method:

```
# Create a new network instance
sudo ip net add iperf3

# Create a veth pair
sudo ip link add type veth

# Assign one veth to the new namespace
sudo ip link set veth1 netns iperf3

# Assign primary and secondary IPs to the server side
sudo ip net exec iperf3 ip addr add 172.16.0.101/24 dev veth1
sudo ip net exec iperf3 ip addr add 172.16.0.102/32 dev veth1
sudo ip net exec iperf3 ip link set lo up
sudo ip net exec iperf3 ip link set veth1 up

# Setup local/client side
sudo ip addr add 172.16.0.1/24 dev veth0
sudo ip link set veth0 up

# Start the server
sudo ip net exec iperf3 iperf3 -sD

# Start the UDP test
iperf3 -c 172.16.0.102 -u
```

Once executed, the client fails after a timeout:

```
iperf3 -c 172.16.0.102 -u
Connecting to host 172.16.0.102, port 5201
iperf3: error - unable to read from stream socket: Resource temporarily unavailable
```

The packet analysis shows that the server replies from the primary IP instead of the secondary one that was requested by the client:

```
$ sudo tcpdump -ni veth0
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on veth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
00:48:23.057082 IP 172.16.0.1.45366 > 172.16.0.102.5201: Flags [S], seq 4141894149, win 64240, options [mss 1460,sackOK,TS val 3389603211 ecr 0,nop,wscale 10], length 0
00:48:23.057136 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [S.], seq 1522049059, ack 4141894150, win 65160, options [mss 1460,sackOK,TS val 947939836 ecr 3389603211,nop,wscale 10], length 0
00:48:23.057161 IP 172.16.0.1.45366 > 172.16.0.102.5201: Flags [.], ack 1, win 63, options [nop,nop,TS val 3389603211 ecr 947939836], length 0
00:48:23.057215 IP 172.16.0.1.45366 > 172.16.0.102.5201: Flags [P.], seq 1:38, ack 1, win 63, options [nop,nop,TS val 3389603211 ecr 947939836], length 37
00:48:23.057229 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [.], ack 38, win 64, options [nop,nop,TS val 947939836 ecr 3389603211], length 0
00:48:23.057367 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [P.], seq 1:2, ack 38, win 64, options [nop,nop,TS val 947939836 ecr 3389603211], length 1
00:48:23.057391 IP 172.16.0.1.45366 > 172.16.0.102.5201: Flags [.], ack 2, win 63, options [nop,nop,TS val 3389603211 ecr 947939836], length 0
00:48:23.057503 IP 172.16.0.1.45366 > 172.16.0.102.5201: Flags [P.], seq 38:42, ack 2, win 63, options [nop,nop,TS val 3389603211 ecr 947939836], length 4
00:48:23.057537 IP 172.16.0.1.45366 > 172.16.0.102.5201: Flags [P.], seq 42:256, ack 2, win 63, options [nop,nop,TS val 3389603211 ecr 947939836], length 214
00:48:23.057623 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [.], ack 256, win 64, options [nop,nop,TS val 947939836 ecr 3389603211], length 0
00:48:23.057844 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [P.], seq 2:3, ack 256, win 64, options [nop,nop,TS val 947939837 ecr 3389603211], length 1

vvvvvvvvvvvvvvvvvvvv
00:48:23.057997 IP 172.16.0.1.58529 > 172.16.0.102.5201: UDP, length 4
00:48:23.058222 IP 172.16.0.101.5201 > 172.16.0.1.58529: UDP, length 4
00:48:23.058266 IP 172.16.0.1 > 172.16.0.101: ICMP 172.16.0.1 udp port 58529 unreachable, length 40
^^^^^^^^^^^^^^^^^^^^^

00:48:23.058571 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [P.], seq 3:4, ack 256, win 64, options [nop,nop,TS val 947939837 ecr 3389603211], length 1
00:48:23.058625 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [P.], seq 4:5, ack 256, win 64, options [nop,nop,TS val 947939837 ecr 3389603211], length 1
00:48:23.061890 IP 172.16.0.102.5201 > 172.16.0.1.45366: Flags [P.], seq 4:5, ack 256, win 64, options [nop,nop,TS val 947939841 ecr 3389603211], length 1
00:48:23.061905 IP 172.16.0.1.45366 > 172.16.0.102.5201: Flags [.], ack 5, win 63, options [nop,nop,TS val 3389603216 ecr 947939837,nop,nop,sack 1 {4:5}], length 0
```

TCP tests are not affected.

As far as I understand, it is the default Linux kernel behaviour for UDP sockets. However, I wonder if iperf3 can be involved more in the process and instruct the kernel to use the IP address that was requested by the client.

Thank you.

Regards,
Garri

Aniketnegi03:
A UDP test against a server that has more than one address (IPv4 or IPv6) on the outgoing interface fails: the server replies from a different address than the one the client sent its datagrams to, the client's connected socket discards those replies, and the client eventually dies with

    _**_iperf3: error - unable to read from stream socket:
            Resource temporarily unavailable_**_

The server's UDP socket is created by netannounce() with no local address unless -B was given, so it is bound to the wildcard address and has no source address of its own.  iperf_udp_accept() then connect(2)s it to the client, at which point the kernel has to choose a source address and does so with a route lookup towards the client.  That yields the interface's primary address, which need not be the address the client used.  TCP has been unaffected because accept(2) inherits the full address pair of the connection.

The client uses the same server address for the control connection and for the data streams, so the local address of the control connection is the address the client expects replies from.  Bind the UDP socket to it and leave the kernel nothing to guess.  If that address cannot be determined or cannot be bound, fall back to the previous wildcard bind.

As an effect the server now also reports the correct local address for UDP streams in its human-readable and JSON output.

Fixes #2059

```
```

    * Issues fixed (if any): #2059
  
      Introduce `iperf_udp_announce()` which binds the server-side UDP data         
      socket to the local address of the control connection rather than the         
      wildcard address, ensuring the kernel uses the correct source address         
      when connect(2)ing to the client.


bmah888:
Thanks for the PR!

davidBar-On:
@Aniketnegi03 I am not from the iperf3 maintenance team and I did not try running the PR code.  However, I am not sure I understand the fix, and I will appreciate it if you can briefly explain it.

What I don't understand is that the main change is done for `Create a new "listening" socket to replace the one we were using before.`.  I.e. for creating the new listener socket for the next stream.  However, I thought that the problem is in `connect()`.  That is, that the `connect()` is done to different server IP address than the address the "connect message" was received from.  Isn't this the case?  Thanks.

Aniketnegi03:
@davidBar-On, thanks for taking the time to review - let me explain in brief.

You're right that `connect()` is where the wrong source address gets chosen. Might you already know the reason underneath it:

```
> /* __ip4_datagram_connect(), net/ipv4/datagram.c */
> if (!inet->inet_saddr) inet->inet_saddr = fl4->saddr;  /* -> 172.16.0.101 */

```
So when the socket has no source address of its own, the kernel fills one in from a route lookup toward the client. That lookup returns the route's preferred source - the interface's primary address - which need not be the address the client actually contacted (or on which the connect message was received). The problem isn't that the right address is unavailable; it's that we let the route table choose instead of replying from the address the connect message arrived on.

Regarding "_Create a new "Listening" socket to replace the one we were using before_" - I think you read that as the patch only fixing the next stream's socket. It isn't: the socket serving the current stream was already bound with the local address by `iperf_udp_listen()`, when it was first set up to listen.

The ordering makes this clearer. After the TCP handshake and parameter exchange (`get_parameters()`), the server calls `protocol->listen() -> iperf_udp_listen()`, which creates and binds the UDP socket - and only then does it send `CREATE_STREAMS` to the client, which is what prompts the client to send its UDP connect message. So the socket is bound before the probe can arrive:

### Just an example to help get the context:

```text
Server has primary 192.168.1.101 and secondary 192.168.1.102
Client is 192.1681.1 and contacts the iperf server at secondary, 192.168.1.102

--- issue case ---                      --- after fix ---
socket()                      = 5       socket()                       = 5
bind(5, "::")                 = 0       bind(5, "::ffff:192.168.1.102") = 0
recvfrom(5, <CONNECT_MSG>)    = 4       recvfrom(5, <CONNECT_MSG>)     = 4
connect(5, ...192.168.1.1)     = 0       connect(5, ...192.168.1.1)      = 0
  ^ kernel picks 192.168.1.101             ^ source already set, nothing to pick
```

This is also why the fix can't go in `connect()` itself: there is no source argument (`connect(fd, dst, len)`), and it can't be corrected afterwards either - `netannounce()` has already bound the socket to the wildcard. **The source has to be right before `connect()` runs, i.e. at socket creation.**

Once that socket is promoted to stream 1's data socket, a connected UDP socket only receives from its peer, so it can no longer hear the next client - hence the new listening socket in `iperf_udp_accept()`, which needs the same binding and so calls `iperf_udp_announce()` too. That covers streams 2..N. 
In a single-stream test that second socket is created and then closed unused, so the hunk we have discussed does nothing there - incase of single-stream `iperf_udp_listen()` change is helping only.

---

```

